PR #6, opened 2019-02-14
Add CoreCLR bindings to NovelRT

Erlite:
This is still in progress, but I figured I could give the draft PR feature a test.
This PR aims to add CoreCLR support to NovelRT, allowing developers to create their visual novels using C#. This will also be helpful for implementing Ink.

The first step is to add the CoreCLR binaries for the following platforms:
- [x] Linux x64
- [x] Linux ARM x86
- [x] Linux ARM x64
- [x] Linux Alpine x64
- [ ] Mac OS X
- [ ] Windows x86
- [ ] Windows x64

Then:
- [ ] We add some black magic that adds the binaries to the TPA list (courtesy of Centi)
- [ ] We write a CMake to automatically add the correct version of the CoreCLR binaries depending on OS and architecture.
- [ ] Create an interface for C# to access.

Once all of this is done, we should have basic C# support.


Erlite:
I've run into a couple of issues, since the binaries for Mac OS X and Windows lack some of the required files. Currently waiting for a solution.

RubyNova:
> I've run into a couple of issues, since the binaries for Mac OS X and Windows lack some of the required files. Currently waiting for a solution.

Was this resolved after the discussion with tanner? @Erlite

tannergooding:
For reference, I left my notes on this here: https://github.com/novelrt/NovelRT/issues/14

RubyNova:
PR no longer needed. Also out of date.

PR #9, opened 2019-02-21
C wrappers for C++ core

Arkrait:
This PR will contain C wrappers for exisiting C++ code so that it can be easily linked to.


src/NovelRunner_C.cpp
```diff
@@ -0,0 +1,32 @@
+//
+// Created by miron on 2/20/19.
+//
+
+#include "NovelRunner.h"
+#include "NovelRunner_C.h"
```
RubyNova: Any chance we could change the name of this file to something that doesn't include an underscore here? make-do snake case is a little weird.

src/NovelRunner_C.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef NOVELRT_NOVELRUNNER_C_H
+#define NOVELRT_NOVELRUNNER_C_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+    typedef void (*NovelSubscriber)(const float);
+
+    struct NovelRunner_C;
+    typedef struct NovelRunner_C NovelRunner_t;
```
RubyNova: Can we fix the type names at all or does it have to be this way?

RubyNova:
Closed. Outed and abandoned.

PR #16, opened 2019-08-22
Updating the code so MSVC builds succesfully

tannergooding:
This fixes various bits that were preventing NovelRT from successfully building and running on Windows.

After these changes:
image

### DIFF
--- a/src/NovelBasicFillRect.cpp
+++ b/src/NovelBasicFillRect.cpp
@@ -35,14 +35,12 @@ void NovelBasicFillRect::configureBuffer() {
       bottomRight.getX(), bottomRight.getY(), 0.0f,
   };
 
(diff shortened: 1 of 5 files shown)


src/NovelBasicFillRect.cpp
```diff
@@ -35,14 +35,12 @@ void NovelBasicFillRect::configureBuffer() {
       bottomRight.getX(), bottomRight.getY(), 0.0f,
   };
 
-  GLfloat targetArray[_vertexBufferData.size()];
```
tannergooding: This is depending on C99 behavior, which MSVC doesn't support.

However, directly using the underlying buffer is better anyways.

tannergooding:
CC. @Tylertron1998 who was also trying Windows.

PR #22, opened 2019-08-29
Fixing rendering for ImageRect by ensuring a size is set.

tannergooding:
This allows a world space size to be passed into the NovelImageRect constructor. This ensures it has a size so when rendered it is visible.

It might also be beneficial to expose a way to default to the image size, but I'll leave that to a future PR.


tannergooding:
Also added a very basic .editorconfig to ensure that other editors (like VS) respect the 2 space indentation style, rather than inserting tabs or 4 space indents.

PR #26, opened 2019-08-30
Updating GeoVector to wrap glm::vec<2, T>

tannergooding:
This updates GeoVector to wrap the glm vector type instead. This should make supporting additional math functionality simpler moving forward.

### DIFF
--- a/src/GeoVector.h
+++ b/src/GeoVector.h
@@ -4,8 +4,8 @@
 
 #ifndef NOVELRT_COORDINATEVECTOR_H
 #define NOVELRT_COORDINATEVECTOR_H
-#define _USE_MATH_DEFINES
-#include <math.h>
+#include <glm/glm.hpp>
+#include <glm/gtx/rotate_vector.hpp>
 
 namespace NovelRT {
 template<typename T>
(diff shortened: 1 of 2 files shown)


src/GeoVector.h
```diff
@@ -1,137 +1,129 @@
...
-  auto sin0 = sinf(angle);
-
-  setX(((x * cos0) - (y * sin0)) + ox);
-  setY(((y * cos0) + (x * sin0)) + oy);
-}
-
-}
-
-#endif //NOVELRT_COORDINATEVECTOR_H
+//
```
tannergooding: Most of the files look to be using `\r\n` currently.

Going to submit a PR to normalize them to `\n` and to enforce `\n` checkin in the `.gitattributes`.

CC. @RubyNova

tannergooding: Fixed with https://github.com/novelrt/NovelRT/pull/27

tannergooding:
Rebased to resolve merge conflicts now that https://github.com/novelrt/NovelRT/pull/27 has been merged (diff is now smaller and more readable as a result, only impacting the actually changed lines).

tannergooding:
Its a private field, as you don't necessarily want to expose that you are wrapping a `glm::vec` (otherwise, people could depend on that and its a break to change it out later).

The operators are themselves members of the struct and so they should just access the field directly.

tannergooding:
After a chat on Discord, updated `GeoVector` to access the `_value` field through a get/set method. This will allow easier refactoring/changes later; if determined they are needed. It also makes adding support for logging or other diagnostics simpler.

src/GeoVector.h
```diff
@@ -39,97 +39,101 @@ class GeoVector {
...
 template<typename T>
 T GeoVector<T>::getX() const {
-  return _x;
+  return getValue().x;
 }
 
 template<typename T>
 void GeoVector<T>::setX(T value) {
-  GeoVector::_x = value;
+  _value.x = value;
```
tannergooding: This, and the `setY` method are the only places still accessing `_value` directly.

I don't think its desirable to construct a new vector, setting both x and y in this case; since the code would need to look like (or similar to):
```cpp
auto vec = getValue();
vec.x = value;
setValue(vec);
```

tannergooding: Maybe if it returned a ref instead?

RubyNova: Yep im happy with that.

PR #27, opened 2019-08-30
Normalizing line endings to be 'lf' (\n) throughout the codebase

tannergooding:
Looks like most of the files, as checked in to the blob database, were `\r\n` for some reason.


.gitattributes
```diff
@@ -1,2 +1,15 @@
+###############################################################################
```
tannergooding: This file is the only additions (the +13 lines), everything else is just line ending changes

PR #28, opened 2019-08-30
Move to depending on the vcpkg of lua

tannergooding:
This removes the checked in copy of lua and moves to the vcpkg version instead. Since this doesn't have native cmake support, resolving the header and library files is a little bit different.


tannergooding:
Would be good if someone could validate this on Ubuntu as well.

tannergooding:
Rebased to resolve merge conflicts now that https://github.com/novelrt/NovelRT/pull/27 was merged.

First commit removes the checked in lua files (`./cmake` and `./lib` folders, as well as the corresponding `.gitattribute` entries).

Second commit fixes up the `./CMakeLists.txt` to reference the vcpkg lua files and to fix up `./src/main.cpp` to use the proper `#include` declarations.

.gitattributes
```diff
@@ -10,6 +10,3 @@
 #   treat as binary
 ###############################################################################
 *.png       binary
-
-lib/* linguist-vendored
```
tannergooding: `.gitattributes` change is already here

Issue #13, opened 2019-06-05
Basic Audio Implementation


capnkenny:
[AB#2](https://dev.azure.com/NovelRT/16e5528b-8bb4-4e2b-b064-e26926458e42/_workitems/edit/2)

Requirements for a basic Audio Implementation (new SDL_Mixer v1):

Stuff needed:
- [x] Add SDL_mixer to project
- [ ] Add music / sound
- [ ] Play music / sound individually
- [ ] Play music & sound together

capnkenny:
@RubyNova as we discussed, seems like these would be the initial requirements for a MVP. Can you think of anything else to be added in the "MVP" phase?

RubyNova:
WAV is not required for MVP as it is windows only.

Can't see anything else that's required at this time.

capnkenny:
Updated so that this reflects current state for audio

capnkenny:
A new revelation based upon how the current audio branch works - currently to have an "event-driven" sequence, the example I have written in `main.cpp` utilizes `SDL_Delay()` which is undesirable as it freezes the main (rendering) thread for NovelRT while allowing audio to continue.

For this, would we want to incorporate an event system? If so, would the Interaction Service be a good starting point for this?
Also, would the above be out of scope of this issue?

RubyNova:
Why is this utilising `SDL_Delay()`?

capnkenny:
Without delaying the thread, the main thread would attempt to run every function immediately (remember, only in `main`, it is not used in the actual service).

RubyNova:
Im not sure i entirely understand. Please post a permalink to the line in question.

capnkenny:
https://github.com/novelrt/NovelRT/blob/8f570910bdf1c3237ea732d66c2b9f6c2b09e7c4/src/main.cpp#L105

RubyNova:
Ok Ive had a read.

You are correct - `SDL_Delay()` is a bad idea here.

What I propose instead is we implement a queueing system somewhere inside the audio service heirarchy. The best way to do this would simply be with an `std::queue<NovelRT::AudioInformation>` where `AudioInformation` is simply a `struct` that is defined like so:
```cpp
struct AudioInformation {
  std::string fileName;
  int playTime;
}
```
From here you can assemble instances of this and add them to the queue (based on your API as per this comment).

[...]

PR #33, opened 2019-09-04
First Audio Implementation

capnkenny:
#13, [AB#14](https://dev.azure.com/NovelRT/16e5528b-8bb4-4e2b-b064-e26926458e42/_workitems/edit/14), [AB#15](https://dev.azure.com/NovelRT/16e5528b-8bb4-4e2b-b064-e26926458e42/_workitems/edit/15)

Initial Audio Implementation:
-Supports WAV, OGG, MP3
-Implements AudioService to manage all channels, effects, as well as loading/start/stop/pause of music (BGM) and sounds (SFX)


src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "NovelAudioChannel.h"
+#include <string>
+#include <map>
+
+namespace NovelRT {
+NovelAudioChannel::NovelAudioChannel(std::string input, bool isBgm, int channel){
+  if (isBgm)
+  {
+    _loadFile(input, 0);
+    _isMusic = true;
```
RubyNova: This could be simplified to `_isMusic = isBgm`

src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+  else
+  {
+    _loadFile(input, channel);
+    _isMusic = false;
+    _channel = channel;
+  }
+}
+
+bool NovelAudioChannel::_loadFile(std::string in, int channelAssigned){
+  if (!channelAssigned == 0)
```
RubyNova: uhhh that hurts my head to read. Can you clarify this? I'm unsure what the condition is supposed to be.

src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+  {
+    _effect = Mix_LoadWAV(in.c_str());
+    if (_effect == NULL)
```
RubyNova: were in C++, please use `nullptr`.

src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+      std::cout << Mix_GetError() << std::endl;
+      return false;
+    }
+  }
+  else
+  {
+    _bgm = Mix_LoadMUS(in.c_str());
+    if (_bgm == NULL)
```
RubyNova: were in C++, please use nullptr.

src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+Mix_Chunk* NovelAudioChannel::getChunk(){
+  return _effect;
+}
+
+void NovelAudioChannel::play()
+{
+  if (_isMusic)
+  {
+  if (Mix_PlayingMusic() == 0)
```
RubyNova: looks like the indentation is a little messed up here.

src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+  {
+    Mix_PlayMusic(_bgm, 0);
+    std::cout << "Playing music on Channel 0!" << std::endl;
+  }
+  else
+  {
+    if (Mix_PausedMusic() == 1)
```
RubyNova: Couldn't you move this condition up a scope and add a new `else` block for `Mix_PauseMusic`?

src/NovelAudioChannel.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+    else
+    {
+      Mix_PauseMusic();
+    }
+  }
+  }
+  else
+  {
+    if (Mix_Playing(_channel) == 0)
```
RubyNova: With all these numbers im wondering if we can't begin treating this more like `bool` rather than `int`.

src/NovelAudioChannel.h
```diff
@@ -0,0 +1,27 @@
...
+#include <SDL2/SDL_mixer.h>
+#include <iostream>
+#include <map>
+
+namespace NovelRT{
+class NovelAudioChannel {
+  public:
+    NovelAudioChannel(std::string input, bool isBgm, int channel);
+    void play();
+    Mix_Chunk* getChunk();
```
RubyNova: Raw pointers are not that great. I know I'm using them in the layering service but that's because I hadn't quite figured out unique pointers for our use case (yet, though I think that may change very soon). Prefer `unique_ptr<T>` or `shared_ptr<T>` depending on use case.

src/NovelAudioChannel.h
```diff
@@ -0,0 +1,27 @@
...
+  private:
+    int _channel;
+    bool _isMusic;
+    bool _loadFile(std::string in, int channelAssigned);
+    Mix_Music *_bgm;
```
RubyNova: See my previous comment about raw pointers.

src/NovelAudioChannel.h
```diff
@@ -0,0 +1,27 @@
...
+    Mix_Chunk *_effect;
```
RubyNova: See my previous comment about raw pointers.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,173 @@
...
+    std::cout << "SDL_Mixer Error: " << Mix_GetError() << std::endl;
+    isInitialized = false;
+    return -1;
+  }
+  std::cout << "INFO: Allocating channels..." << std::endl;
+  if (Mix_AllocateChannels(2) < 0)
+  {
+    std::cout << "ERROR: Failed to allocate channels." << std::endl;
+    isInitialized = false;
+    return -1;
```
RubyNova: can we change to `bool` please? :)

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,173 @@
...
+  }
+  std::cout << "INFO: Mixer engaged!" << std::endl;
+
+  return 1;
```
RubyNova: can we change to `bool` please? :)

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,173 @@
...
+//Specifying the filename for playMusic() infers that it has not been played
+//or that the music should be switched.
+void NovelAudioService::playMusic(std::string musicName) {
+  auto existingMusic = _music.find(musicName);
+  if (existingMusic == _music.end())
+  {
+    loadMusic(musicName);
+    existingMusic = _music.find(musicName);
+    if (existingMusic == _music.end())
+    {
```
RubyNova: you could probably remove the bracing and one-line this.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+
+  Mix_PlayMusic(_music[musicName],-1);
+}
+
+//Not specifying a filename assumes that the music channel has been paused.
+//We doublecheck to make sure that is the case
+void NovelAudioService::playMusic() {
+  if (Mix_PausedMusic() == 1)
```
RubyNova: maybe invert this?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,173 @@
...
+      return;
+    }
+    Mix_PlayChannel(_channelMap[soundName], _sounds[soundName], 0);
+  }
+
+}
+
+void NovelAudioService::pauseMusic() {
+  if(Mix_PausedMusic() == 0)
+  {
```
RubyNova: bracing braeks convention. Also maybe consider inverting or one-lining this?

src/NovelAudioService.cpp
capnkenny: Yep, we can. I was trying to maintain a likeness to how both SDL_mixer and SDL2 report errors, but I see that this can get confusing as hell for anyone looking for a failure. Will resolve when updated.

src/NovelAudioService.cpp
capnkenny: Same as previous - will resolve when updated.

src/NovelAudioService.cpp
capnkenny: I'll take a look at this and see about one lining it. Should be possible.

src/NovelAudioService.cpp
capnkenny: This should not be inverted to resume, as it will always return 1 or 0, 1 being that the Music channel is paused. One-lined it though and removed bracing.

src/NovelAudioChannel.cpp
capnkenny: As a FYI, any NovelAudioChannel convos were resolved as this was meant to be removed. (I did read through to see what mistakes were there though 👍 )

CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ set(NOVELRT_RESOURCES
 
 file(COPY ${NOVELRT_RESOURCES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
 
+
```
tannergooding: nit: stray newline

CMakeLists.txt
```diff
@@ -52,6 +53,8 @@ find_package(glad CONFIG REQUIRED)
 find_package(glm CONFIG REQUIRED)
 find_package(SDL2 CONFIG REQUIRED)
 find_package(sdl2-image CONFIG REQUIRED)
+find_package(sdl2-mixer CONFIG REQUIRED)
+find_package(Freetype REQUIRED)
```
tannergooding: nit: freetype is now being added twice

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include "NovelAudioService.h"
+#include <iostream>
+#include <functional>
+
+namespace NovelRT {
+
+NovelAudioService::NovelAudioService(){
+  initializeAudio();
+  _nextChannel = 0;
```
tannergooding: nit: initialization of locals should be done via:
```
NovelAudioService::NovelAudioService() : _nextChannel(0) {
```

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+bool NovelAudioService::initializeAudio(){
+  if (SDL_Init(SDL_INIT_AUDIO) < 0 )
```
tannergooding: It's been a while since I've looked at SDL, but don't you need to call `SDL_Init(0)` somewhere and then use `SDL_InitSubSystem(flags)` if you want to initialize things (like audio vs graphics) separately?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,162 @@
...
+  {
+    std::cout << "ERROR: Cannot play audio!" << std::endl;
+    std::cout << "SDL Error: " << SDL_GetError() << std::endl;
+    return isInitialized;
```
tannergooding: nit: You shouldn't use a local before it was declared nor before it was initialized. Just return the constant `false` or declare  it at the top and return once (which is more efficient codegen).

src/main.cpp
```diff
@@ -57,9 +64,22 @@ int main(int argc, char* argv[]) {
   rectArgs.orderInLayer = 1;
   rectArgs.startingRotation = 0.0f;
 
+<<<<<<< 932d0f710e6a8972c7f0cb87149a34108874135e
```
tannergooding: Unresolved merge conflict?

src/NovelAudioService.cpp
capnkenny: After review, it looks like `SDL_Init()` just makes the subsystem call after setting up crash handling, etc. Probably minimal but additional overhead thats unnecessary, so I'll revise this.

src/main.cpp
capnkenny: Oof.. I have not re-commited my modified main.cpp yet, but that is definitely not a part of the local copy I have. I'll fix this.

src/NovelAudioService.cpp
capnkenny: Fixed, ty

src/NovelAudioService.cpp
capnkenny: I believe I corrected this - maybe a re-review of the new commit would confirm?

src/NovelAudioService.cpp
```diff
@@ -32,7 +32,6 @@ bool NovelAudioService::initializeAudio(){
     std::cout << "INFO: Mixer engaged!" << std::endl;
     isInitialized = true;
   }
-
   return isInitialized;
```
capnkenny: @tannergooding In reference to the last review you did with calling a local that was not initialized, would this fit the bill?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    }
+  }
+
+  auto onChannel = _channelMap.find(soundName);
+  if (onChannel == _channelMap.end())
+  {
+    _channelMap[soundName] = _nextChannel;
+    Mix_PlayChannel(_nextChannel, _sounds[soundName], 0);
+    int nextChannelTest = _nextChannel + 1;
+    _nextChannel = (nextChannelTest >= 2 || nextChannelTest < 0) ? 0 : nextChannelTest;
```
RubyNova: There are so many magic numbers here I don't know what any of this means. Please consider `constexpr`ifying some of this stuff.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+  else
+  {
+    if (Mix_Playing(_channelMap[soundName]) == 1)
+    {
+      std::cout << "Already playing on channel " << _channelMap[soundName] << std::endl;
+      return;
+    }
+
+    Mix_PlayChannel(_channelMap[soundName], _sounds[soundName], 0);
```
RubyNova: magic numbers make me sad.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    _channelMap[soundName] = _nextChannel;
+    Mix_PlayChannel(_nextChannel, _sounds[soundName], 0);
+    int nextChannelTest = _nextChannel + 1;
+    _nextChannel = (nextChannelTest >= 2 || nextChannelTest < 0) ? 0 : nextChannelTest;
+  }
+  else
+  {
+    if (Mix_Playing(_channelMap[soundName]) == 1)
+    {
+      std::cout << "Already playing on channel " << _channelMap[soundName] << std::endl;
```
RubyNova: Consider pre-appending "INFO: " onto this to keep it uniform (we really need a logging framework...).

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    }
+  }
+
+  auto onChannel = _channelMap.find(soundName);
+  if (onChannel == _channelMap.end())
+  {
+    _channelMap[soundName] = _nextChannel;
+    Mix_PlayChannel(_nextChannel, _sounds[soundName], loops-1);
+    int nextChannelTest = _nextChannel + 1;
+    _nextChannel = (nextChannelTest >= 2 || nextChannelTest < 0) ? 0 : nextChannelTest;
```
RubyNova: THE NUMBERS MASON, WHAT DO THEY MEAN?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+  else
+  {
+    if (Mix_Playing(_channelMap[soundName]) == 1)
+    {
+      std::cout << "Already playing on channel " << _channelMap[soundName] << std::endl;
+      return;
+    }
+    Mix_PlayChannel(_channelMap[soundName], _sounds[soundName], loops-1);
```
RubyNova: MASON PLZ

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    _channelMap[soundName] = _nextChannel;
+    Mix_PlayChannel(_nextChannel, _sounds[soundName], loops-1);
+    int nextChannelTest = _nextChannel + 1;
+    _nextChannel = (nextChannelTest >= 2 || nextChannelTest < 0) ? 0 : nextChannelTest;
+  }
+  else
+  {
+    if (Mix_Playing(_channelMap[soundName]) == 1)
+    {
+      std::cout << "Already playing on channel " << _channelMap[soundName] << std::endl;
```
RubyNova: Consider pre-appending "INFO: " to this message for uniformity.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+  return _channelMap[soundName];
+}
+
+void NovelAudioService::setSoundVolume(std::string soundName, float value) {
+  Mix_VolumeChunk(_sounds[soundName], convertToMixVolume(value));
+}
+
+void NovelAudioService::setSoundPosition(std::string soundName, int angle, int distance) {
+  if (_channelMap.find(soundName) != _channelMap.end() && _channelMap[soundName] != -1)
```
RubyNova: N U M B E R S

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  {
+    Mix_SetPosition(_channelMap[soundName], angle, distance);
+  }
+  else
+  {
+    std::cerr << soundName << " is not allocated to a channel already!" << std::endl;
```
RubyNova: Consider pre-appending "INFO: " to this for uniformity.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+}
+
+void NovelAudioService::setSoundDistance(std::string soundName, int distance) {
+    if (_channelMap.find(soundName) != _channelMap.end() && _channelMap[soundName] != -1)
+  {
+    Mix_SetDistance(_channelMap[soundName], distance);
+  }
+  else
+  {
+    std::cerr << soundName << " is not allocated to a channel already!" << std::endl;
```
RubyNova: Consider pre-appending "INFO: " to this for uniformity.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    Mix_SetPosition(_channelMap[soundName], angle, distance);
+  }
+  else
+  {
+    std::cerr << soundName << " is not allocated to a channel already!" << std::endl;
+  }
+}
+
+void NovelAudioService::setSoundDistance(std::string soundName, int distance) {
+    if (_channelMap.find(soundName) != _channelMap.end() && _channelMap[soundName] != -1)
```
RubyNova: I feel like I'm watching a magic show with all these magic numbers

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    Mix_SetDistance(_channelMap[soundName], distance);
+  }
+  else
+  {
+    std::cerr << soundName << " is not allocated to a channel already!" << std::endl;
+  }
+}
+
+void NovelAudioService::setSoundPanning(std::string soundName, int leftChannelVolume, int rightChannelVolume) {
+    if (_channelMap.find(soundName) != _channelMap.end() && _channelMap[soundName] != -1)
```
RubyNova: your spacing appears to be off here? Also magic number.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  {
+    Mix_SetPanning(_channelMap[soundName], leftChannelVolume, rightChannelVolume);
+  }
+  else
+  {
+    std::cerr << soundName << " is not allocated to a channel already!" << std::endl;
```
RubyNova: Consider pre-appending "INFO: " to this for uniformity.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  if (exists != _music.end()) return;
+
+  Mix_Music* newMusic = Mix_LoadMUS(input.c_str());
+  if (newMusic != nullptr)
+  {
+    _music[input] = newMusic;
+  }
+  else
+  {
+    std::cerr << Mix_GetError() << std::endl;
```
RubyNova: Consider pre-appending "ERROR: " to this for uniformity.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  auto existingMusic = _music.find(musicName);
+  if (existingMusic == _music.end())
+  {
+    loadMusic(musicName);
+    existingMusic = _music.find(musicName);
+    if (existingMusic == _music.end()) return;
+  }
+  _channelMap[musicName] = -3;
+  Mix_PlayMusic(_music[musicName],-1);
+  _musicTime = SDL_GetPerformanceCounter();
```
RubyNova: Nothing wrong with this per se but I can't actually understand why were assigning the SDL performance counter. Could you elaborate?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+void NovelAudioService::playMusic(std::string musicName) {
+  auto existingMusic = _music.find(musicName);
+  if (existingMusic == _music.end())
+  {
+    loadMusic(musicName);
+    existingMusic = _music.find(musicName);
+    if (existingMusic == _music.end()) return;
+  }
+  _channelMap[musicName] = -3;
+  Mix_PlayMusic(_music[musicName],-1);
```
RubyNova: Mason's having a field day!

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  auto existingMusic = _music.find(musicName);
+  if (existingMusic == _music.end())
+  {
+    loadMusic(musicName);
+    existingMusic = _music.find(musicName);
+    if (existingMusic == _music.end()) return;
+  }
+  _channelMap[musicName] = -3;
+  Mix_PlayMusic(_music[musicName], loops-1);
```
RubyNova: WINGARDIUM LEVIOSA!

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+    Mix_FadeInMusic(_music[musicName], 0, ms);
+  }
+  else
+  {
+    Mix_FadeOutMusic(1000);
+    switch(Mix_GetMusicType(_music[musicName]))
+    {
+      case MUS_OGG:
+      {
+        double pos = (double)((_musicPausedTime - _musicTime)*1000) / (double)SDL_GetPerformanceFrequency();
```
RubyNova: EXPECTO PATRONUM!

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+      {
+        Mix_RewindMusic();
+        double pos = (double)((_musicPausedTime - _musicTime)*1000) / (double)SDL_GetPerformanceFrequency();
+        Mix_SetMusicPosition(pos);
+        Mix_FadeInMusic(_music[musicName], 0, ms);
+        break;
+      }
+      default:
+      {
+        Mix_FadeInMusic(_music[musicName], 0, ms);
```
RubyNova: You know I'm running out of spells so I'll just leave this comment at the end of this block because boi do I feel like I'm in harry potter and the flying magic numbers right now.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+        break;
+      }
+    }
+  }
+}
+
+void NovelAudioService::fadeMusicIn(std::string musicName, int loops, int ms) {
```
RubyNova: This entire method has magic numbers everywhere plzfixkthx

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  _musicPausedTime = SDL_GetPerformanceCounter();
+}
+
+void NovelAudioService::setGlobalVolume(float value) {
+  Mix_Volume(-1, convertToMixVolume(value));
+  Mix_VolumeMusic(convertToMixVolume(value));
+}
+
+int NovelAudioService::convertToMixVolume(float value) {
+  int converted = (value > 1.0f || value < 0.0f) ? 1.0f : (int)(SDL_MIX_MAXVOLUME * value);
```
RubyNova: NUMBAAAAAHZ

src/NovelAudioService.cpp
capnkenny: We use the SDL perf counter as a means of getting the time that music started, and then once again when music is paused or halted. This is for the functionality of resuming where you left off (once MP3/OGG support is working; all other formats do not support this and will never resume if you use a fade effect, it will only resume if you pause).

Also, SDL2-Mixer provides no public functionality for keeping track of position

src/NovelAudioService.cpp
tannergooding: Yes. It's being explicitly set in the constructor, so looks good.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+  else
+  {
+    std::cout << "INFO: Mixer engaged!" << std::endl;
+    isInitialized = true;
+  }
+  return isInitialized;
+}
+
+void NovelAudioService::loadSound(std::string input) {
```
tannergooding: Not for this PR, but as a future design note.

It would be good if you gave back some kind of "opaque handle" that could be used instead of the file name. Basically just something that, for an already loaded sound, can be used to play a sound with `O(1)` (or really cheap) lookup.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  if (newSound != nullptr)
+  {
+    _sounds[input] = newSound;
+  }
+  else
+  {
+    std::cerr << Mix_GetError() << std::endl;
+  }
+
+
```
tannergooding: nit: stray newline

src/NovelAudioService.cpp
tannergooding: 👍

It's too bad C++ doesn't have named parameters like C#.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  Mix_FreeChunk(_sounds[input]);
+  _sounds.erase(existingSound);
+}
+
+void NovelAudioService::playSound(std::string soundName) {
+  auto existingSound = _sounds.find(soundName);
+  if (existingSound == _sounds.end())
+  {
+    loadSound(soundName);
+    existingSound = _sounds.find(soundName);
```
tannergooding: Having an `opaque handle` returned would also prevent the double lookup here.

For the overload that takes a name, you would effectively just have 
```cpp
auto soundHandle = loadSound(soundName);
playSound(soundHandle);
```

src/NovelAudioService.cpp
tannergooding: You only support stereo audio right now?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+      std::cout << "Already playing on channel " << _channelMap[soundName] << std::endl;
+      return;
+    }
+
+    Mix_PlayChannel(_channelMap[soundName], _sounds[soundName], 0);
+  }
+
+}
+
+void NovelAudioService::playSound(std::string soundName, int loops) {
```
tannergooding: nit: The two overloads of `playSound` could really just be a single implementation with a single `if check` to handle `loops == 0`

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+void NovelAudioService::stopSound(std::string soundName) {
+  Mix_HaltChannel(_channelMap[soundName]);
+}
+
+int NovelAudioService::getSoundChannel(std::string soundName) {
+  auto existingSound = _channelMap.find(soundName);
+  if (existingSound == _channelMap.end())
+  {
+    std::cerr << "ERROR: Cannot get channel for sound " << soundName << std::endl;
+    return -9;
```
tannergooding: Why `-9`?

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  {
+    Mix_SetPanning(_channelMap[soundName], leftChannelVolume, rightChannelVolume);
+  }
+  else
+  {
+    std::cerr << soundName << " is not allocated to a channel already!" << std::endl;
+  }
+}
+
+void NovelAudioService::loadMusic(std::string input) {
```
tannergooding: What is the intended difference between `music` and `sound`?

Is this not something that could just be handled via an `enum` type or auto-detection; so you could have a single `load`/`play` API?

src/NovelAudioService.cpp
capnkenny: SDL_Mixer by default only supports stereo audio - currently, I do define to use only two mixing channels.

src/NovelAudioService.cpp
tannergooding: Seems like it would be better to define a `Timestamp` type which abstracts this away from the perf counters.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  if (!Mix_PausedMusic())
+  {
+    Mix_PauseMusic();
+    _musicPausedTime = SDL_GetPerformanceCounter();
+  }
+}
+
+void NovelAudioService::stopMusic() {
+  Mix_HaltMusic();
+  _musicPausedTime = SDL_GetPerformanceCounter();
```
tannergooding: `Stop` and `Pause` seem like two different things...

src/NovelAudioService.cpp
capnkenny: Music and Sound are referenced differently in SDL_Mixer (i.e. `MixMusic` vs `MixChunk`), hence the separation here.

Would wrapping it in a `struct` (so `enum` type and file name) work to condense these?

src/NovelAudioService.cpp
capnkenny: Let me remove that....

src/NovelAudioService.cpp
tannergooding: Here is another reason why having a `Timestamp` type would be good.

There is a subtle, but non obvious bug here that deals with floating-point values and timestamps in general. Namely around `Frequency` being significantly larger than the `TicksPerSecond`, which you've defined to be `1000`. For example, on my box, `Frequency` is `10'000'000`, which means that this has a risk of producing a really tiny result and incurring rounding error.

RubyNova:
As an additional note - you should update the readme as it says audio is coming soon, but with this branch it'll actually exist. 😄

src/main.cpp
```diff
@@ -78,10 +92,56 @@ int main(int argc, char* argv[]) {
...
+  novelAudio->fadeMusicIn("sparta.wav", -1, 5000);
+  novelAudio->setGlobalVolume(0.5);
+
+
+  auto rect = runner.getInteractionService()->getBasicInteractionRect(NovelRT::GeoVector<float>(200, 200), rectArgs2);
+
+  rect->subscribeToInteracted([&novelAudio]{
+      std::cout << "Commencing SFX test..." << std::endl;
+      novelAudio->fadeMusicOut(500);
+      SDL_Delay(600);
```
tannergooding: Isn't this going to prevent the render loop from running >.>

capnkenny: Yes, its only for demonstration as of now. Ruby and I had discussed this on Discord - most likely we'll need to handle this, but it would be part of the HLAPI and separate from this PR (unless you have some input that would go against that?)

RubyNova: Ah...yeah this should probably be cleaned up as per #13 .

RubyNova: To clarify - just make this not block the render loop for now, the HLAPI stuff will be a different PR.

src/NovelAudioService.cpp
tannergooding: Right, but I think my question is: "does this difference need to be exposed to the user as part of the method to call".

It seems like taking a struct that defines the name, enum type, and any other info and then having the APIs internally lookup what the "type" is, so it knows which mixer API to call is better.

tannergooding: -- This also allows easier use of other audio backends if you ever want to support them 😄

src/NovelAudioService.cpp
tannergooding: Might be good to define this as a field/property; so it can be easily changed in the future (and it isn't a hunt and search for everywhere you hardcoded `2`).

src/main.cpp
capnkenny: Erm, that will make every channel play on top of itself which will not work - in that case, should I just remove this demo?

tannergooding: In this case, you could pretty trivially not block the render loop by just checking the elapsed time since the last call was played.

tannergooding:
Please fix the commit history (happy to help if needed).

This is reintroducing practically every commit from master (it looks like a bad rebase).

src/NovelAudioService.cpp
capnkenny: (inb4 I'm asked about OpenAL in the future)
It shouldn't be exposed really, so I'll condense this in the next push

capnkenny:
> 
> 
> Please fix the commit history (happy to help if needed).
> 
> This is reintroducing practically every commit from master (it looks like a bad rebase).

_plz_ halp ;_;
Once I am active again, I'll ping in discord?

src/NovelAudioService.cpp
capnkenny: Most recent push actually addresses this by putting in a `const` in the header so it can be reassigned if need be.

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+  if (exists != _sounds.end()) return;
+
+  Mix_Chunk* newSound = Mix_LoadWAV(input.c_str());
+  if (newSound != nullptr)
+  {
+    _sounds[input] = newSound;
+  }
+  else
+  {
+    std::cerr << Mix_GetError() << std::endl;
```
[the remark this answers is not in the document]
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
```diff
@@ -0,0 +1,367 @@
...
+#include <functional>
+
+namespace NovelRT {
+
+NovelAudioService::NovelAudioService() : _nextChannel(1), _musicTime(0), _musicPausedTime(0), isInitialized(false) {
+  initializeAudio();
+}
+
+bool NovelAudioService::initializeAudio() {
+  if (SDL_InitSubSystem(SDL_INIT_AUDIO) < 0 )
```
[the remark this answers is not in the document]
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: Fixed in latest commit 👍

src/NovelAudioService.cpp
capnkenny: To be honest.... -3 is an internally used number (exclusive to NovelRT), -2 is the Mixer's postmix channel, -1 is Mixer's any/all channel definition, and anything 0 or positive can be a possible channel. Hence (-9) - it'll never be accepted by Mixer, and it's easily identifiable by us. :D
(in retrospect, I'll probably change this since it's random. Just realized it was defined as -9 in the newest commit too >.<)

FYI @RubyNova

RubyNova: Ahahahahaha oh my lord that's amazing.

Maybe make a `const` for it if you absolutely must?

capnkenny: Newest commit already did, but I'm going to remove it and change it tbh. Re-reading the code, if I find a `-1` with this function, it should return the error as a channel hasn't actually been assigned yet.

capnkenny: Removed in latest commit 👍 (after cleaning up history)

src/main.cpp
capnkenny: I removed this test and replaced with an interactive one - I'd like to re-implement at a later time.

src/NovelAudioService.cpp
capnkenny: Removed 👍

src/NovelAudioService.cpp
capnkenny: Condensed the music and sound loading functions in the latest push

src/NovelAudioService.cpp
capnkenny: Added two if checks (if 0, no loop / if -1, infinite loop), and condensed

src/NovelAudioService.cpp
capnkenny: Fixed the harry potter sorcery over here!

src/NovelRTUtilities.h
```diff
@@ -1,7 +1,19 @@
...
 
-#ifndef NOVELRT_NOVELRTMACROUTILITIES_H
-#define NOVELRT_NOVELRTMACROUTILITIES_H
+#ifndef NOVELRT_NOVELRTUTILITIES_H
+#define NOVELRT_NOVELRTUTILITIES_H
+
+namespace NovelRT {
+class NovelUtilities {
+public:
+  static const int SDL_SUCCESS = 0;
```
tannergooding: These seems like things that would be better as enums or class constants...

src/main.cpp
```diff
@@ -61,6 +68,14 @@ int main(int argc, char* argv[]) {
   //basicFillRect = runner.getRenderer()->getBasicFillRect(NovelRT::GeoVector<float>(200, 200), NovelRT::RGBAConfig(0, 255, 255, 255), rectArgs);
   auto textRect = runner.getRenderer()->getTextRect(NovelRT::RGBAConfig(0, 255, 0, 255), 70, "Gayathri-Regular.ttf", rectArgs);
   textRect->setText("RubyGnomer");
+  auto rectArgs2 = NovelRT::NovelCommonArgs();
```
tannergooding: nit: Maybe give this a better name, like `playAudioButtonArgs`....

src/main.cpp
```diff
@@ -61,6 +68,14 @@ int main(int argc, char* argv[]) {
...
+  rectArgs2.startingPosition = novelChanArgs.startingPosition;
+  rectArgs2.startingPosition.setX(rectArgs.startingPosition.getX() - 800);
+  rectArgs2.layer = 0;
+  rectArgs2.orderInLayer = 1;
+  rectArgs2.startingRotation = 0.0f;
+
+  basicFillRect = runner.getRenderer()->getBasicFillRect(NovelRT::GeoVector<float>(200, 200), NovelRT::RGBAConfig(0, 255, 255, 255), rectArgs2);
```
tannergooding: nit: Maybe give this a better name, like `playAudioButton`.

A small text overlay that says `Play Audio` might also be a nice touch :P

src/NovelRTUtilities.h
capnkenny: `SDL_SUCCESS` is universally used (@RubyNova said he'd want to switch to this for the other SDL calls that return a 0 on success) - but you'd mean to move the others to the class instead of just having them in a utilities class?

tannergooding: Sorry, this was meant to apply to all the constants in here.

`SDL_SUCCESS` in particular seems like something that could be solved by a helper method. Namely something like `ThrowIfFailed(MyFunc(args), "Error Message")`

This would then check for success, log the error, and throw an exception.

tannergooding: The others seem like they would be better exposed as class constants where they are relevant or as global constants exposed as part of the appropriate header file.

src/NovelAudioService.cpp
capnkenny: Issue created!

src/NovelAudioService.cpp
capnkenny: Done, issue created!

src/NovelRTUtilities.h
capnkenny: Moved the constants to the header file for NovelAudioService (as they're related to the service or SDL_Mixer). Opened an issue for the helper method example with `SDL_SUCCESS`.

capnkenny:
Addressed the bugs / nits / cleanup from Tanner's second review, and created issues for outstanding topics. Leaving for Matt to dismiss his requested change and to merge in the AM 👍

tannergooding:
Found two things:
1. Could you modify `scripts\build.ps1` and `scripts\build.sh`. In both, just change `install freetype glad glm lua sdl2 sdl2-image` to include `sdl2-mixer` at the end
2. The sample doesn't work out of the box because the wav files aren't checked in, but they are binary files so we probably don't want to check them in. @RubyNova ?

capnkenny:
Good catch, I'll make those changes in a bit. And yeah, I didn't want to check in large binary files (oof .wav filesizes) so I uploaded them to GDrive and posted it in Discord. Let's see what @RubyNova wants to do about that.

PR #34, opened 2019-09-12
Adding a basic StepTimer to allow both fixed and variable rate fps.

tannergooding:
This removes the old delta time logic and replaces it with a `StepTimer` (based on the type of the same name available in the DirectX Toolkit).

The timer supports both fixed-rate and variable-rate updates and should work appropriately on systems with very fast hardware timer frequencies.

This also adds very basic support for printing the current refresh rate out to the terminal, it should later be replaced with a UI text element that can be dynamically enabled/disabled.


CMakeLists.txt
```diff
@@ -41,10 +41,10 @@ include_directories(${LUA_INCLUDE_DIR})
...
+add_executable(NovelRT src/NovelObject.cpp src/NovelObject.h src/NovelInteractionService.h src/NovelLayeringService.cpp  src/GeoVector.h src/NovelBasicFillRect.cpp src/NovelBasicFillRect.h src/RGBAConfig.cpp src/RGBAConfig.h src/main.cpp src/NovelImageRect.cpp src/NovelImageRect.h src/NovelRunner.cpp src/NovelRunner.h src/NovelRenderingService.cpp src/NovelRenderingService.h src/NovelCommonArgs.h  src/NovelRenderObject.cpp src/NovelRenderObject.h src/NovelTextRect.cpp src/NovelTextRect.h src/NovelInteractionService.cpp src/NovelInteractionObject.cpp src/NovelInteractionObject.h src/NovelObjectSortComparison.h src/NovelBasicInteractionRect.cpp src/NovelBasicInteractionRect.h src/GeoBounds.cpp src/GeoBounds.h src/NovelRTMacroUtilities.h src/NovelStepTimer.h src/NovelStepTimer.cpp)
```
tannergooding: It would be nice to glob these, if possible :)

src/NovelRunner.cpp
```diff
@@ -7,22 +7,27 @@
...
   while (_exitCode) {
-    previous = current;
-    current = SDL_GetPerformanceCounter();
-    deltaTime = ((current - previous) * 1000 / SDL_GetPerformanceFrequency()) * 0.001f;
-    executeUpdateSubscriptions(deltaTime);
+    _stepTimer.tick(_updateSubscribers);
+
+    auto framesPerSecond = _stepTimer.getFramesPerSecond();
+
+    if (framesPerSecond != lastFramesPerSecond) {
```
tannergooding: This should be replaced once the support for rendering text elements is merged

src/NovelRunner.h
```diff
@@ -7,23 +7,22 @@
...
-
-typedef void (* NovelSubscriber)(const float);
+#include "NovelStepTimer.h"
 
 namespace NovelRT {
 class NovelRunner {
 public:
-  void runOnUpdate(NovelSubscriber);
-  void stopRunningOnUpdate(NovelSubscriber);
-  void executeUpdateSubscriptions(const float deltaTime) const;
```
tannergooding: This was removed as updates are now executed by the `StepTimer::tick` method.

src/NovelStepTimer.cpp
```diff
@@ -0,0 +1,109 @@
...
+// This is based on the StepTimer provided in the DirectX ToolKit
+// Original code is available under the MIT License
+
+#include "NovelStepTimer.h"
+#include <SDL2/SDL.h>
+
+namespace NovelRT {
+  StepTimer::StepTimer(uint32_t targetFrameRate) :
+    _frequency(SDL_GetPerformanceFrequency()),
+    _maxCounterDelta(_frequency / 10), // 1/10th of a second
```
tannergooding: It might be nice to make this configurable, it would need to be taken in as a `double maxSecondsDelta` and stored as ticks

src/NovelStepTimer.cpp
```diff
@@ -0,0 +1,109 @@
...
+    if (counterDelta > _maxCounterDelta)
+    {
+      // This handles excessibly large deltas to avoid overcompting.
+      // It is particularly beneficial to do this when debugging, for example
+      counterDelta = _maxCounterDelta;
+    }
+
+    // Convert to the "canonicalized" tick format of TicksPerSecond
+    // This will never overflow due to the clamping we did above
+    auto ticksDelta = (counterDelta * TicksPerSecond) / _frequency;
```
tannergooding: Technically this can bug out if `_frequency` is greater than TicksPerSecond.

That can be fixed by converting to double first, but you have to adjust other logic as well (which isnt as straightforward to do since you have excessively small deltas that cant be converted to anything other than integer 0 (which leads to other logic getting "off").

tannergooding: There are other workarounds as well, such as only dealing in terms of _frequency for your "ticks per second", but that also comes with tradeoffs.

At the end of the day, using 100ns ticks should be good enough as that is what most OS normalizes their High Performance Frequency Counters at and is significantly faster than most things can process frames in real world applications anyways.

(It also only causes bugs for variable refresh rate and can be easily solved by locking the refresh rate at some number such that 1 frame takes at least 100ns)

src/NovelStepTimer.h
```diff
@@ -0,0 +1,125 @@
...
+
+  private:
+    const uint64_t _frequency;
+    const uint64_t _maxCounterDelta;
+
+    uint64_t _lastCounter;
+    uint64_t _secondCounter;
+    uint64_t _remainingTicks;
+
+    uint64_t _elapsedTicks;
```
tannergooding: Most values are tracked in terms of "ticks" which are unsigned integers. This ensures rounding error doesn't need to be thought about and results are always as accurate as possible.

src/main.cpp
```diff
@@ -59,11 +59,11 @@ int main(int argc, char* argv[]) {
 
   basicFillRect = runner.getRenderer()->getBasicFillRect(NovelRT::GeoVector<float>(200, 200), NovelRT::RGBAConfig(0, 255, 255, 255), rectArgs);
 
-  runner.runOnUpdate([](const float delta) {
+  runner.runOnUpdate([](const double delta) {
```
tannergooding: The update delta, however, is a double as that allows interpolation and other things. Die to clamping and other logic it is sufficiently small that rounding error is generally not a concern. It is also double instead of float to avoid these issues as well (and gives users the ability to do more precise updates when/if needed).

I can link to other articles/technical reports explaining why the delta should be double (rather than float) if desired.

CMakeLists.txt
RubyNova: The problem with GLOBbing is that you lose the real time updates from your IDE AFAIK. Every time I looked into it its also generally frowned upon because of merge-related concerns. You can find out a little more here: https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil

src/NovelRunner.cpp
RubyNova: What would you want to replace it with?

Issue #37, opened 2019-09-12
Update CMake with a more readable configuration for adding source files.


RubyNova:
As per a discussion on PR #34 started by @tannergooding the CMake's source file handling is becoming a little unwieldy and unreadable. After the discussion the proposed solution is to do the following:
```
set (NOVELRT_SOURCES
 a.cpp
 b.cpp
 c.cpp
)

set (NOVELRT_HEADERS
  a.h
  b.h
  c.h
)
```
Followed by:
```
add_executable(NovelRT ${NOVELRT_SOURCES} ${NOVELRT_HEADERS})
```

This should help make it more readable.

PR #42, opened 2019-09-12
Update CMake with a more readable configuration for adding source files.

tannergooding:
This resolves https://github.com/novelrt/NovelRT/issues/37.

This updates the CMakeLists.txt to use cmake variables to track the lists of files used by various portions of the build. This makes it easier to see what flows where, what is included where, and makes diffs much smaller/easier to review.


RubyNova:
I need to quickly test how CLion handles auto updating the build targets with this. I don't think it's really a big deal myself but if this stops working and someone else objects then they need to speak up before it's merged.

CMakeLists.txt
```diff
@@ -45,12 +60,99 @@ include_directories(${LUA_INCLUDE_DIR})
...
+      ${NOVELRT_LINK_LIBRARIES}
+      m
+    )
+endif()
+
+set(NOVELRT_EXECUTABLE_HEADERS
+  ${NOVELRT_HEADERS}
+)
+
+set(NOVELRT_EXECUTABLE_SOURCES
```
tannergooding: Was thinking about this.... It might be better if the exe, rather than pulling in the sources directly, instead depends on the lib (whether static or shared; although only a shared is being produced right now).

RubyNova: Ticket #48 now exists to discuss this further. :)

PR #47, opened 2019-09-13
Texture loading failure fix.

### LINKED ISSUE #36: Fix Intermittent Texture Failures on Linux
Currently, the text rendering PR introduces an intermittent bug where the rendering of textures (i.e. Novel-Chan in this case) fails. So far, it's only reproducing in LInux (specifically Ubuntu atm).

image

RubyNova:
This resolves #36 by implementing lazy loading and simplifying buffer creation.


src/LazyFunction.h
```diff
@@ -0,0 +1,43 @@
...
+//
+
+#ifndef NOVELRT_LAZYFUNCTION_H
+#define NOVELRT_LAZYFUNCTION_H
+#include <functional>
+#include <memory>
+
+namespace NovelRT {
+template<typename T>
+class LazyFunction {
```
tannergooding: just interested: Why `LazyFunction` as the name?

RubyNova: I was tired and it seemed appropriate. Shrug. I can change it if you think it should just be Lazy or something else.

tannergooding: I'm not sure its the most fitting since its a value that is lazily initialized and not necessarily a function that is lazy to call.
Maybe `LazyValue` if you didn't like just `Lazy`?

RubyNova: I'll address this now.

src/NovelImageRect.cpp
```diff
@@ -9,16 +9,35 @@
 #include "GeoBounds.h"
 
 namespace NovelRT {
+//glGenBuffers(1, &_uvBuffer);
```
tannergooding: stray comment?

src/NovelImageRect.cpp
```diff
@@ -83,86 +121,85 @@ void NovelImageRect::drawObject() const {
...
+
+  glTexImage2D(GL_TEXTURE_2D, 0, mode, surface->w, surface->h, 0, mode, GL_UNSIGNED_BYTE, surface->pixels);
+  glGenerateMipmap(GL_TEXTURE_2D);
+  SDL_FreeSurface(surface);
 }
-void NovelImageRect::setTextureInternal(const GLuint textureId) {
-  _textureId = textureId;
+
+void NovelImageRect::setTextureInternal(GLuint textureId) {
+  _textureIsSelfManaged = false;
```
tannergooding: why this rather than just checking for an empty `imageDir`?

src/NovelBasicFillRect.cpp
```diff
@@ -77,14 +82,8 @@ void NovelBasicFillRect::configureObjectBuffers(bool refreshBuffers) {
...
-  if(refreshBuffers) {
-    glGenBuffers(1, &_colourBuffer);
-    glBindBuffer(GL_ARRAY_BUFFER, _colourBuffer);
-  }
-
+  auto colourBuffer = _colourBuffer.getActual();
+  glBindBuffer(GL_ARRAY_BUFFER, colourBuffer);
   glBufferData(GL_ARRAY_BUFFER, sizeof(GLfloat) * _colourData.size(), _colourData.data(), GL_STATIC_DRAW);
 }
-NovelBasicFillRect::~NovelBasicFillRect() {
```
tannergooding: If you expose a `Lazy<T>.isCreated()` method, you can do:
```
if (_vertexArrayObject.isCreated())
{
    glDeleteVertexArrays(1, &_vertexArrayObject.getActual());
}
```

tannergooding: Would probably be good to start adding ogl cleanup everywhere now

RubyNova: Unfortunately I can't use a reference on an rvalue of type `GLuint` but ill add the `isCreated()` check.

src/NovelRenderObject.cpp
```diff
@@ -39,13 +49,10 @@ void NovelRenderObject::configureObjectBuffers(const bool refreshBuffers) {
       bottomRight.getX(), bottomRight.getY(), 0.0f,
   };
 
-  if (refreshBuffers) {
-    glGenVertexArrays(1, &_vertexArrayObject);
-    glGenBuffers(1, &_buffer);
-  }
+  _vertexArrayObject.getActual(); //this is just here to force initialisation.
```
tannergooding: Does this need to be here? It doesn't look to be used anywhere by this function...

RubyNova: Yes otherwise the VAO will not exist and I figured using a `Lazy` would be easier than adding my own custom check around the VAO generation call. Also it makes teardown based on `Lazy` objects easier.

valgrind.xml
```diff
@@ -0,0 +1,3565 @@
+<?xml version="1.0"?>
```
tannergooding: I don't think this is supposed to be checked in, maybe add to .gitignore?

RubyNova: Oof, I don't know how that got in. Will remove.

src/NovelBasicFillRect.cpp
tannergooding:
```
auto temp = _vertexArrayObject.getActual();
glDeleteVertexArrays(1, &temp);
_vertexArrayObject.reset();
```

src/NovelRenderObject.cpp
tannergooding: Right, but at what point does it need to be created, because it will eventually be created in the respective `drawObject` call.

RubyNova: the VAO must exist before the first buffer is created. Buffers are tied to VAOs as their "parent" so to speak, hence it is created before anything else.

src/NovelImageRect.cpp
```diff
@@ -65,7 +65,7 @@ NovelImageRect::NovelImageRect(NovelLayeringService* layeringService,
                                                                  glGenTextures(1, &tempTexture);
                                                                  return tempTexture;
                                                                })){
-  _textureIsSelfManaged = false;
+  _imageDir = "";
```
tannergooding: `: _imageDir("")`

src/NovelBasicFillRect.cpp
RubyNova: is the reset call actually required? Also this functionality was moved to the base render object type.

tannergooding: no, probably not. In the case of C++, the finalizer is called under object deletion scenario, so the object won't stay around (unlike C#).

src/NovelImageRect.cpp
```diff
@@ -9,16 +9,33 @@
...
-                               const RGBAConfig& colourTint) :
-    NovelRenderObject(layeringService, screenScale, size, args, programId),
-    _imageDir(imageDir), _colourTint(colourTint) {
+                               const RGBAConfig& colourTint) : NovelRenderObject(layeringService,
+                                                                                 screenScale,
+                                                                                 size,
+                                                                                 args,
+                                                                                 programId),
+                                                               _imageDir(imageDir), _colourTint(colourTint),
+                                                               _uvBuffer(Lazy<GLuint>([] {
```
tannergooding: Since this pattern is repeated and so small, would it be a good idea to add `protected: static NovelRenderObject::createGenBuffer() const`?

These all then become `_uvBuffer(Lazy<GLuint>(createGenBuffer))`

src/NovelImageRect.cpp
```diff
@@ -27,25 +44,43 @@ NovelImageRect::NovelImageRect(NovelLayeringService* layeringService,
                                const GeoVector<float>& size,
                                const NovelCommonArgs& args,
                                GLuint programId,
-                               const RGBAConfig& colourTint) :  NovelRenderObject(layeringService, screenScale, size, args, programId),
-_imageDir(""), _colourTint(colourTint) {
+                               const RGBAConfig& colourTint) : NovelRenderObject(layeringService,
```
tannergooding: Why repeat all this logic, rather than just forwarding to the above constructor with an empty `imageDir`?

src/NovelRenderObject.cpp
```diff
@@ -11,19 +11,29 @@ NovelRenderObject::NovelRenderObject(NovelLayeringService* layeringService,
...
+                                                                 GLuint tempBuffer;
+                                                                 glGenBuffers(1, &tempBuffer);
+                                                                 return tempBuffer;
+                                                               })) {
 }
 
 void NovelRenderObject::executeObjectBehaviour() {
-  if(!_bufferInitialised) {
-    configureObjectBuffers(true);
+  if (!_bufferInitialised) {
```
tannergooding: Just to ensure I'm on the same page...

Now, if the buffer needs to be refreshed, you'll just reset `_bufferInitialised` to `false`, correct?

src/NovelRenderingService.cpp
```diff
@@ -18,6 +18,24 @@
 #include <fstream>
 #include <sstream>
 
+//DO NOT DELETE THIS, MOVE THIS TO DEBUG SERVICE WHEN IT EXISTS
```
tannergooding: Maybe move this into `DebugService.h` since the file already exists?

src/NovelTextRect.cpp
```diff
@@ -20,67 +20,66 @@ void NovelTextRect::setColourConfig(const RGBAConfig& value) {
...
+        GeoVector<int>(face->glyph->bitmap.width, face->glyph->bitmap.rows),
+        GeoVector<int>(face->glyph->bitmap_left, face->glyph->bitmap_top),
+        static_cast<AdvanceInteger>(face->glyph->advance.x)
+    };
+    _fontCharacters.insert(std::pair<GLchar, GraphicsCharacterRenderData>(c, character));
   }
+  FT_Done_Face(face);
+  FT_Done_FreeType(freeTypeLoader);
+
+  //glGenBuffers(1, &_textBuffer);
```
tannergooding: This can be removed, right?

RubyNova: Yep. Comment removed.

src/NovelRenderingService.cpp
RubyNova: I will move the comment but not implement it as I was unaware this was a 4.3 API. Will resolve that in another branch however.

src/NovelRenderObject.cpp
RubyNova: Erm, good point actually. I had completely forgotten about this switch and will need to look further into this before this branch is merged.

src/NovelImageRect.cpp
RubyNova: Fixed.

src/NovelImageRect.cpp
RubyNova: Fixed!

src/NovelRenderObject.cpp
RubyNova: OK I've had a further look. This switch is there to ensure it is only initialised once via executing default object behaviour.

By "refresh" if you mean overwrite existing data in the buffers, that is the default behaviour of each object. If you mean to delete and generate new buffers, that is out of the scope of this PR with the changes introduced here. At the moment the only way to tear down buffers is to delete the object that manages it.

Perhaps you should open a ticket if you believe we should support having objects rebuild their buffers from scratch without also destroying the object in its entirety?

RubyNova: Resolved in Discord.

tannergooding: As I commented on Discord:
> ah, ok. I think I see.
Previously (current state, without the PR) refreshBuffers was used to decide if genBuffers should be called
now, you just rely on the buffer properly getting created via the lazy initializer
so, the parameter isn't needed
I was thinking it was also doing some other bits, but I'm just crazy
LGTM then

PR #51, opened 2019-09-14
Object deletion API.

RubyNova:
This PR is for implementing the object deletion API as per #AB40 for all `NovelObject`s.

### DIFF
--- a/src/NovelLayeringService.cpp
+++ b/src/NovelLayeringService.cpp
@@ -6,16 +6,13 @@
 #include "NovelLayeringService.h"
 #include "NovelObjectSortComparison.h"
 
+
 namespace NovelRT {
 void NovelLayeringService::updateLayerInfo(const int layer, NovelObject* targetObject, const bool migrate) {
-  if (migrate) {
-    auto vec = _layerMatrix[targetObject->getLayer()];
-    vec.erase(std::remove_if(vec.begin(), vec.end(), [targetObject](const NovelObject* x) {
-      auto result = x == targetObject;
-      return result;
-    }), vec.end());
-  }
   auto it = _layerMatrix.find(layer);
+  if (migrate && it != _layerMatrix.end()) {
+    deleteFromMatrix(targetObject);
+  }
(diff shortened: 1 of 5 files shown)


RubyNova:
I need more eyes on this. I am unhappy with the current implementation but cannot see a way to make it nicer at the present.

tannergooding:
More a comment on an over-arching design here; but, basically you have two sides of resources to manage (CPU side and GPU side). For the purposes of below, I'm going to refer to create/delete as the CPU side and bind/unbind as the GPU side.

Essentially, users will create objects on the CPU side. Some of these are not-expensive (i.e. rectangles), others can be potentially expensive (i.e. images), and some will eventually be a mix (i.e. text, where the font will be shared/cached and is "expensive", but the glyphs and string to render are "cheap").

For this, you'll probably want to allow users to create an object but not load the assert (into CPU memory). You probably want loading of the asset to be lazy (it doesn't happen until its actually needed the first time).

Then, the next step is actually using it on the GPU, which involves "binding" (copying the resource to the GPU, creating the appropriate resource ids, etc). This again should be "lazy" (you don't want to be binding 100s of objects in a scene graph if, on average, only a handful will be used per frame).

Creation is then pretty straightforward. But, for deletion and unbinding, there is both a bit of "needs to automatically happen" and a bit of "the user should be able to control this". For example, you probably don't want to hardcode it so that a resource is unbound as soon as `someProperty == false`, as the user might be displaying it again in a few frames (in which case unbinding is more expensive).

To that effect, you ultimately have a few states:
* isActive -- is the object bound to the GPU
* isEnabled -- can the object receive input
* isVisible -- is the object actually being rendered

These states work in conjunction with each other to determine everything with `isActive` and `isVisible` largely playing off one another. For example, `isVisible` can't be true unless `isActive` is also true. As long as `isActive` is true, the resource is a GPU resource and can be rendered etc (and for some things, like a primitive rectangle, `isActive` might always be `true`).

RubyNova:
I'm closing this due to its age and having another go at it this week.

PR #52, opened 2019-09-14
[HOTFIX] - Inverted Y axis for RenderObjects.

RubyNova:
This is a quick fix for an issue that cropped up in Discord around a few hours before this PR was created. This should resolve the Y axis being the other way around for rendering-based objects.


RubyNova:
This branch is no longer valid as this will become part of an upcoming feature branch.

PR #53, opened 2019-09-15
Various correctness and reliability fixes

Nihlus:
RE: Our discussion on Discord earlier, I've enabled a variety of useful GCC options and fixed a number of issues raised by the enabled inspections. Among these are some signed-to-unsigned comparisons (which can produce unexpected results), some undefined behaviour (calling nonvirtual destructors on abstract classes), and some initialization ordering issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,8 @@ target_compile_options(NovelRT
     -Wdouble-promotion
     -Wshadow
     -Wformat=2
+    -Wno-error=unused-parameter
+    -Wno-error=unused-variable
 )
 
 add_library(NovelRTLib SHARED ${NOVELRT_LIBRARY_HEADERS} ${NOVELRT_LIBRARY_SOURCES})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
 cmake_minimum_required(VERSION 3.13)
-set(CMAKE_TOOLCHAIN_FILE /home/jarl/Programming/source-projects/vcpkg/scripts/buildsystems/vcpkg.cmake)
 
 project(NovelRT VERSION 0.0.1)
 include(GNUInstallDirs)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,6 +152,22 @@ set(NOVELRT_LIBRARY_LINK_LIBRARIES
   ${NOVELRT_LINK_LIBRARIES}
 )
 
+target_compile_options(NovelRT
+  PRIVATE
+    -Werror
+    -Wall
+    -Wextra
+    -Wduplicated-cond
+    -Wduplicated-branches
+    -Wlogical-op
+    -Wrestrict
+    -Wnull-dereference
+    -Wuseless-cast
+    -Wdouble-promotion
+    -Wshadow
+    -Wformat=2
+)
+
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -150,6 +151,84 @@ set(NOVELRT_LIBRARY_LINK_LIBRARIES
   ${NOVELRT_LINK_LIBRARIES}
 )
 
+target_compile_options(NovelRT
```
tannergooding: This is going to break Windows, it needs to be conditioned to just GCC/Clang

src/NovelStepTimer.cpp
```diff
@@ -58,12 +58,12 @@ namespace NovelRT {
...
       // from accumulating over time. Without this clamping, a game that requested a 60 fps
       // fixed update, running with vsync enabled on a 59.94 NTSC display, would eventually
-      // accumulate enough tiny errors that it would drop a frame. It is better to just round 
+      // accumulate enough tiny errors that it would drop a frame. It is better to just round
       // small deviations down to zero to leave things running smoothly.
 
       auto targetElapsedTicks = _targetElapsedTicks;
 
-      if (abs((int64_t)(ticksDelta - targetElapsedTicks)) < (TicksPerSecond / 4000)) {
+      if (abs((int64_t)(ticksDelta - targetElapsedTicks)) < (long)(TicksPerSecond / 4000)) {
```
tannergooding: What was the warning given for needing to add the cast to `long`? `TicksPerSecond` is an `int64_t` constant, so the type should have been preserved....

However, if it is needed, it should also be cast to `int64_t` since `long` is 32-bits on Windows.

Nihlus: Comparision between signed and unsigned types. `TicksPerSecond` is declared as a `uint64_t`, not `int64_t`.

src/NovelRenderingService.cpp
```diff
@@ -52,12 +70,14 @@ bool NovelRenderingService::initializeRenderPipeline(const int displayNumber) {
...
   SDL_GL_MakeCurrent(_window.get(), _openGLContext);
-  SDL_GL_SetSwapInterval(0);
 
   if (!gladLoadGL()) {
     std::cerr << "ERROR: Failed to initialise glad." << std::endl;
     return -1;
   }
+    // During init, enable debug output
+    glEnable              ( GL_DEBUG_OUTPUT );
+    glDebugMessageCallback( MessageCallback, 0 );
```
tannergooding: I believe this was temporarily disabled since it requires 4.3 or higher and the minimum is currently 3.2.

Windows (and some Linux configurations) will actually give you explicitly the version requested, in which case:
* The versioning logic needs to request 3.2 as a minimum and allow a later version if availble
* The `glDebugMessageCallback` needs to be "light up" based on what version is actually being used

It might be easiest to log a bug for that and to set 4.3 as the explicit version right now. @RubyNova ?

RubyNova: Yeah that might be wise. I wanted to deal with it later but if its being reenabled we should put it in the `NovelDebugService` as discussed and leave a TODO to revisit the issue later. :)

Nihlus: I can drop it for now and leave it for you folks to reimplement at a later time :)

PR #54, opened 2019-09-20
Adding a basic service for loading and using .NET

tannergooding:
This adds a basic service for loading and using .NET. This uses the official hosting APIs and therefore works with normal redistribution of the .NET Runtime.


tannergooding:
image

src/NovelCoreClrService.cpp
```diff
@@ -0,0 +1,130 @@
...
+#if defined(_WIN32)
+  #include <Windows.h>
+#else
+  #include <dlfcn.h>
+#endif
+
+const auto HostApiBufferTooSmall = 0x80008098;
+
+static void* getNativeExport(void* nativeLibrary, const char* exportName) {
+#if defined(_WIN32)
```
tannergooding: Need to have some logic for loading and finding the hostfxr exports that differs between Windows and Unix.

src/NovelCoreClrService.cpp
```diff
@@ -0,0 +1,130 @@
...
+    std::cerr << "ERROR: failed to locate hostfxr" << std::endl;
+    throw -1;
+  }
+
+  return loadNativeLibrary(buffer.data());
+}
+
+namespace NovelRT {
+  NovelCoreClrService::NovelCoreClrService() :
+    _hostContextHandle(Lazy<hostfxr_handle>([&, this] {
```
tannergooding: All the function pointers, the hostfxr resolution logic, etc are lazy.

src/NovelCoreClrService.h
```diff
@@ -0,0 +1,54 @@
...
+    Lazy<hostfxr_handle> _hostContextHandle;
+    Lazy<void*> _hostfxr;
+    Lazy<hostfxr_initialize_for_runtime_config_fn> _hostfxr_initialize_for_runtime_config;
+    Lazy<hostfxr_get_runtime_delegate_fn> _hostfxr_get_runtime_delegate;
+    Lazy<hostfxr_close_fn> _hostfxr_close;
+    Lazy<load_assembly_and_get_function_pointer_fn> _load_assembly_and_get_function_pointer;
+  };
+
+  template <class _Fty>
+  std::function<_Fty> NovelCoreClrService::getFunction(const char_t* assemblyPath, const char_t* typeName, const char_t* methodName, const char_t* delegateTypeName) {
```
tannergooding: This could be made a bit "simpler" and have some strings be auto-constructed, etc.
However, this is zero overhead and most things (like the Ink service) will be their own APIs that can abstract this away.

src/main.cpp
```diff
@@ -80,6 +80,9 @@ int main(int argc, char* argv[]) {
   auto rect = runner.getInteractionService()->getBasicInteractionRect(NovelRT::GeoVector<float>(200, 200), rectArgs);
   rect->subscribeToInteracted([]{novelChanRect->setActive(!novelChanRect->getActive());});
 
+  auto coreclrService = runner.getCoreClrService();
+  auto testFunction = coreclrService->getFunction<void()>(STR("ClassLibrary1.dll"), STR("ClassLibrary1.Class1, ClassLibrary1"), STR("Test"), STR("ClassLibrary1.TestSignature, ClassLibrary1"));
```
tannergooding: ClassLibrary1.dll isn't included here, since it was just a test.

Before the PR is merged, we should determine how cmake will pull down any managed NuGet packages that are dependencies and replace this with using that instead...

tannergooding:
Test

src/main.cpp
tannergooding: This was changed to a new `NovelRT.Dotnet` csproj which is checked in, but which requires a bit more infrastructure work to properly integrate with the build system.

tannergooding:
Updated this to use the official vcpkg and renamed a few components. The unresolved comments from above are still valid and apply to the corresponding locations in the renamed files.

include/NovelRT.h
```diff
@@ -120,6 +128,7 @@ namespace NovelRT::Windowing {
 //Engine service types
 #include "NovelRT/Audio/AudioService.h"
 #include "NovelRT/DebugService.h"
+#include "NovelRT/Dotnet/DotnetRuntimeService.h"
```
RubyNova: This is just a question really about naming - since its already in `NovelRT::Dotnet` would `RuntimeService` be a better type name? or perhaps `ClrService`? Maybe this is just me, but were calling it `Dotnet` twice here.

I'm also possibly wondering about renaming the namespace to `DotNet` as opposed to `Dotnet` maybe? What do you think?

src/main.cpp
RubyNova: Just to flag this up, we should probably decide together what's handled by the SDK vs what's handled by the engine core in this aspect.

Specifically, should the SDK be handling `dotnet restore` instead as a whole, for example? Or did you want to split it out so that it only handles `dotnet restore` for end-user csproj files?

include/NovelRT.h
tannergooding: Renamed

include/NovelRT/DotNet/RuntimeService.h
```diff
@@ -0,0 +1,51 @@
...
+#define NOVELRT_RuntimeService_H
+
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::DotNet {
+  class RuntimeService {
+  private:
+    NovelRT::Utilities::Lazy<hostfxr_handle> _hostContextHandle;
```
RubyNova: Base namespace is the same; you could just use `Utilities::Lazy<T>` instead. This applies to the others too. 😄

src/CMakeLists.txt
```diff
@@ -32,25 +25,36 @@ set(NOVELRT_LIB_LINK_LIBRARIES
...
+# nethost doesn't provide native cmake support
+
+find_path(NETHOST_INCLUDE_DIRS nethost.h)
+include_directories(${NETHOST_INCLUDE_DIRS})
+
+find_library(NETHOST_LIBRARY NAMES libnethost nethost)
+link_libraries(${NETHOST_LIBRARY})
 
 if (WIN32)
   set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
RubyNova: Is there a reason this is WIN32 specific?

src/NovelRT/DotNet/RuntimeService.cpp
```diff
@@ -0,0 +1,148 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+
+#include <NovelRT.h>
+
+#include <nethost.h>
```
RubyNova: Uhhhh is there a reason this isn't just the singular `NovelRT.h` include?

src/NovelRT/DotNet/RuntimeService.cpp
```diff
@@ -0,0 +1,148 @@
...
+    throw - 1;
+  }
+
+  auto buffer = std::vector<char_t>(buffer_size);
+  result = get_hostfxr_path(buffer.data(), &buffer_size, nullptr);
+
+  if (result != 0)
+  {
+    std::cerr << "ERROR: failed to locate hostfxr" << std::endl;
+    throw - 1;
```
RubyNova: We should probably throw `std::runtime_error` here in the same vein as the other places (even if its not what we want to do in the future, it'd be nice to have this uniform), unless this is going to be exported as a C function or something?

src/NovelRT/DotNet/RuntimeService.cpp
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  auto buffer = std::vector<char_t>(buffer_size);
+  result = get_hostfxr_path(buffer.data(), &buffer_size, nullptr);
+
+  if (result != 0)
+  {
+    std::cerr << "ERROR: failed to locate hostfxr" << std::endl;
```
RubyNova: We should probably be using our logger type here.

src/NovelRT/DotNet/RuntimeService.cpp
```diff
@@ -0,0 +1,148 @@
...
+      const char_t* runtime_config_path = runtimeConfigJsonPath.c_str();
+#else
+      const char_t* runtime_config_path = runtimeConfigJsonPath.string().c_str();
+#endif
+
+      int result = _hostfxr_initialize_for_runtime_config.getActual()(runtime_config_path, nullptr, &hostContextHandle);
+
+      if (result != 0)
+      {
+        std::cerr << "ERROR: failed to initialize the runtime" << std::endl;
```
RubyNova: I won't repeat myself on these next two lines of code, but basically the same suggestions as lines 76 and 77. 😄

src/NovelRT/DotNet/RuntimeService.cpp
tannergooding: It's a header only needed by the implementation and is not needed for 3rd party libraries to also include.

src/NovelRT/DotNet/RuntimeService.cpp
tannergooding: Ah right, forgot to fix these now that we have a logger.

src/CMakeLists.txt
tannergooding: Linux automatically does the right thing.

src/NovelRT/DotNet/RuntimeService.cpp
RubyNova: Ah okay, awesome.

src/CMakeLists.txt
RubyNova: Ah oki doki. Resolving.

PR #67, opened 2019-10-04
Adding some build scripts to the repository

tannergooding:
This adds some basic build scripts to the repository.

The current requirements are that a user first:
1) Install the C++ Toolset
2) Install Vcpkg
3) Use Vcpkg to install `freetype glad glm lua sdl2 sdl2-image`

After that, they need to run:
4) `./generate.sh -DCMAKE_TOOLCHAIN_FILE=$VcpkgDir/scripts/buildsystems/vcpkg.cmake`
5) `./build.sh`

This should produce the necessary build files, etc.

Users can also optionally run just `./build.sh --ci` (or `./scripts/cibuild.sh`) which will automatically do steps 2 thru 5 (inclusive).

### DIFF
--- a/scripts/azure-ci.yml
+++ b/scripts/azure-ci.yml
@@ -1,0 +1,5 @@
+trigger:
+- master
+
+jobs:
+- template: azure-pipelines.yml
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -29,8 +29,8 @@ function Create-Directory([string[]] $Path) {
 
 function Generate() {
   if ($ci) {
-    $vcpkgToolchainFile = Join-Path -Path $VcpkgInstallDir -ChildPath "scripts/buildsystems/vcpkg.cmake"
-    $remaining = ,"-DCMAKE_TOOLCHAIN_FILE=$vcpkgToolchainFile" + $remaining
+    $VcpkgToolchainFile = Join-Path -Path $VcpkgInstallDir -ChildPath "scripts/buildsystems/vcpkg.cmake"
+    $remaining = ,"-DCMAKE_TOOLCHAIN_FILE=$VcpkgToolchainFile" + $remaining
   }
 
   & cmake -S $RepoRoot -B $BuildDir -Wdev -Werror=dev -Wdeprecated -Werror=deprecated -A x64 -DCMAKE_INSTALL_PREFIX=$InstallDir $remaining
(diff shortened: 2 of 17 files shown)


tannergooding:
On Ubuntu 16.04, this looks to be the minimal command line to get all required tools that CI won't auto-download.
```
sudo apt-get update
sudo apt-get install apt-transport-https ca-certificates gnupg software-properties-common wget -y

wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo apt-key add -
sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'

wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key 2>/dev/null | sudo apt-key add -
sudo apt-add-repository 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial main'

sudo apt-get update
sudo apt-get install clang cmake curl g++ tar unzip -y
```

Later Ubuntu versions would be similar, but with the respective distro repos (such as `bionic`, rather than `xenial`, for 18.04).

tannergooding:
Windows requires the C++ Toolchain (generally acquired from Visual Studio) and CMake installed.

scripts/azure-ci.yml
```diff
@@ -0,0 +1,5 @@
+trigger:
```
tannergooding: These yml files are what we can point AzDO towards.

Then, ideally we would enable them only for `feature/ci` (or some branch) while I get dependencies for the hosted pools worked out. Then, one that PR is merged, we can enable CI globally.

scripts/build.ps1
```diff
@@ -0,0 +1,132 @@
+[CmdletBinding(PositionalBinding=$false)]
+Param(
+  [switch] $build,
+  [switch] $ci,
+  [ValidateSet("Debug", "MinSizeRel", "Release", "RelWithDebInfo")][string] $configuration = "Debug",
```
tannergooding: Lets you choose debug vs release, etc

scripts/build.sh
```diff
@@ -0,0 +1,207 @@
...
+}
+
+function CreateDirectory {
+  if [ ! -d "$1" ]; then
+    mkdir -p "$1"
+  fi
+}
+
+function Generate {
+  if [ -z "$remaining" ]; then
```
tannergooding: I was having trouble getting bash to do what I wanted here; so I just manually expanded everything out. It's not pretty, but it works for now.

scripts/build.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+    if (!(Test-Path -Path $VcpkgExe)) {
+      & $VcpkgInstallDir/bootstrap-vcpkg.bat
+
+      if ($LastExitCode -ne 0) {
+        throw "'bootstrap-vcpkg' failed"
+      }
+    }
+
+    & $VcpkgExe install freetype glad glm lua sdl2 sdl2-image --triplet x64-windows
```
tannergooding: I hardcoded this to x64 for now. 32-bit should still work for non-CI, but I didn't want to try to figure this out right now. Will log a bug.

PR #68, opened 2019-10-05
Documentation generation

QuantumlyTangled:
This request configures Doxygen and adds a build command to the CMake file.
To have some reference of how it looks I have also added some different types of comments and very basic documentation to the NovelRunner class.

P.S. If you wish to see how it looks atm open docs/html/index.html in almost any browser.


.gitignore
```diff
@@ -130,3 +130,5 @@ compile_commands.json
 CTestTestfile.cmake
 
 # End of https://www.gitignore.io/api/c++,cmake,clion
+
+.vs/
```
tannergooding: This shouldn't be needed anymore. I added a `.gitignore` for both `vs` and `vscode` in the PR that added build scripts.

docs/html/_geo_bounds_8h_source.html
```diff
@@ -75,7 +75,7 @@
 <div class="ttc" id="aclass_novel_r_t_1_1_geo_bounds_html"><div class="ttname"><a href="class_novel_r_t_1_1_geo_bounds.html">NovelRT::GeoBounds</a></div><div class="ttdef"><b>Definition:</b> GeoBounds.h:9</div></div>
 <!-- start footer part -->
 <hr class="footer"/><address class="footer"><small>
-Generated on Sat Oct 5 2019 18:23:31 for NovelRT by &#160;<a href="http://www.doxygen.org/index.html">
+Generated on Sun Oct 6 2019 17:35:19 for NovelRT by &#160;<a href="http://www.doxygen.org/index.html">
```
tannergooding: Is there a way to avoid this timestamp?

It seems non-ideal to have hundreds of one line edits changing the date/time.

QuantumlyTangled: It was a option I enabled, so yes I can disable it again.

Doxyfile
```diff
@@ -0,0 +1,2567 @@
+# Doxyfile 1.8.16
```
tannergooding: This file should be `.gitignored`. Running `configure_file(${DOXYFILE_IN} ${DOXYFILE_OUT} @ONLY)` as part of build generates it from `Doxyfile.in`, so there is no need to have it checked in.

Doxyfile
```diff
@@ -0,0 +1,2567 @@
...
+# the logo to the output directory.
+
+PROJECT_LOGO           = ./novel-chan_doxy.png
+
+# The OUTPUT_DIRECTORY tag is used to specify the (relative or absolute) path
+# into which the generated documentation will be written. If a relative path is
+# entered, it will be relative to the location where doxygen was started. If
+# left blank the current directory will be used.
+
+OUTPUT_DIRECTORY       = ./docs
```
tannergooding: This should be a path relative to the cmake output directory (likely `${CMAKE_BINARY_DIR}/docs`). Until we decide to make it implicitly checked in (and enforced), we don't want users accidentally including build artifacts as part of a commit.

More likely than not, I would imagine that the CI build takes docs generated as artifacts from a build and pushes them to the website repo, so they can be automatically updated.

PR #75, opened 2019-10-13
Fixing up CI so that it works.

tannergooding:
This fixes up CI so that it works end to end. After this is merged, PR jobs can be enabled and status widgets can be added to the readme.

The passing job is here: https://dev.azure.com/NovelRT/NovelRT/_build/results?buildId=32

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,7 @@ include_directories(${LUA_INCLUDE_DIR})
 find_library(LUA_LIBRARY lua)
 link_libraries(${LUA_LIBRARY})
 
-find_package(Doxygen)
+find_package(Doxygen QUIET)
 
 set(NOVELRT_HEADERS
   src/GeoBounds.h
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ include_directories(${LUA_INCLUDE_DIR})
 find_library(LUA_LIBRARY lua)
 link_libraries(${LUA_LIBRARY})
 
-find_package(Doxygen)
+find_package(Doxygen QUIET)
```
RubyNova: What's this do?

tannergooding: It just prevents cmake from spewing a message about not being able to find Doxygen.

Doxygen is currently optional and we don't want it's non-existence failing CI.

RubyNova: True. So what's the plan for Doxygen generation in our CI/CD process?

tannergooding: Log an issue and figure out what's required to set it up on the CI machines. It will be easier once the initial support is merged as PR jobs can start running.

RubyNova: Logged. Resolving and approving.

PR #93, opened 2019-12-18
feature: Add the option to dynamically set the window title

QuantumlyTangled:
Adds the `setWindowTitle` and `getWindowTitle` methods to `NovelRenderingService`. This allows the user to dynamically set the window title.


src/NovelRenderingService.cpp
```diff
@@ -17,7 +17,7 @@
 #include <sstream>
 
 namespace NovelRT {
-  bool NovelRenderingService::initializeRenderPipeline(int displayNumber) {
+  bool NovelRenderingService::initializeRenderPipeline(int displayNumber, std::string windowTitle) {
```
RubyNova: can you make this a `const std::string&` rather than an `std::string`?

src/NovelRenderingService.cpp
```diff
@@ -61,8 +61,8 @@ namespace NovelRT {
       return -1;
     }
 
-    std::string glVersion = (const char*) glGetString(GL_VERSION);
-    std::string glShading = (const char*) glGetString(GL_SHADING_LANGUAGE_VERSION);
+    std::string glVersion = (const char*)glGetString(GL_VERSION);
```
RubyNova: Please use a static cast here.

src/NovelRenderingService.cpp
```diff
@@ -61,8 +61,8 @@ namespace NovelRT {
...
+    std::string glShading = (const char*)glGetString(GL_SHADING_LANGUAGE_VERSION);
```
RubyNova: Please use a static cast here.

src/NovelRenderingService.cpp
```diff
@@ -176,8 +176,8 @@ namespace NovelRT {
     return returnProg;
   }
 
-  int NovelRenderingService::initialiseRendering(int displayNumber) {
-    if (!initializeRenderPipeline(displayNumber)) {
+  int NovelRenderingService::initialiseRendering(int displayNumber, std::string windowTitle) {
```
RubyNova: can you make this a `const std::string&` rather than an `std::string`?

src/NovelRenderingService.cpp
```diff
@@ -187,6 +187,13 @@ namespace NovelRT {
     return 0;
   }
 
+  std::string NovelRenderingService::getWindowTitle() const {
+    return SDL_GetWindowTitle(getWindow().get());
+  }
+  void NovelRenderingService::setWindowTitle(std::string value) {
```
RubyNova: can you make this a `const std::string&` rather than an `std::string`?

src/NovelRunner.cpp
```diff
@@ -5,10 +5,10 @@
 #include <SDL2/SDL.h>
 
 namespace NovelRT {
-NovelRunner::NovelRunner(int displayNumber, uint32_t targetFrameRate)
+NovelRunner::NovelRunner(int displayNumber, std::string windowTitle, uint32_t targetFrameRate)
```
RubyNova: can you make this a `const std::string&` rather than an `std::string`?

PR #97, opened 2019-12-19
Created centralised header file with all includes added

Banananaman91:


src/Central.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
RubyNova: Please replace `#pragma "once"` with appropriate include guards instead.

Issue #87, opened 2019-12-14
Refactor NovelRunner into Dispatcher/Application Model


capnkenny:
In order to set up multi-threading as well as making it easier to develop future services, we should break `NovelRunner` apart, allowing NovelRT to have the following setup:

> -- Application (Singleton, Exposes Services)
> -- Various Services
> -- Dispatcher (Per Thread, Runs the Event Pump)

Currently, `NovelRunner` does both of what the "Application" and the "Dispatcher" would do, and is not currently a singleton. This can pose many problems, as theoretically end users can create as many runners as they want, which would multiply the number of services per runner.

The suggested method of doing this, as per @tannergooding's comments in Discord, would be to:
> * Rename Runner to Application
>   * It exposes a static getCurrent() method so anyone can retrieve services, etc
> * Break out the event loop logic into some kind of Dispatcher (or Runner) class
>   * This allows you to queue/dispatch pending events
>   * This is "per thread" and are created lazily

This would be a larger refactoring job, and may need to wait until other PRs are merged (logging, UBO, etc).

RubyNova:
Since we've agreed to go full DOD, Application/Dispatcher will probably won't work as well. Instead, we probably want to look into a job system or something similar.

RubyNova:
This will be addressed in an upcoming PR.

PR #98, opened 2019-12-19
LLAPI modularisation.

RubyNova:
[AB#61](https://dev.azure.com/NovelRT/16e5528b-8bb4-4e2b-b064-e26926458e42/_workitems/edit/61)

This PR is the result of a discussion on discord surrounding the current LLAPI structure and how we want to handle object deletion. As a result this PR resolves #51 .


src/NovelDebugService.cpp
```diff
@@ -25,8 +25,8 @@ messageCallback(GLenum source,
   glDebugMessageCallback(messageCallback, 0);*/
 
 #include <limits>
-#include "NovelDebugService.h"
-#include "NovelRunner.h"
+#include "../include/NovelDebugService.h"
+#include "../include/NovelRunner.h"
```
tannergooding: it might be a good idea to just change these to `#include <NovelRT.h>` while you're here

tannergooding: (for all the cpp files)

RubyNova: Only for C++ files? Is there a reason for that?

tannergooding: For the header files, it depends on if we want to enforce users only reference `NovelRT.h` (in which case they should have no includes and might want to error if `NOVELRT_H` isn't defined) or if you want them to be "self-contained"

RubyNova: I see....well, for end-users I'd probably only want them using `NovelRT.h` in their projects, but what do you think is better?

tannergooding: I think that's probably a reasonable starting position and we can change later if needed

RubyNova: Alright, so do I stick with your original request for that and just check for the NovelRT define?

tannergooding: Yes, that should work

RubyNova: Got it!

CMakeLists.txt
```diff
@@ -159,12 +159,12 @@ endif()
```
tannergooding: We should probably rename `NovelRTLib` to `libNovelRT` as that matches the "normal" naming convention

CMakeLists.txt
```diff
@@ -159,12 +159,12 @@ endif()
 
 set(NOVELRT_LIBRARY_HEADERS
   ${NOVELRT_HEADERS}
-  include/NovelRT/NovelRT-C/NovelRunner_C.h
+  #include/NovelRT/NovelRT-C/NovelRunner_C.h
```
tannergooding: We should either remove this (and redo it in the future when we cross that road) or move it to a `libNovelRT-C` project

CMakeLists.txt
tannergooding: With this refactoring, we *should* be able to remove `NOVELRT_HEADERS` and move it to be exclusively in `NOVELRT_LIBRARY_HEADERS`.

`NOVELRT_EXECUTABLE_HEADERS` would then be mostly empty and would find the files via `include_directories`, which makes it match what external users would do.

tannergooding: (we can do this in a separate PR if desired)

src/NovelBasicFillRect.cpp
```diff
@@ -6,11 +6,11 @@ namespace NovelRT::Graphics {
 
 BasicFillRect::BasicFillRect(LayeringService* layeringService,
                                        const RGBAConfig& fillColour,
-                                       const CommonArgs& args,
+                                       const Utilities::CommonArgs& args,
```
tannergooding: Just commenting here as well, since we already talked about it in Discord.

`CommonArgs`, given that its part of the "public" API, should likely not be in `Utilities`

src/NovelBasicFillRect.cpp
```diff
@@ -6,11 +6,11 @@ namespace NovelRT::Graphics {
...
                                        ShaderProgram shaderProgram,
                                        Camera* camera) :
     RenderObject(layeringService, args, shaderProgram, camera), _colourConfig(fillColour),
-    _colourBuffer(Lazy<GLuint>(generateStandardBuffer)) {}
+    _colourBuffer(Utilities::Lazy<GLuint>(generateStandardBuffer)) {}
```
tannergooding: `Lazy`, on the other hand; is really an internal impl detail; so I think `Utilities` is fine 😄

src/NovelDebugService.cpp
RubyNova: Since this branch is now working again (as of this commit) and this structure has been implemented, I am marking this as resolved before I forget.

CMakeLists.txt
```diff
@@ -63,42 +63,44 @@ find_library(LUA_LIBRARY lua)
 link_libraries(${LUA_LIBRARY})
```
tannergooding: it might be nice, as part of this, to move the executable into a `samples` folder since it isn't going to be a shipping thing

CMakeLists.txt
```diff
@@ -63,42 +63,44 @@ find_library(LUA_LIBRARY lua)
...
+  include/NovelRT/Graphics/CameraViewChangedEventArgs.h
+  include/NovelRT/Graphics/ShaderProgram.h
+  include/NovelRT/Graphics/CameraBlock.h
+  include/NovelRT/Windowing/WindowingService.h
+  include/NovelRT/Input/KeyState.h
+  include/NovelRT.h)
 
 set(NOVELRT_SOURCES
-        src/NovelWorldObject.cpp
+  src/NovelWorldObject.cpp
```
tannergooding: nit: sources folder structure should mirror the include folder structure

CMakeLists.txt
RubyNova: Well, this executable might not be - however we will need an executable build that auto starts Lua and any core dependencies such as CoreCLR (since Lua cannot start itself), though I guess that would be a separate project perhaps?

CMakeLists.txt
RubyNova: this is on my todo already. 😄

CMakeLists.txt
```diff
@@ -156,12 +159,12 @@ endif()
 
 set(NOVELRT_LIBRARY_HEADERS
```
tannergooding: You should add a `source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${NOVELRT_LIBRARY_HEADERS})` after this `set` group.

It will cause the generated VS solution to have a nice directory structure (mirroring the on disk one), rather than pooling everything into one `Include Files` folder

CMakeLists.txt
RubyNova: To clarify - this would specifically be for deployments that cannot start themselves, such as Lua, if no "piloting" deployment/API is specified.

CMakeLists.txt
tannergooding: Similar for the other groups (`NOVELRT_LIBRARY_SOURCES` and the corresponding `NOVELRT_EXECUTABLE_*` ones)

include/NovelRT.h
```diff
@@ -0,0 +1,116 @@
+#ifndef NOVELRT_H
+#define NOVELRT_H
+
+
+typedef void (*NovelUpdateSubscriber)(double);
```
tannergooding: This should probably be in the `NovelRT` namespace, yes?

include/NovelRT.h
```diff
@@ -0,0 +1,116 @@
+#ifndef NOVELRT_H
```
tannergooding: nit: Licence header

include/NovelRT.h
RubyNova: Ah...probably. I was just desperate for a working compilation. 🤣

include/NovelRT.h
```diff
@@ -0,0 +1,116 @@
...
+#define NOVELRT_H
+
+
+typedef void (*NovelUpdateSubscriber)(double);
+namespace NovelRT::Maths
+{
+  template<typename T>
+  class GeoVector;
```
tannergooding: We could probably remove the typedefs for the "value types", since they shouldn't be passed around as pointers anyways (and so the typedefs don't matter)

include/NovelRT.h
```diff
@@ -0,0 +1,116 @@
...
+#include "NovelRT/Input/BasicInteractionRect.h"
+
+#include "NovelRT/Audio/AudioService.h"
+#include "NovelRT/DebugService.h"
+#include "NovelRT/Input/InteractionService.h"
+#include "NovelRT/Windowing/WindowingService.h"
+#include "NovelRT/Graphics/RenderingService.h"
+
+
+
```
tannergooding: nit: extra newlines here and at the start of the file

include/NovelRT/Audio/AudioService.h
```diff
@@ -1,17 +1,16 @@
 // Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
 
-#ifndef NOVELRT_NOVELAUDIOSERVICE_H
-#define NOVELRT_NOVELAUDIOSERVICE_H
+#ifndef NOVELRT_AUDIOSERVICE_H
+#define NOVELRT_AUDIOSERVICE_H
 
 #include <SDL2/SDL.h>
```
tannergooding: We can also centralize these includes to `NovelRT.h`

include/NovelRT/Audio/AudioService.h
```diff
@@ -30,7 +29,7 @@ namespace NovelRT {
     SoundBank _sounds;
     MusicBank _music;
     ChannelMap _channelMap;
-    NovelLoggingService _logger;
+    LoggingService _logger;
```
tannergooding: I'd like to better understand why `LoggingService` is being used as a value type

tannergooding: @capnkenny ?

include/NovelRT/DebugService.h
```diff
@@ -3,24 +3,23 @@
 #ifndef NOVELRT_NOVELDEBUGSERVICE_H
 #define NOVELRT_NOVELDEBUGSERVICE_H
 
-#include "NovelTextRect.h"
 #include <glad/glad.h>
 
 namespace NovelRT {
 
   typedef class NovelRunner NovelRunner;
```
tannergooding: We should be able to remove these typedefs now.

include/NovelRT/Graphics/Camera.h
```diff
@@ -1,14 +1,9 @@
...
-#include "GeoMatrix4.h"
-#include "Lazy.h"
-#include "NovelUtilities.h"
-#include "CameraViewChangedEventArgs.h"
-
-namespace NovelRT {
-  class NovelCamera {
+#ifndef NOVELRT_GRAPHICS_CAMERA_H
+#define NOVELRT_GRAPHICS_CAMERA_H
+namespace NovelRT::Graphics {
```
tannergooding: nit: We consistently have a newline between the `#define` and `namespace` elsewhere

include/NovelRT/Graphics/CameraBlock.h
```diff
@@ -1,9 +1,9 @@
...
-#ifndef NOVELRT_CAMERABLOCK_H
-#define NOVELRT_CAMERABLOCK_H
+#ifndef NOVELRT_GRAPHICS_CAMERABLOCK_H
+#define NOVELRT_GRAPHICS_CAMERABLOCK_H
 #include <glm/glm.hpp>
 
-namespace NovelRT {
+namespace NovelRT::Graphics {
   struct CameraBlock {
     glm::mat4 cameraMatrix;
```
tannergooding: nit: This should probably be `GeoMatrix`. IIRC, we had decided `glm` should be an internal implementation detail for now.

include/NovelRT/Audio/AudioService.h
```diff
@@ -1,17 +1,16 @@
...
 #include <SDL2/SDL_mixer.h>
 #include <string>
 #include <map>
 #include <typeinfo>
-#include "NovelLoggingService.h"
 
-namespace NovelRT {
-  class NovelAudioService {
+namespace NovelRT::Audio {
+  class AudioService {
```
tannergooding: nit: We aren't consistent on whether there should be an extra newline between the namespace/class or not

tannergooding:
Overall LGTM.

I called out a few nits/nice to haves, but didn't repeat them if they occured more than once.

include/NovelRT/Audio/AudioService.h
capnkenny: Hmm...after reading through master, it appears I used it as a value instead of reference type throughout the code base. I believe that I had switched it due to having issues with only using one pointer....however specifics I do not remember at this time 😅. 

Also, I believe we had planned to switch it to reference after the NovelRunner restructure from #87

(That's IF I understood this question correctly >.>)

include/NovelRT/Audio/AudioService.h
RubyNova: Is this something we could enforce with `clang-format`?

CMakeLists.txt
tannergooding: Sure, but that sounds like something that belongs in a `tools` folder :)

CMakeLists.txt
```diff
@@ -63,59 +63,63 @@ find_library(LUA_LIBRARY lua)
...
-  src/NovelLayeringService.cpp
-  src/NovelLoggingService.cpp
-  src/NovelRenderingService.cpp
-  src/NovelRenderObject.cpp
-  src/NovelRunner.cpp
-  src/NovelStepTimer.cpp
-  src/NovelTextRect.cpp
-  src/RGBAConfig.cpp
-  src/NovelCamera.cpp)
+set(NOVELRT_LIBRARY_HEADERS
```
tannergooding: Do we want to try and keep these alphabetized?

include/NovelRT.h
```diff
@@ -0,0 +1,112 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+
+#ifndef NOVELRT_H
+#define NOVELRT_H
+
+#include <math.h>
```
tannergooding: nit: might be nice to sort these. Maybe:
* stdlib includes
* glad includes
* glm includes
* sdl2 includes
* spdlog includes

would also be good to use `cmath` rather than `math.h`

include/NovelRT.h
```diff
@@ -0,0 +1,112 @@
...
+
+namespace NovelRT {
+  typedef void (*NovelUpdateSubscriber)(double deltaSeconds);
+  typedef class DebugService DebugService;
+  typedef class LoggingService LoggingService;
+  typedef class NovelRunner NovelRunner;
+}
+
+namespace NovelRT::Audio {
+  typedef std::map<std::string, Mix_Chunk*> SoundBank;
```
tannergooding: not for this pr, but it would be better to create an actual `SoundBank` type which is a simple wrapper

I'm pretty sure this will be needed for supporting sdl, xaudio, and openal

CC. @capnkenny

tannergooding: (same for the other two below)

include/NovelRT.h
```diff
@@ -0,0 +1,112 @@
...
+  typedef std::map<std::string, Mix_Music*> MusicBank;
+  typedef std::map<std::string, int> ChannelMap;
+  typedef class AudioService AudioService;
+}
+
+namespace NovelRT::Graphics {
+  typedef std::conditional<sizeof(signed long) <= 4, GLuint, GLuint64>::type AdvanceInteger;
```
tannergooding: What is `AdvanceInteger` for again?

include/NovelRT.h
```diff
@@ -0,0 +1,112 @@
...
+#include "NovelRT/Maths/GeoMatrix4.h"
+#include "NovelRT/Maths/GeoBounds.h"
+#include "NovelRT/Transform.h"
+#include "NovelRT/Graphics/GraphicsCharacterRenderData.h"
+#include "NovelRT/Graphics/ShaderProgram.h"
+#include "NovelRT/Graphics/RGBAConfig.h"
+#include "NovelRT/Utilities/Lazy.h"
+#include "NovelRT/Utilities/Misc.h"
+
+#include "NovelRT/LoggingService.h"
```
tannergooding: maybe have a `//base types` comment
and then a `// service types`, `//render object types`, etc (to clarify they are grouped by inheritance heirarchy)?

include/NovelRT/Audio/AudioService.h
```diff
@@ -1,21 +1,14 @@
...
-#include <string>
-#include <map>
-#include <typeinfo>
-#include "NovelLoggingService.h"
-
-namespace NovelRT {
-  class NovelAudioService {
+#ifndef NOVELRT_AUDIOSERVICE_H
+#define NOVELRT_AUDIOSERVICE_H
+#ifndef NOVELRT_H
```
tannergooding: nit: maybe have a newline between the `#define` and `#ifndef`?

include/NovelRT/DebugService.h
```diff
@@ -2,25 +2,23 @@
...
 
-  typedef class NovelRunner NovelRunner;
-
-  class NovelDebugService {
+  class DebugService {
 
   private:
     NovelRunner* _runner;
-    NovelTextRect* _fpsCounter;
+    std::unique_ptr<Graphics::TextRect> _fpsCounter;
```
tannergooding: just a note.

eventually this should *probably* be something added to the scene graph and enabled/disabled by the debug service

include/NovelRT/Graphics/Camera.h
```diff
@@ -0,0 +1,61 @@
...
+#endif
+
+namespace NovelRT::Graphics {
+  class Camera {
+
+    NOVELRT_EVENT(CameraViewChanged, CameraViewChangedEventArgs)
+
+  //TODO: Look Im just trying to make this legal code okay?
+  public:
+    enum class CameraMode : unsigned int {
```
tannergooding: nit: this should probably be its own enum
I also don't think we need an enum to differentiate
instead, we probably just need `static unique_ptr<Camera> createPerspective` and `createOrthographic`.

include/NovelRT/Graphics/Camera.h
```diff
@@ -0,0 +1,61 @@
...
+      return _projectionMatrix;
+    }
+
+    inline void setProjectionMatrix(Maths::GeoMatrix4<float> value) {
+      _cameraUboMatrix.reset();
+      _projectionMatrix = value;
+      raiseCameraViewChanged(CameraViewChangedEventArgs(getCameraUboMatrix()));
+    }
+
+    inline Maths::GeoMatrix4<float> getCameraUboMatrix() { //TODO: I'm unsure about getters like these since its coming from a Lazy. Thoughts?
```
tannergooding: you mean since it's not a ref return?

include/NovelRT/Graphics/ImageRect.h
```diff
@@ -0,0 +1,52 @@
...
+#ifndef NOVELRT_GRAPHICS_IMAGERECT_H
+#define NOVELRT_GRAPHICS_IMAGERECT_H
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Graphics {
+
+  class ImageRect : public RenderObject {
+    friend class TextRect;
```
tannergooding: Why friend rather than making the necessary bits `protected`?

include/NovelRT/Graphics/Camera.h
```diff
@@ -0,0 +1,61 @@
...
+#ifndef NOVELRT_GRAPHICS_CAMERA_H
+#define NOVELRT_GRAPHICS_CAMERA_H
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Graphics {
+  class Camera {
+
+    NOVELRT_EVENT(CameraViewChanged, CameraViewChangedEventArgs)
```
tannergooding: I wonder if it might be better to have an `isModified` flag that says whether or not the camera has changed since last frame...

tannergooding: we already have that, effectively, by checking if `_cameraUboMatrix` has been initialized or if it is "reset"

include/NovelRT/Graphics/ImageRect.h
```diff
@@ -0,0 +1,52 @@
...
+  private:
+    std::string _imageDir;
+    std::string _previousImageDir;
+    Utilities::Lazy<GLuint> _textureId;
+    std::vector<GLfloat> _uvCoordinates;
+    Utilities::Lazy<GLuint> _uvBuffer;
+    Utilities::Lazy<GLuint> _colourTintBuffer;
+    RGBAConfig _colourTint;
+    std::vector<GLfloat> _colourTintData;
+    LoggingService _logger;
```
tannergooding: Given the services are meant to be `unique_ptr` and aren't meant to be cached outside a method call; this shouldn't be a field.

Instead we could either:
* Have `NovelRT` be a "singleton" (you can have one active runner at a time; which you retrieve via a `static NovelRunner* getCurrent()` (which you can then get services from)
-or-
* You could have render objects track the parent render service (since they are tied to the service and need to be recreated if the device is lost anyways); in which case you could walk up to the NovelRunner and then get other services.

include/NovelRT/Graphics/RenderObject.h
```diff
@@ -0,0 +1,36 @@
...
+    virtual void configureObjectBuffers();
+    static GLuint generateStandardBuffer();
+    Maths::GeoMatrix4<float> generateViewData();
+    Maths::GeoMatrix4<float> generateCameraBlock();
+
+    Utilities::Lazy<GLuint> _vertexBuffer;
+    Utilities::Lazy<GLuint> _vertexArrayObject;
+    ShaderProgram _shaderProgram;
+    std::vector<GLfloat> _vertexBufferData;
+    bool _bufferInitialised = false;
```
tannergooding: nit: we should probably initialize fields via the protected constructor

include/NovelRT/Graphics/RenderingService.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Graphics {
+  class RenderingService {
+  private:
+    bool initialiseRenderPipeline();
+    LoggingService _logger;
+    Windowing::WindowingService* const _windowingService;
```
tannergooding: we probably shouldn't cache the windowing service
we do, however, need to know if the window goes down as that invalidates the render service

include/NovelRT/Graphics/RenderingService.h
```diff
@@ -0,0 +1,45 @@
...
+    ShaderProgram _texturedRectProgram;
+    ShaderProgram _fontProgram;
+
+    Utilities::Lazy<GLuint> _cameraObjectRenderUbo;
+    std::unique_ptr<Camera> _camera;
+
+    void bindCameraUboForProgram(GLuint shaderProgramId);
+
+  public:
+    RenderingService(Windowing::WindowingService* const windowingService);
```
tannergooding: we should probably take a reference to the runner that owns the service and then looking the active windowing service from it; so the uniqueness of the pointer stays correct

include/NovelRT/Graphics/TextRect.h
```diff
@@ -1,50 +1,51 @@
...
+namespace NovelRT::Graphics {
+  class TextRect : public RenderObject {
   private:
     void reloadText();
 
     std::string _fontFileDir;
     std::string _previousFontFileDir = "";
     std::string _text = "";
-    std::vector<NovelImageRect*> _letterRects;
+    std::vector<ImageRect*> _letterRects;
```
tannergooding: this should probably be a vector of unique_ptr (although eventually we want a single imagerect anyways)

include/NovelRT/Graphics/TextRect.h
```diff
@@ -1,50 +1,51 @@
...
-      NovelCamera* camera,
+      Camera* camera,
       float fontSize,
       const std::string& fontFileDir,
       const RGBAConfig& colourConfig);
 
     void drawObject() final;
 
-    RGBAConfig getColourConfig() const;
+    inline const RGBAConfig& getColourConfigReadonly() const {
```
tannergooding: I don't think you need to call this `Readonly`

include/NovelRT/Input/InteractionObject.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef NOVELRT_INPUT_INTERACTIONOBJECT_H
+#define NOVELRT_INPUT_INTERACTIONOBJECT_H
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Input {
+  class InteractionObject : public WorldObject {
+    friend class InteractionService; //how to make shit tightly coupled oh god
```
tannergooding: we should log bugs for these comments or consistently prefix with `TODO` (or both) 😄

include/NovelRT/Transform.h
```diff
@@ -0,0 +1,55 @@
...
+    Maths::GeoVector<float> _scale;
+    float _rotation;
+    bool _active = true;
+    int _orderInLayer;
+    int _layer;
+
+  public:
+    Transform(const Maths::GeoVector<float>& position, float rotation, const Maths::GeoVector<float>& scale);
+
+  inline const Maths::GeoVector<float>& getPositionReadonly() const {
```
tannergooding: nit: indentation

include/NovelRT/Utilities/Misc.h
```diff
@@ -1,19 +1,21 @@
...
-  static inline const std::string CONSOLE_LOG_APP = "Application";
-  static inline const std::string CONSOLE_LOG_GFX = "GFX";
-  static inline const std::string CONSOLE_LOG_AUDIO = "Audio";
-  static inline const std::string CONSOLE_LOG_INPUT = "Input";
-};
+namespace NovelRT::Utilities {
+  class Misc {
+  public:
+    static const int SDL_SUCCESS = 0;
+    static inline const std::string CONSOLE_LOG_GENERIC = "NovelRT";
```
tannergooding: `inline const`?

src/NovelRT/Graphics/BasicFillRect.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+  BasicFillRect::BasicFillRect(const Transform& transform,
+    int layer,
+    Camera* camera,
+    ShaderProgram shaderProgram,
+    const RGBAConfig& fillColour) :
+    RenderObject(transform, layer, shaderProgram, camera), _colourConfig(fillColour),
+    _colourBuffer(Utilities::Lazy<GLuint>(generateStandardBuffer)) {}
+
+  void BasicFillRect::drawObject() {
```
tannergooding: just a note: a lot of this is consistent between the different render objects and could be abstracted
e.g. the real differentiation is the shader programs being set and the number of vertices/indices

src/NovelRT/Graphics/ImageRect.cpp
```diff
@@ -0,0 +1,170 @@
...
+        glDisableVertexAttribArray(2);
+        glDisableVertexAttribArray(1);
+        glDisableVertexAttribArray(0);
+        glBindVertexArray(0);
+      }
+
+      void ImageRect::configureObjectBuffers() {
+        RenderObject::configureObjectBuffers();
+
+        _uvCoordinates = {
```
tannergooding: just a note: a lot of this could be instanced and shared across multiple instances (considering its all "unit sized" and then scaled to the correct size by the shader)

src/NovelRT/Graphics/RenderObject.cpp
```diff
@@ -0,0 +1,89 @@
...
+    }
+    drawObject();
+  }
+
+  void RenderObject::configureObjectBuffers() {
+    auto topLeft = Maths::GeoVector<GLfloat>(-0.5f, 0.5f);
+    auto bottomRight = Maths::GeoVector<GLfloat>(0.5f, -0.5f);
+    auto topRight = Maths::GeoVector<GLfloat>(0.5f, 0.5f);
+    auto bottomLeft = Maths::GeoVector<GLfloat>(-0.5f, -0.5f);
+    _vertexBufferData = {
```
tannergooding: same thing, this could be shared across multiple instances.

src/NovelRT/Input/InteractionObject.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <NovelRT.h>
+
+namespace NovelRT::Input {
+  InteractionObject::InteractionObject(const Transform& transform, int layer, const std::function<void(InteractionObject*)> notifyHasBeenDrawnObject)
+    : WorldObject(transform, layer) {
+    _notifyHasBeenDrawnObject = notifyHasBeenDrawnObject;
+  }
+
+  void InteractionObject::executeObjectBehaviour() {
+    _notifyHasBeenDrawnObject(this);
```
tannergooding: what is the purpose of `notifyHasBeenDrawn`?

CMakeLists.txt
RubyNova: I can reorder this to be alphabetical if you want, I just threw stuff on the end as I added.

include/NovelRT.h
RubyNova: This is for the non-compliance of integer sizes and text character information. On different platforms, this will compile to the correct numeric type for that platform.

include/NovelRT/DebugService.h
RubyNova: We can subscribe to the scene construction. I'll do that now. Forgot. Oops lol.

include/NovelRT/Graphics/Camera.h
RubyNova: Correct. This is returning from the lazy but do we forward that by reference?

include/NovelRT/Graphics/ImageRect.h
RubyNova: I'm unsure how that would achieve the same effect.

TextRect is invoking an `Internal` method on ImageRect. How would `protected` support this?

include/NovelRT/Graphics/ImageRect.h
RubyNova: I think this is a flaw with the current API design for the logging service AFAIK.

I'll see what I can come up with.

include/NovelRT/Graphics/RenderingService.h
RubyNova: What do you propose as an alternative?

include/NovelRT/Input/InteractionObject.h
RubyNova: Do we even still need this with the centralised header?

include/NovelRT/Utilities/Misc.h
RubyNova: That was not me but I'll rectify, lol.

src/NovelRT/Input/InteractionObject.cpp
RubyNova: This is how the service and the interaction rects talk to each other. The rect tells the service its been drawn, which triggers the service to manage if its been "clicked on", or not.

include/NovelRT/Graphics/TextRect.h
RubyNova: Oh, alright. Then what do we call it? I thought it needed a different name or the compiler would screech. I also like the clarity.

FWIW, I have used this in other places for consistency.

include/NovelRT/Utilities/Misc.h
RubyNova: Uh, hold on, maybe I am misunderstanding. What is wrong with this?

src/NovelRT/Graphics/BasicFillRect.cpp
RubyNova: Considering our approach to restructure buffer usage, I think this can wait until then. I'd rather not break things any further.

src/NovelRT/Graphics/ImageRect.cpp
RubyNova: Same as before, I think we should look to do these sorts of changes alongside the buffer overhaul.

src/NovelRT/Graphics/RenderObject.cpp
RubyNova: Same as before, I think we should look to do these sorts of changes alongside the buffer overhaul.

include/NovelRT.h
tannergooding: I wonder if we should abstract that behind something....

If its just the `advance` (distance from this character to the next character in pixels) then 4 bytes is a reasonable size to expose (as we should never have a distance of > 4 billion pixels 😄).

We could also just expose it as `signed long`, but that puts the burden on the end user to normalize.

tannergooding: (We probably also don't want to be tied to `GL*` types, for when we eventually support Vulkan and others).

include/NovelRT/Graphics/Camera.h
tannergooding: We could have the lazy expose a `T&` (and still ensure it is a valid one by enforcing initialization).

include/NovelRT/Graphics/ImageRect.h
tannergooding: Which internal method?

I'm trying to think about how this could be configured (probably in a future PR) so users could define their own `RenderObject` types if desired (we aren't very extensible if only we can define specialized render objects).

include/NovelRT/Graphics/ImageRect.h
tannergooding: > I think this is a flaw with the current API design for the logging service AFAIK.

Right, but I think it also applies to the other services. Many of them could require recreation under various scenarios and so we probably don't want users caching bits themselves.

include/NovelRT/Graphics/RenderingService.h
tannergooding: Similar to https://github.com/novelrt/NovelRT/pull/98#discussion_r362263945. We should have some way the user can get the current `runner` and be able to get services from that.

The rendering service would then retrieve the window service on construction and subscribe to the resize and close notifications (so it can update itself as appropriate).

include/NovelRT/Graphics/TextRect.h
tannergooding: You can just give it the same name and the compiler picks the correct overload for the use-case.
Types in the stdlib do the same thing: <http://www.cplusplus.com/reference/vector/vector/at/>

include/NovelRT/Utilities/Misc.h
tannergooding: Why is it both `inline` and `const`?

include/NovelRT/Graphics/ImageRect.h
RubyNova: there's a private method that allows for the texture to be reassigned in an `ImageRect` with a buffer handle, which only `TextRect` uses. Specifically, `setTextureInternal`.

As for custom `RenderObject` types, there's nothing stopping them inheriting `RenderObject` and implementing them as we've done with our default ones.

include/NovelRT/Utilities/Misc.h
tannergooding: Ah, I see; its a new C++ 17 feature.

src/NovelRT/Graphics/ImageRect.cpp
tannergooding: Right, was just calling stuff out as a I reviewed the code (hence `just a note` 😄)

include/NovelRT/Graphics/ImageRect.h
tannergooding: Right, but users can't (for example) inherit `ImageRect` and create their own `TextRect`.

include/NovelRT/Graphics/TextRect.h
RubyNova: Alright, then we'll do this if you believe this to be clearer. If the stdlib does it then I assume people know what they can and can't do based on the context of being `const` or not.

include/NovelRT/Graphics/ImageRect.h
capnkenny: Wouldn't reworking this be a part of #87 ?

include/NovelRT/Graphics/ImageRect.h
RubyNova: I see, so you reckon by making it `protected` they could also use this internal method for their own extensibility purposes?

include/NovelRT/Graphics/ImageRect.h
tannergooding: Probably; 90% of my comments are things I'm fine with being addressed in a follow up PR. But things that I wanted to call out as I was reviewing the code.

capnkenny: Makes sense - I also just wanted to remind that we do have an issue open so that we can refer back to this when working through it 👍

RubyNova: So do I rework this now or not? I'm confused.

include/NovelRT/Utilities/Misc.h
RubyNova: Okay, marking as resolved then.

include/NovelRT/Graphics/ImageRect.h
tannergooding: Right. `friend class` is basically C# `internal` (but with access to `private` members as well) so it tightly couples bits together and prevents external users from doing the same.

In general, it would be good to only use it where really necessary and try to design with `protected` and `public` members instead.

RubyNova: okay, but I didn't think `protected` resolves me completely removing `friend class` in this context does it? I thought `protected` meant only this type and inheritors?

include/NovelRT/Graphics/ImageRect.h
tannergooding: It can be reworked later. It, like many of the the items I called out, are mostly just observations noted while reviewing the code.

They should be resolved eventually (or at least discussed and we reach consensus on how to handle it or not) but don't have to be resolved "now".

include/NovelRT/Graphics/ImageRect.h
tannergooding: I was thinking the member should probably be public -or- there should be `TextureRect` class which `ImageRect` inherits from.

There are multiple scenarios where devs may want to set a texture without having a physical image on disk first (including procedural generated data)

tannergooding: (also, to clarify, this could be a small follow up PR)

RubyNova: Okay, I'll make an issue.

include/NovelRT/Audio/AudioService.h
RubyNova: Propagated to #101 . Resolving.

include/NovelRT.h
RubyNova: Well, the plan was I'd deal with that when we worked on a different render impl as I cannot think of a "nice" way to do it right now. Do you have a nice solution for this right now? If not I'll propagate it to then as I originally planned. 😄

include/NovelRT/Graphics/ImageRect.h
RubyNova: Propagated to #100 . Resolving.

include/NovelRT/Graphics/ImageRect.h
RubyNova: Propagated to #101 . Resolving.

include/NovelRT.h
tannergooding: My suggestion is to basically make the `std::conditional` an internal implementation detail (as it is only needed for ogl anyways).

We would then decide to expose a consistent size to the end user. For example, we could choose:
* `int32_t`, should be big enough for all cases, will need extension on Unix 64-bit
* `int64_t`, will always be big enough, will need downcast on Windows and Unix 32-bit
* `intptr_t`, will always be big enough, will need downcast on Windows 64-bit
* `signed long`, exactly the right size, user will need to rationalize platform differences

include/NovelRT/Graphics/Camera.h
RubyNova: Should I delete the enum then?

include/NovelRT/Graphics/Camera.h
RubyNova: Just checked the header - seems `getActual` is already returning a `T&`. I'll just forward it then. Awesome. 😄

include/NovelRT/Graphics/Camera.h
RubyNova: Enum deleted. I'll look into implementing these. :)

include/NovelRT.h
RubyNova: Propagated to #102 . Resolving.

include/NovelRT/Graphics/Camera.h
RubyNova: I'm making a ticket for the perspective view. Orthographic has been implemented. Resolving.

src/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ set(NOVELRT_LIBRARY_HEADERS
   ../include/NovelRT/Utilities/Misc.h
   ../include/NovelRT/Windowing/WindowingService.h)
 
-  source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${NOVELRT_LIBRARY_HEADERS})
+  source_group(TREE ../ FILES ${NOVELRT_LIBRARY_HEADERS})
```
tannergooding: I think you want ` ${CMAKE_CURRENT_SOURCE_DIR}/../`

RubyNova: Ah ok, I did wonder. Thanks.

CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
 set(CMAKE_CXX_STANDARD 17)
 set(OpenGL_GL_PREFERENCE GLVND)
 
+add_subdirectory(src)
```
tannergooding: What's the output layout in this case?

Should we specify `binary_dir` so we can separate product output from test/samples output?

CMakeLists.txt
```diff
@@ -38,132 +41,23 @@ set(NOVELRT_RESOURCES
...
-find_package(sdl2-mixer CONFIG REQUIRED)
-find_package(spdlog CONFIG REQUIRED)
-
-# lua doesn't provide native cmake support
-find_path(LUA_INCLUDE_DIR lua.h)
-include_directories(${LUA_INCLUDE_DIR})
-
-find_library(LUA_LIBRARY lua)
-link_libraries(${LUA_LIBRARY})
+include_directories(src)
```
tannergooding: Why do we need to search `src` for header files?

CMakeLists.txt
```diff
@@ -38,132 +41,23 @@ set(NOVELRT_RESOURCES
...
-      ${NOVELRT_LINK_LIBRARIES}
-      m
-    )
-endif()
-
-set(NOVELRT_EXECUTABLE_HEADERS
-  ${NOVELRT_HEADERS}
-)
-
 set(NOVELRT_EXECUTABLE_SOURCES
```
tannergooding: Shouldn't this (and others) be moved to the respective projects rather than be in the root?

samples/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(NovelRTTest VERSION 0.0.1)
+include(GNUInstallDirs)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+set(CMAKE_CXX_STANDARD 17)
```
tannergooding: We should centralize and import several of these options, rather than having to redeclare them in each `CMakeLists.txt`

CMakeLists.txt
RubyNova: Its the default at this moment in time, so currently its `out/build/x64-Debug/samples` for this target. It copies from the specified `bin` folder of the lib project output for dynamic linkage.

I think as a result the samples program(s) is/are already separated?

samples/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(NovelRTTest VERSION 0.0.1)
```
tannergooding: Why `Test` rather than `NovelRTSamples`?

CMakeLists.txt
RubyNova: This is for Doxygen, afaik. I've not touched that logic yet so I just put that line in to ensure it had access (it probably doesn't need it but still).

CMakeLists.txt
RubyNova: I think any remnants are Doxygen specific/have just not been cleaned out yet or I mistook them for shared settings such as `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`.

CMakeLists.txt
tannergooding: Yep, that looks to be the case. Sounds good to me.

samples/CMakeLists.txt
RubyNova: Because that is the current name of the window that is produced, and it it was 3 AM when I wrote this 🤣 I'll modify it.

samples/CMakeLists.txt
RubyNova: What would be the approach for doing this? Would we have another CMake file that contains these settings or?

PR #81, opened 2019-10-27
Basic file loading service

rob5300:
This is to introduce the basic structure idea of the Asset loader.
I have replaced an instance of a shader load from disk with the loader showing how it would be used.
I have modified main to get the launch path, this will probably need changes to fit styling elsewhere.
this PR resolves #24

EDIT: Corrected ticket number - Ruby.

PR #107, opened 2020-01-08
Cleaning up the CMakeLists.txt

tannergooding:
This cleans up the CMakeLists.txt so we have distinct `src` and `samples` projects and does some minimal cleanup so that resources can be resolved relative to the host executable.


tannergooding:
(I expect I have some cleanup for Unix still)

CMakeLists.txt
```diff
@@ -1,277 +1,11 @@
 cmake_minimum_required(VERSION 3.13)
```
tannergooding: The root project defines the minimum version and central options (such as the `CXX_STANDARD`)

include/NovelRT.h
```diff
@@ -3,24 +3,25 @@
...
 #define NOVELRT_H
 
-//stdlib
+//cstdlib
 #include <cmath>
-#include <type_traits>
+#include <cstdint>
+
+//stdlib
+#include <algorithm>
```
tannergooding: Sorted these alphabetically and removed duplicates

include/NovelRT/Utilities/Misc.h
```diff
@@ -17,6 +17,12 @@ namespace NovelRT::Utilities {
     static inline const std::string CONSOLE_LOG_AUDIO = "Audio";
     static inline const std::string CONSOLE_LOG_INPUT = "Input";
     static inline const std::string CONSOLE_LOG_WINDOWING = "WindowManager";
+
+    static std::filesystem::path getExecutablePath();
```
tannergooding: Small helper function for resolving the path to the host executable

samples/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+set(NOVELRT_SAMPLES_HEADERS)
```
tannergooding: The samples project just includes `main.cpp`, references `libNovelRT`, and copies over its output.

samples/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+source_group(TREE ${CMAKE_SOURCE_DIR} FILES ${NOVELRT_SAMPLES_HEADERS})
+
+file(GLOB_RECURSE NOVELRT_SAMPLES_SOURCES CONFIGURE_DEPENDS *.cpp)
```
tannergooding: @RubyNova, I know you said you had problems with `GLOB`; but from what I can tell `CONFIGURE_DEPENDS` ensures that the configuration is updated if the glob changes (i.e. if you add a new file to the disk).

samples/main.cpp
```diff
@@ -42,22 +42,29 @@ std::unique_ptr<NovelRT::Input::BasicInteractionRect> memeInteractionRect;
 
 int main(int argc, char *argv[])
 {
+  std::filesystem::path executableDirPath = NovelRT::Utilities::Misc::getExecutableDirPath();
```
tannergooding: This just uses the C++ 17 APIs to combine paths. Ideally, we would have a centralized asset manager that helps manage and resolve these paths (e.g. https://github.com/novelrt/NovelRT/pull/81).

samples/main.cpp
```diff
@@ -42,22 +42,29 @@ std::unique_ptr<NovelRT::Input::BasicInteractionRect> memeInteractionRect;
...
-  luaL_dofile(L, "avg.lua");
+  luaL_dofile(L, (scriptsDirPath / "avg.lua").string().c_str());
   lua_close(L);
   auto runner = NovelRT::NovelRunner(0, "NovelRTTest");
   auto console = NovelRT::LoggingService(NovelRT::Utilities::Misc::CONSOLE_LOG_APP);
 
   auto novelChanTransform = NovelRT::Transform(NovelRT::Maths::GeoVector<float>(1920 / 2, 1080 / 2), 2, NovelRT::Maths::GeoVector<float>(456, 618));
 
-  novelChanRect = runner.getRenderer()->createImageRect(novelChanTransform, 3, "novel-chan.png", NovelRT::Graphics::RGBAConfig(255, 0, 255, 255));
+  novelChanRect = runner.getRenderer()->createImageRect(novelChanTransform, 3, (imagesDirPath / "novel-chan.png").string(), NovelRT::Graphics::RGBAConfig(255, 0, 255, 255));
```
tannergooding: it would probably be helpful if we took `std::filesystem::path` where we want a file path. It would avoid the need for consumers of the API to call `.string()`

samples/main.cpp
```diff
@@ -42,22 +42,29 @@ std::unique_ptr<NovelRT::Input::BasicInteractionRect> memeInteractionRect;
...
+  std::filesystem::path fontsDirPath = resourcesDirPath / "Fonts";
+  std::filesystem::path imagesDirPath = resourcesDirPath / "Images";
+  std::filesystem::path scriptsDirPath = resourcesDirPath / "Scripts";
+
   //setenv("DISPLAY", "localhost:0", true);
   L = luaL_newstate();
   luaL_openlibs(L);
   lua_register(L, "average", average);
-  luaL_dofile(L, "avg.lua");
+  luaL_dofile(L, (scriptsDirPath / "avg.lua").string().c_str());
```
tannergooding: We currently include this script as part of `libNovelRT`, but it (in particular) should likely be exclusively an asset to `NovelRTSamples`

src/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
+find_package(Freetype REQUIRED)
```
tannergooding: We don't define `project` or other items as these aren't meant to be built outside the root cmake file

src/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_IMAGE_RESOURCES} $<TARGET_FILE_DIR:libNovelRT>/Resources/Images
+  COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:libNovelRT>/Resources/Scripts
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_SCRIPT_RESOURCES} $<TARGET_FILE_DIR:libNovelRT>/Resources/Scripts
+  COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:libNovelRT>/Resources/Shaders
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_SHADER_RESOURCES} $<TARGET_FILE_DIR:libNovelRT>/Resources/Shaders
+)
+
+# Docs
+
+if (DOXYGEN_FOUND)
```
tannergooding: I don't think we need doxygen for things like samples, but its trivial to pull this out and share if we wanted (basically just move this stuff to a `Doxygen.cmake` file and then `include(Doxygen.cmake)` from any project wanting docs).

src/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+
+add_library(libNovelRT SHARED ${NOVELRT_LIB_HEADERS} ${NOVELRT_LIB_SOURCES})
+target_link_libraries(libNovelRT ${NOVELRT_LIB_LINK_LIBRARIES})
+
+file(GLOB_RECURSE NOVELRT_LIB_FONT_RESOURCES CONFIGURE_DEPENDS Resources/Fonts/*.ttf)
+file(GLOB_RECURSE NOVELRT_LIB_IMAGE_RESOURCES CONFIGURE_DEPENDS Resources/Images/*.png)
+file(GLOB_RECURSE NOVELRT_LIB_SCRIPT_RESOURCES CONFIGURE_DEPENDS Resources/Scripts/*.lua)
+file(GLOB_RECURSE NOVELRT_LIB_SHADER_RESOURCES CONFIGURE_DEPENDS Resources/Shaders/*.glsl)
+
+add_custom_command(TARGET libNovelRT POST_BUILD
```
tannergooding: Notably, rather than dumping all assets to the root output directory, this mirrors the directory structure (e.g. `<out>/Resources/Images/novel-chan.png`).

This helps organize the output and isolate assets from eachother.

src/NovelRT/Utilities/Misc.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+      path[pathLength] = L'\0';
+      return path;
+#else
+      char path[PATH_MAX + 1];
+      auto pathLength = readlink("/proc/self/exe", path, PATH_MAX);
+
+      if ((pathLength < 0) || (pathLength == PATH_MAX))
+      {
+        // TODO: Get command line from /proc/self/cmdline
```
tannergooding: I don't know if Unix has an API like `CommandLineToArgv()` which is what is blocking this bit.

samples/main.cpp
```diff
@@ -71,7 +78,7 @@ int main(int argc, char *argv[])
   auto vec = playButtonTransform.getPosition();
   vec.setX(playButtonTransform.getPosition().getX() - 75);
   playAudioTextTransform.setPosition(vec);
-  playAudioText = runner.getRenderer()->createTextRect(playAudioTextTransform, 1, NovelRT::Graphics::RGBAConfig(0, 0, 0, 255), 36, "Gayathri-Regular.ttf");
+  playAudioText = runner.getRenderer()->createTextRect(playAudioTextTransform, 1, NovelRT::Graphics::RGBAConfig(0, 0, 0, 255), 36, (fontsDirPath / "Gayathri-Regular.ttf").string());
   playAudioText->setText("Play Audio");
```
tannergooding: Not visible here, but further down the audio service only takes in a file name (such as `sparta.wav`): https://github.com/novelrt/NovelRT/pull/107/files#diff-ef810af08ed542e7c7cf6edddb5bd627R113)

Ideally, we would have the asset manage track the full paths and we'd be able to internally track these via custom ids (ints, strings, or something else). That way, you can have `<custom-path>/sparta.wav` as the loaded asset, but still access it via `sparta.wav`

scripts/machine-setup.sh
```diff
@@ -5,7 +5,7 @@ sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'
 sudo apt-add-repository 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main'
 
 sudo apt-get update
-sudo apt-get install clang-8 cmake yasm -y
+sudo apt-get install clang-8 cmake g++-8 yasm -y
```
tannergooding: Some C++17 features, like `filesystem`, weren't part of `libstdc++` until `G++ 8`

src/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+  SDL2::SDL2_mixer
+  SDL2::SDL2main
+  spdlog::spdlog
+)
+
+if(NOT WIN32)
+  set(NOVELRT_LIB_LINK_LIBRARIES
+    ${NOVELRT_LIB_LINK_LIBRARIES}
+    m
+    stdc++fs
```
tannergooding: Needed on Unix for `filesystem`

src/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+  )
+endif()
+
+if (WIN32)
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
+
+add_library(NovelRT SHARED ${NOVELRT_LIB_HEADERS} ${NOVELRT_LIB_SOURCES})
```
tannergooding: Unix automatically prepends `lib` to the target name, so we want just `NovelRT`. This gives us `NovelRT.dll`/`NovelRT.lib` on Windows and `libNovelRT.so`/libNovelRT.a` on Unix. We were getting `libNovelRT.dll`/`libNovelRT.lib` and `liblibNovelRT.so`/`liblibNovelRT.a` which I believe is undesirable.

capnkenny: I can confirm as well - this was the output being delivered by Emscripten on WSL Ubuntu. Leaving here so Matt can review :)

RubyNova:
I've merged in a fix to this branch for the worldobject active state as discussed in the discord. I'll just quickly verify everything is working but I don't see why it wouldn't, the fixes were minor.

src/CMakeLists.txt
RubyNova: I am happy with this. Resolving.

src/NovelRT/Graphics/TextRect.cpp
```diff
@@ -152,4 +155,9 @@ namespace NovelRT::Graphics {
       ptr->setActive(false);
       });
   }
+
+  void TextRect::setActive(bool value) {
+    WorldObject::setActive(value);
+    std::for_each(_letterRects.begin(), _letterRects.end(), [&value](auto& x){x->setActive(value);});
```
tannergooding: Do we need to set each letter as active? Shouldn't the textrect being active/inactive impact the individual letters at a higher point (such that leaving the letters always active is fine)

PR #112, opened 2020-01-10
Port engine over to GLFW/OpenAL

### LINKED ISSUE #105: Migrate from SDL2 to GLFW.
As per #104 and #106 , SDL2 is breaking in multiple places that are completely unexpected, at least on windows.

It appears modern GLFW may work better for our needs.

This means replacing things such as sdl_image with libpng, but I don't personally mind that as one of the main developers working on the graphics pipelines. It also has better support for Vulkan OOTB, it seems, unlike SDL2.

I am very much for this, given the duct tape in #98 to get around one of these issues.

capnkenny:
This resolves #105 .
What's to be done:

- [x] SDL Removed from project
- [x] Switched to GLFW for Windowing/Rendering/Input
- [x] Switched to libpng for images
- [x] Switched to OpenAL for Audio


scripts/machine-setup.sh
```diff
@@ -5,7 +5,7 @@ sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'
 sudo apt-add-repository 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main'
 
 sudo apt-get update
-sudo apt-get install clang-8 cmake g++-8 yasm -y
+sudo apt-get install clang-8 cmake g++-8 yasm libxinerama-dev libxcursor-dev xorg-dev libglu1-mesa-dev -y
```
capnkenny: nit: Alphabetize this @capnkenny pls :)

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -4,16 +4,24 @@
...
 namespace NovelRT::Audio {
 
-AudioService::AudioService() {
-  initializeAudio();
-}
-
-bool AudioService::initializeAudio() {
-  return false;
+AudioService::AudioService() : _logger(LoggingService(Utilities::Misc::CONSOLE_LOG_AUDIO)) {
+  auto device = alcOpenDevice(NULL);
```
tannergooding: nit: Everything else is slightly lazy

That is, creating the service doesn't necessarily create all the relevant contexts/devices. Instead they are created when first needed.

This avoids issues where the audio service failing to initialize doesn't bring down the game, for example.

capnkenny: Ahh so you're saying that the creation should still call an `initialize` function so that even if it does fail, its not killing everything? (The `initialize` function would actually create the contexts/devices as needed when called.)

capnkenny: Tbh, not sure what "slightly lazy" actually means in most contexts

tannergooding: Most of the other services are using `Utilities::Lazy<T>` so that the bits are created on first access (and so they can be easily recreated if a device is lost, for example).

src/NovelRT/Timing/StepTimer.cpp
```diff
@@ -7,9 +7,9 @@
 
 namespace NovelRT::Timing {
   StepTimer::StepTimer(uint32_t targetFrameRate, double maxSecondDelta) :
-    _frequency(SDL_GetPerformanceFrequency()),
+    _frequency(glfwGetTimerFrequency()),
```
capnkenny: This fails and defaults to 0 as GLFW is not initialized yet - refer to the initialization order in `NovelRunner.cpp`.

src/NovelRT/NovelRunner.cpp
```diff
@@ -13,6 +13,7 @@ namespace NovelRT {
     _novelDotNetRuntimeService(std::make_unique<DotNet::RuntimeService>()),
     _novelRenderer(std::make_unique<Graphics::RenderingService>(this)) {
     _novelWindowingService->initialiseWindow(displayNumber, windowTitle);
+    _stepTimer.initializeValues();
```
capnkenny: Added a function to StepTimer to initialize the `_frequency` and `_lastCounter` values.

src/NovelRT/Timing/StepTimer.cpp
```diff
@@ -21,17 +19,21 @@ namespace NovelRT::Timing {
...
     _secondCounter = 0;
     _remainingTicks = 0;
     _framesPerSecond = 0;
     _framesThisSecond = 0;
   }
 
   void StepTimer::tick(const std::vector<NovelUpdateSubscriber>& update) {
-
-    auto currentCounter = glfwGetTime();
+    auto currentCounter = glfwGetTimerValue();
```
capnkenny: We were incorrectly getting the value in seconds, and not the raw tick count. (Thanks @tannergooding for explaining this one)

src/NovelRT/Timing/StepTimer.cpp
```diff
@@ -7,9 +7,7 @@
 
 namespace NovelRT::Timing {
   StepTimer::StepTimer(uint32_t targetFrameRate, double maxSecondDelta) :
-    _frequency(glfwGetTimerFrequency()),
     _maxCounterDelta((uint64_t)(_frequency * maxSecondDelta)),
-    _lastCounter(glfwGetTime()),
```
capnkenny: This was removed, but needs to be corrected.

Due to `glfw` initializing _after_ the StepTimer, these were getting a value of 0, so they were moved to an initializing function. However, the StepTimer should probably be made lazily instead, so that it can be called after `glfw` is initialized in `NovelRunner` properly.

include/NovelRT/Timing/StepTimer.h
```diff
@@ -13,7 +13,7 @@
 namespace NovelRT::Timing {
   class StepTimer {
   private:
-    const uint64_t _frequency;
+    uint64_t _frequency;
```
capnkenny: This poses a potential codegen issue. If StepTimer is made lazy, we can put this back to being `const`.

src/NovelRT/Timing/StepTimer.cpp
capnkenny: Based on this, we _could_ also move the `glfwInit()` call and its applicable error check out of `WindowingService` and make it the earliest call in `NovelRunner` to help avoid issues of this failing.

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -334,6 +336,13 @@ AudioService::~AudioService() {
   alDeleteSources(1, &_soundSource);
   alDeleteSources(1, &_musicSource);
 
+  for (auto it = _sounds.begin(); it != _sounds.end(); it++) {
+    unload(it, false);
+  }
+  for (auto it = _music.begin(); it != _music.end(); it++) {
+    unload(it, true);
+  }
+
```
capnkenny: Is this something that we should keep? From testing, it appears that something is making OpenAL automatically delete the buffers that are created, but its not sure it's the dtor since it seems like the dtor is never being called outright. Unless the logging service is killed prior to OpenAL destruction, the logging calls that were introduced to delete the buffers is never called, hence me thinking that this dtor isn't called.

If it's unnecessary because the dtor is implicitly called elsewhere, or because having the lazy context get destroyed destroys the buffers, I'll remove this since it's redundant.

RubyNova: Assuming they are needed - these should most likely be `for(auto sound : _sounds)` and `for(auto musicObject : _music)` accordingly.

Alternatively, you can use `std::foreach`.

include/NovelRT.h
```diff
@@ -58,9 +65,8 @@ namespace NovelRT {
 }
 
 namespace NovelRT::Audio {
-  typedef std::map<std::string, Mix_Chunk*> SoundBank;
-  typedef std::map<std::string, Mix_Music*> MusicBank;
-  typedef std::map<std::string, int> ChannelMap;
+  typedef std::vector<ALuint> SoundBank;
```
tannergooding: Just note for the audio branch...

This should become an explicit type, that way we aren't tied to `std::vector` nor to `ALuint`

include/NovelRT/Audio/AudioService.h
```diff
@@ -11,51 +11,43 @@
...
     bool initializeAudio();
-    void load(std::string input, bool isMusic);
-    void unload(std::string input, bool isMusic);
-    void playSound(std::string soundName, int loops);
-    void stopSound(std::string soundName);
-    void setSoundVolume(std::string soundName, float val);
-    void setSoundPosition(std::string soundName, int angle, int distance);
-    void setSoundDistance(std::string soundName, int distance);
-    void setSoundPanning(std::string soundName, int leftChannel, int rightChannel);
+    std::vector<ALuint>::iterator load(std::string input, bool isMusic);
```
tannergooding: Similar note for audio branch...

We probably don't want to tie ourselves to `ALuint` here in the public API.

src/CMakeLists.txt
```diff
@@ -15,14 +16,19 @@ source_group(TREE ${CMAKE_SOURCE_DIR} FILES ${NOVELRT_LIB_SOURCES})
...
-  SDL2::SDL2main
+  OpenAL::OpenAL
+  png
   spdlog::spdlog
 )
 
+if (WIN32)
+  set(NOVELRT_LIB_LINK_LIBRARIES ${NOVELRT_LIB_LINK_LIBRARIES} sndfile-shared)
+else()
+  set(NOVELRT_LIB_LINK_LIBRARIES ${NOVELRT_LIB_LINK_LIBRARIES} sndfile-static)
```
tannergooding: Why statically link on Unix?

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -3,363 +3,349 @@
 #include <NovelRT.h>
 
 namespace NovelRT::Audio {
-
-AudioService::AudioService() : _nextChannel(1), _musicTime(0), _musicPausedTime(0), isInitialized(false), _logger(Utilities::Misc::CONSOLE_LOG_AUDIO) {
-  initializeAudio();
+AudioService::AudioService() : _device(Utilities::Lazy<std::unique_ptr<ALCdevice, void(*)(ALCdevice*)>> (std::function<ALCdevice*()>([this] {
+    auto device = alcOpenDevice((_deviceName.empty())? nullptr : _deviceName.c_str());
+    if (!device) {
+      throw std::runtime_error("OpenAL: Could not get audio devices!");
```
tannergooding: No log?

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -3,363 +3,349 @@
...
-    }
-    else
-    {
-      _logger.logError("SDL_Mixer error occurred during load. Error: ", getSDLError());
-    }
-  }
-}
+  std::vector<uint16_t> data;
+  std::vector<short> readBuffer;
+  readBuffer.resize(4096);
```
tannergooding: Why 4096? Also the default buffer size should probably be a defined constant.

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -3,363 +3,349 @@
...
-    {
-      return;
-    }
+/*Note: Due to the current design, this will currently block the thread it is being called on.
+  If it is called on the main thread, please do all loading of audio files at the start of
+  the engine (after NovelRunner has been created).
+*/
+std::vector<ALuint>::iterator AudioService::load(std::string input, bool isMusic) {
+  if (!isInitialised) {
+    _logger.logError("Cannot load new audio into memory while the service is uninitialised! Aborting...");
```
tannergooding: For the audio branch, I wonder if we should do something similar to the render API, where we have explicit buffers and audio objects.

src/CMakeLists.txt
capnkenny: This is actually a requirement of libsndfile (used for reading the different file formats). I'd need to look into it more, but vcpkg specifies to use static linking on Unix - dynamic linking fails (it caused a CI failure for the Ubuntu checks prior).

include/NovelRT/Input/KeyCode.h
```diff
@@ -9,165 +9,132 @@
...
+    NumPadNine = GLFW_KEY_KP_9,
+    NumPadZero = GLFW_KEY_KP_0,
+    NumPadFullStop = GLFW_KEY_KP_DECIMAL,
+    NumPadEquals = GLFW_KEY_KP_EQUAL,
+
+    MenuKey = GLFW_KEY_MENU,
+
+    //The next two should only be used when checking MouseEvents
+
+      LeftMouseButton = GLFW_MOUSE_BUTTON_LEFT,
```
RubyNova: Nit: This enum has inconsistent indentation.

src/NovelRT/Graphics/RenderingService.cpp
```diff
@@ -20,10 +15,9 @@ namespace NovelRT::Graphics {
 
     if (completeLaunch) {
       _camera = Camera::createDefaultOrthographicProjection(windowSize);
-      _openGLContext = SDL_GL_CreateContext(_runner->getWindowingService()->getWindow());
-      SDL_GL_MakeCurrent(_runner->getWindowingService()->getWindow(), _openGLContext);
+      glfwMakeContextCurrent(_runner->getWindowingService()->getWindow());
 
-      if (!gladLoadGL()) {
+      if (!gladLoadGLLoader((GLADloadproc) glfwGetProcAddress)) {
```
RubyNova: C-style cast should be replaced with the appropriate C++-style cast, or the compiler is going to guess/assume which is dangerous.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -4,42 +4,80 @@
...
-    float hData = displayData.h * 0.7f;
-    auto window = SDL_CreateWindow(
-      windowTitle.c_str(), SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED,
-      wData, hData, SDL_WINDOW_OPENGL | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE);
-
-    _window = std::unique_ptr<SDL_Window, decltype(&SDL_DestroyWindow)>(window, SDL_DestroyWindow);
-    _windowSize = Maths::GeoVector<float>(wData, hData);
+    float wData = displayData->width * 0.7f;
+    float hData = displayData->height * 0.7f;
+    auto window = glfwCreateWindow(wData, hData, _windowTitle.c_str(), NULL, NULL);
```
RubyNova: Please fix this to not use C-style `NULL`, it looks out of place

src/NovelRT/Audio/AudioService.cpp
capnkenny: This was changed to 2048 and put behind a const.
4096 was a random guess, but 2048 seems to be a happy medium between fast loading and increased memory usage - for the new audio API, we would want to discern between files needing a larger amount of buffer space (meaning a larger buffer size), and files needing a smaller amount (like sfx, which would need a smaller buffer size so that we don't allocate useless space).

src/NovelRT/Windowing/WindowingService.cpp
capnkenny: The NovelRT solution was checked for `NULL`s and changed _most_ to `nullptr`; `ALuint` seems to be incompatible with `nullptr` as per VS (causes build failure) sooooo they should be the only ones left.

PR #114, opened 2020-01-16
Cleaning up the NovelRT cmake logic

tannergooding:
This cleans up the cmake logic for copying around the resources and building the .NET project. They should now correctly build and copy changes when the relevant bits change.


tannergooding:
We still need to clean up the logic that copies the dependencies over for `NovelRTSamples`. However, coming up with logic that "just works" doesn't look to be easy. Namely, you can't use generator expressions as part of the `OUTPUT` of a `custom_command` and so you can't just say "if any of these files has changed, run this command"

samples/CMakeLists.txt
```diff
@@ -9,8 +9,37 @@ set(NOVELRT_SAMPLES_LINK_LIBRARIES
...
+
+set(NOVELRT_SAMPLES_DEPENDENCIES_OUTPUTS
+  ${CMAKE_CURRENT_BINARY_DIR}/NovelRT.txt
+  ${CMAKE_CURRENT_BINARY_DIR}/NovelRTDotNet.txt
+  ${CMAKE_CURRENT_BINARY_DIR}/NovelRTResources.txt
+  ${CMAKE_CURRENT_BINARY_DIR}/NovelRTSamplesDependencies.txt
+)
+
+add_custom_command(OUTPUT ${NOVELRT_SAMPLES_DEPENDENCIES_OUTPUTS}
+  COMMAND ${CMAKE_COMMAND} -E copy_directory $<TARGET_FILE_DIR:NovelRT> $<TARGET_GENEX_EVAL:NovelRTSamples,$<TARGET_PROPERTY:NovelRTSamples,BINARY_OUTDIR>>
```
tannergooding: There needed to be a little bit of magic for `NovelRTSamples` to depend on `NovelRTSamplesDependencies` while still allowing `NovelRTSamplesDependencies` to determine where the executable was being placed.

src/CMakeLists.txt
```diff
@@ -51,50 +51,61 @@ endif()
...
 add_custom_command(TARGET NovelRT POST_BUILD
-  COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:NovelRT>/Resources/Fonts
-  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_FONT_RESOURCES} $<TARGET_FILE_DIR:NovelRT>/Resources/Fonts
-  COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:NovelRT>/Resources/Images
-  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_IMAGE_RESOURCES} $<TARGET_FILE_DIR:NovelRT>/Resources/Images
-  COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:NovelRT>/Resources/Scripts
-  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_SCRIPT_RESOURCES} $<TARGET_FILE_DIR:NovelRT>/Resources/Scripts
-  COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:NovelRT>/Resources/Shaders
-  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${NOVELRT_LIB_SHADER_RESOURCES} $<TARGET_FILE_DIR:NovelRT>/Resources/Shaders
+  COMMAND ${CMAKE_COMMAND} -E touch ${CMAKE_CURRENT_BINARY_DIR}/NovelRT.txt
```
tannergooding: `add_custom_command` doesn't currently allow `generator expressions` (such as `$<TARGET_FILE:NovelRT>`) as part of the `OUTPUT`, so we need a separate file which is output at the same time to track "up to date".

src/CMakeLists.txt
```diff
@@ -51,50 +51,61 @@ endif()
...
+file(GLOB_RECURSE NOVELRT_RESOURCE_INPUTS CONFIGURE_DEPENDS
+  Resources/Fonts/*.ttf
+  Resources/Images/*.png
+  Resources/Scripts/*.lua
+  Resources/Shaders/*.glsl
+)
+source_group(TREE ${CMAKE_SOURCE_DIR} FILES ${NOVELRT_RESOURCE_INPUTS})
+
+set(NOVELRT_RESOURCE_OUTPUTS
+  ${CMAKE_CURRENT_BINARY_DIR}/NovelRTResources.txt
```
tannergooding: The downside to tracking just the `*.txt` file as part of the output is that for non `library`/`executables` if you manually deleted one of the other outputs (say `<output>/Resources/Fonts/Raleway-Regular.ttf`) but not the `*.txt` file, it would think everything is still "up to date".

It does still correctly pick up new input files and modified input files which is what we really wanted however.

src/Resources/Scripts/avg.lua
```diff
@@ -1,4 +1,4 @@
 avg, sum = average(10, 20, 30, 40, 50)
 
 print("The average is ", avg)
-print("The sum is ", sum)
\ No newline at end of file
+print("The sum is ", sum)
```
RubyNova: What actually changed here?

tannergooding: New line at the end of the file

tannergooding:
I dont believe they should impact anyone else. It's just fixing the up to date checks for the local dev experience

PR #116, opened 2020-01-24
Fixed Doxygen and moving it to its own CMakeLists to conform with the new project standards.

QuantumlyTangled:


doxygen/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+find_package(Doxygen QUIET)
```
tannergooding: This looks good for now, but I wonder if it would be better if this was a reusable component that we can include per project instead?

We will likely have multiple tools/components/projects which should each have documentation.

PR #118, opened 2020-01-28
[WIP] Add support for rendering frame diagnostics.

### LINKED ISSUE #30: Add support for reporting frame diagnostics
NovelRunner should be updated such that it can report frame diagnostics.

This should initially just track simple information like current FPS, the min/max FPS, average FPS, and total frame count.

However, it could also be extended to report additional information, like how many objects are rendered in the frame, the loaded shaders, and other information that may be useful for debugging.

The user can then request the frame diagnostic information and display it to the user or perform other suitable actions with it.

Erlite:
This PR resolves #30 and will contain the following steps:

- [X] Create a metrics level enum to determine what metrics to record and/or output.
- [ ] Overhaul the `DebugService` to record metrics it needs to.
- [ ] Implement new overhaul in `NovelRunner`'s update logic. 
- [ ] Output the `DebugService`'s metrics to the main window.

Feel free to suggest metrics it should be able to record below. For now, I'll start simple with framerate, hardware usage and rendering statistics. I might include sound cache too.  

Another thing that could be good to add is outputting metrics to a log file.

### DIFF
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -98,6 +98,7 @@ namespace NovelRT::Windowing {
 }
 
 //enums
+#include "NovelRT/Debug/MetricsLevel.h"
 #include "NovelRT/Input/KeyCode.h"
 #include "NovelRT/Input/KeyState.h"
 #include "NovelRT/Graphics/CameraFrameState.h"
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -59,7 +59,6 @@
 
 namespace NovelRT {
   typedef void (*NovelUpdateSubscriber)(double deltaSeconds);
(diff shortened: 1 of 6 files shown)


Erlite:
I decided to expose the `StepTimer` in `NovelRunner` by adding the `getStepTimer` function. Note that it isn't marked `const` because `getActual()` isn't a pure function.

That way, `DebugService::recordMetrics()` can simply get the step timer if its metrics level is greater than or equal to `MetricsLevel::FramesPerSecond` and poll the current FPS itself. I'll continue in that route for the rest of the metrics unless someone has an objection or something needs to be changed.

RubyNova:
PR abandoned. Closing.

PR #119, opened 2020-01-28
InteractionService: fix out-of-range exception when pressing mouse buttons.

Erlite:
Fixes a small bug in the `InteractionService` where clicking a mouse button that wasn't inserted in `_keyStates` or `_mousePositionsOnScreenPerButton` would crash with an out of range exception due to the keys not existing. 

It'll now directly insert the key and value into the map if the key doesn't exist.

### DIFF
--- a/src/NovelRT/Input/InteractionService.cpp
+++ b/src/NovelRT/Input/InteractionService.cpp
@@ -37,12 +37,14 @@ namespace NovelRT::Input {
     }
   }
 
-  void InteractionService::tryProcessMouseState(KeyCode code) {
-    auto result = _keyStates.find(code);
+  void InteractionService::processMouseStates() {
+    for (int i = (int)KeyCode::FirstMouseButton; i < (int)KeyCode::LastMouseButton; i++) {
+      auto keyCode = static_cast<KeyCode>(i);
+      auto result = _keyStates.find(keyCode);
 
-    if (result != _keyStates.end() && (result->first >= KeyCode::FirstMouseButton && result->first <= KeyCode::LastMouseButton)) {
+      if (result == _keyStates.end()) continue;
       processKeyState(result->first, result->second);
-    }
+    }   
   }
 
   void InteractionService::acceptKeyboardInputBindingPush(int key, int action) {
@@ -89,9 +91,7 @@ namespace NovelRT::Input {
   }
 
 void InteractionService::consumePlayerInput() {
-  for (auto it : _keyStates) {
-    tryProcessMouseState(it.first);
-  }
+  processMouseStates();
   glfwPollEvents();
 }
 
--- a/src/NovelRT/Input/InteractionService.cpp
+++ b/src/NovelRT/Input/InteractionService.cpp
@@ -9,13 +9,17 @@ namespace NovelRT::Input {
     _logger(LoggingService(Utilities::Misc::CONSOLE_LOG_INPUT))
   {}
 
-  void InteractionService::processKeyState(KeyCode code, KeyState state) {
+  void InteractionService::validateIfKeyCached(KeyCode code) {
     auto result = _keyStates.find(code);
-    
+
     if (result == _keyStates.end()) {
-      _keyStates.insert({ code, state });
-      return;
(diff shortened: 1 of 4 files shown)


capnkenny:
Resolves #120

src/NovelRT/Input/InteractionService.cpp
```diff
@@ -47,8 +49,14 @@ namespace NovelRT::Input {
   void InteractionService::acceptMouseButtonClickPush(int button, int action, const Maths::GeoVector<float>& mousePosition) {
     auto keyState = static_cast<KeyState>(action);
     auto keyCode = static_cast<KeyCode>(button);
+    auto value = mousePosition.getVec4Value() * glm::scale(glm::vec3(1920.0f / _screenSize.getX(), 1080.0f / _screenSize.getY(), 0.0f));
 
-    _mousePositionsOnScreenPerButton.at(keyCode) = mousePosition.getVec4Value() * glm::scale(glm::vec3(1920.0f / _screenSize.getX(), 1080.0f / _screenSize.getY(), 0.0f));
+    if (!_mousePositionsOnScreenPerButton.count(keyCode)) {
+      _mousePositionsOnScreenPerButton.insert({ keyCode, value });
```
RubyNova: Why did this have to be moved from the constructor?

src/NovelRT/Input/InteractionService.cpp
```diff
@@ -47,8 +49,14 @@ namespace NovelRT::Input {
   void InteractionService::acceptMouseButtonClickPush(int button, int action, const Maths::GeoVector<float>& mousePosition) {
     auto keyState = static_cast<KeyState>(action);
     auto keyCode = static_cast<KeyCode>(button);
+    auto value = mousePosition.getVec4Value() * glm::scale(glm::vec3(1920.0f / _screenSize.getX(), 1080.0f / _screenSize.getY(), 0.0f));
 
-    _mousePositionsOnScreenPerButton.at(keyCode) = mousePosition.getVec4Value() * glm::scale(glm::vec3(1920.0f / _screenSize.getX(), 1080.0f / _screenSize.getY(), 0.0f));
+    if (!_mousePositionsOnScreenPerButton.count(keyCode)) {
```
RubyNova: Surely we should be using `.find` here then checking if the iterator reach the end?

src/NovelRT/Input/InteractionService.cpp
Erlite: Because any key that wasn't inserted in the constructor would cause an exception when used. So if it doesn't find it, it'll insert it directly.

Erlite:
That should be all, anything else I should change?

Erlite:
The service now iterates through all mouse buttons to process them and no longer contains any hardcoded checks for the left mouse button. Tested with the two additional buttons on my mouse (`Mouse4` and `Mouse5`) and it did recognize them, triggering the interactions. 

This PR should be ready to go now.

src/NovelRT/Input/InteractionService.cpp
```diff
@@ -41,31 +57,43 @@ namespace NovelRT::Input {
...
+      _logger.logDebug("Valid click target detected! Executing...");
+      _clickTarget = target;
     }
   }
 
-  void InteractionService::consumePlayerInput() {
-    processKeyState(KeyCode::LeftMouseButton, _keyStates.at(KeyCode::LeftMouseButton));
-    glfwPollEvents();
+void InteractionService::consumePlayerInput() {
+  for (auto it : _keyStates) {
```
RubyNova: Isn't this going to process all the keystates that we have cached?

Erlite: It calls `tryProcessMouseState(KeyCode code)` which in turn checks if it's a mouse button.

However, I'll make it parameterless and loop through mouse buttons directly, that way it only loops for each mouse button instead of each registered key state which could get big.

RubyNova: Right but the loop as-is will iterate 50 times even if the method doesn't do anything with it. Ty for the change. 😄

PR #121, opened 2020-01-31
Updating the WindowingService to attempt multiple contexts

tannergooding:
This fixes the WindowingService to allow successful window creation on a range of architectures, platforms, and hardware.


tannergooding:
An example of the first two context types failing:
image

PR #122, opened 2020-01-31
Add helper method for throwing on nullptrs and apply where necessary

### LINKED ISSUE #71: Add helper for failure reponses from methods
Currently, we are checking against constants or magic numbers, especially when dealing with SDL-based calls. These should get wrapped in a helper method so that if they fail, we can handle them appropriately with logging.

_As referenced in basic audio PR:_

`SDL_SUCCESS` in particular seems like something that could be solved by a helper method. Namely something like `ThrowIfFailed(MyFunc(args), "Error Message")`

This would then check for success, log the error, and throw an exception.

_Originally posted by @tannergooding in https://github.com/novelrt/NovelRT/pull/33_

mastorm:
I implemented this with @RubyNova. Basically a method in the logger that allows you to pass in an object and an error message, then it throws when the object is a nullptr.

Closes #71


include/NovelRT/LoggingService.h
```diff
@@ -34,6 +34,7 @@ namespace NovelRT {
     void logDebugLine(const std::string& message);
     void logInternal(const std::string& message, LogLevel level);
     void setLogLevel(LogLevel level);
+    void throwIfNullPtr(const void* const object, const std::string& exceptionMessage);
```
capnkenny: Two `const`s here?

RubyNova: Yes - an immutable pointer to an immutable object.

src/NovelRT/Graphics/ImageRect.cpp
```diff
@@ -123,17 +123,11 @@ namespace NovelRT::Graphics {
...
-        }
+        _logger.throwIfNullPtr(png, "Image file cannot be opened! Please ensure the path is correct and that the file is not locked.");
 
         auto info = png_create_info_struct(png);
 
-        if (info == nullptr) {
-          _logger.logError("Image at path " + _imageDir + " failed to provide an info struct! Aborting...");
-          throw std::runtime_error("Unable to continue! File failed to load for texture.");
-        }
+        _logger.throwIfNullPtr(info, "Image at path " + _imageDir + " failed to provide an info struct! Aborting...");
```
capnkenny: @tannergooding based on how we wanted to handle multiple strings,  should the function be changed to support it (a la the `logError` template)? Not sure if we want to stray from  it for this PR or not.

include/NovelRT/LoggingService.h
capnkenny: :+1:

src/NovelRT/Graphics/ImageRect.cpp
tannergooding: This should have been resolved with <https://github.com/novelrt/NovelRT/pull/141>

tannergooding:
Thanks for the fix @mastorm

PR #133, opened 2020-02-16
Adding a basic QuadTree implementation

tannergooding:
This adds a very basic quad tree implementation that will help with collision detection and scene graph management.


include/NovelRT/Maths/QuadTree.h
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    QuadTree* getBottomRight() const noexcept {
+      return _children[BOTTOM_RIGHT].get();
+    }
+
+    bool tryInsert(std::shared_ptr<TQuadTreePoint> point) noexcept {
+      if (point == nullptr || !getBounds().pointIsWithinBounds(point->getPosition())) return false;
+
+      if (_children[TOP_LEFT] == nullptr) {
```
RubyNova: this should be using `.at()` or `.find()` most probably?

include/NovelRT/Maths/QuadTree.h
```diff
@@ -0,0 +1,135 @@
...
+        if (_pointCount < POINT_CAPACITY) {
+          _points[_pointCount++] = point;
```
RubyNova: this should be using `.at()` or `.find()` most probably?

include/NovelRT/Maths/QuadTree.h
```diff
@@ -0,0 +1,135 @@
...
+          return true;
+        }
+        subdivideTree();
+      }
+
+      auto result = _children[TOP_LEFT]->tryInsert(point) ||
+                    _children[TOP_RIGHT]->tryInsert(point) ||
+                    _children[BOTTOM_LEFT]->tryInsert(point) ||
+                    _children[BOTTOM_RIGHT]->tryInsert(point);
```
RubyNova: this should be using `.at()` or `.find()` most probably?

include/NovelRT/Maths/QuadTree.h
```diff
@@ -0,0 +1,135 @@
...
+      const GeoVector<float> BOTTOM_LEFT_SCALE = GeoVector<float>(-0.5, -0.5);
+      const GeoVector<float> BOTTOM_RIGHT_SCALE = GeoVector<float>(+0.5, -0.5);
+
+      GeoVector size = getBounds().getSize() / 2;
+      GeoVector position = getBounds().getPosition();
+
+      _children[TOP_LEFT] = std::make_unique<QuadTree>(GeoBounds(position + (size * TOP_LEFT_SCALE), size, 0));
+      _children[TOP_RIGHT] = std::make_unique<QuadTree>(GeoBounds(position + (size * TOP_RIGHT_SCALE), size, 0));
+      _children[BOTTOM_LEFT] = std::make_unique<QuadTree>(GeoBounds(position + (size * BOTTOM_LEFT_SCALE), size, 0));
+      _children[BOTTOM_RIGHT] = std::make_unique<QuadTree>(GeoBounds(position + (size * BOTTOM_RIGHT_SCALE), size, 0));
```
RubyNova: this should be using `.at()` or `.find()` most probably?

include/NovelRT/Maths/QuadTree.h
```diff
@@ -0,0 +1,135 @@
...
+    void getIntersectingPoints(const GeoBounds& bounds, std::vector<std::shared_ptr<TQuadTreePoint>>& intersectingPoints) {
+      if (getBounds().intersectsWith(bounds)) {
+        return;
+      }
+
+      if (_children[TOP_LEFT] == nullptr) {
+        for (uint32_t index = 0; index < getPointCount(); index++) {
+          auto point = getPoint(index);
+          if (bounds.pointIsWithinBounds(point->getPosition())) {
+            intersectingPoints.push_back(point);
```
RubyNova: would `emplace_back` not be better here?

include/NovelRT/Maths/QuadTree.h
```diff
@@ -0,0 +1,135 @@
...
+          }
+        }
+      } else {
+        _children[TOP_LEFT]->getIntersectingPoints(bounds, intersectingPoints);
+        _children[TOP_RIGHT]->getIntersectingPoints(bounds, intersectingPoints);
+        _children[BOTTOM_LEFT]->getIntersectingPoints(bounds, intersectingPoints);
+        _children[BOTTOM_RIGHT]->getIntersectingPoints(bounds, intersectingPoints);
```
RubyNova: this should be using `.at()` or `.find()` most probably?

include/NovelRT/Maths/QuadTree.h
tannergooding: this is `std::array` so it is fixed size and won't risk accidentally growing. We are also using a known constant index, so we don't want `find()`.

include/NovelRT/Maths/QuadTree.h
tannergooding: We can't use `at` and be `noexcept`. In this case, we know the upper limit and that `pointCount` points to the next valid entry in the fixed sized array.

include/NovelRT/Maths/QuadTree.h
tannergooding: Now that this is `std::shared_ptr`, yes. Will fix. There would have been no difference when it was just `T*` since that is a single write either way.

RubyNova: Ah I see, awesome.

PR #134, opened 2020-02-16
Some minor improvements

tannergooding:
This adds some minor improvements to the codebase to improve the overall debugging experience.

### DIFF
--- a/src/NovelRT/LoggingService.cpp
+++ b/src/NovelRT/LoggingService.cpp
@@ -5,11 +5,17 @@
 
 namespace NovelRT {
 
+#ifndef NDEBUG
+  typedef spdlog::synchronous_factory spdlog_factory;
+#else
+  typedef spdlog::async_factory spdlog_factory;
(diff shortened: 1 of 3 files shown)


src/NovelRT/LoggingService.cpp
```diff
@@ -5,11 +5,17 @@
 
 namespace NovelRT {
 
+#ifndef NDEBUG
+  typedef spdlog::synchronous_factory spdlog_factory;
+#else
+  typedef spdlog::async_factory spdlog_factory;
+#endif
```
RubyNova: Is this the only way to do this? Could we not use an `std::conditional` or is that not doable?

tannergooding: I didn't try. This is keyed off a define, so I went with the same approach we use elsewhere

RubyNova: OK, fair enough.

Issue #56, opened 2019-09-27
OpenGL Instancing.


RubyNova:
Currently our approach to loading the same normalised data (upcoming in a WIP feature branch) is not that efficient. We need to instance the data onto the GPU ideally.

RubyNova:
Resolved by #86 .

RubyNova:
lmao closed the wrong issue.

PR #135, opened 2020-02-16
Basic animation toolset and structural changes.

### LINKED ISSUE #72: Add NovelRT::TextureConfig type.
As the title suggests, I think it would be good to add a configurable texture type that houses all the relevant texture information as part of #56 .

This is primarily to allow us to create texture-coordinate-based effects amongst other things while also aiding with the instancing ticket.

It would also bring us more in-line with engines such as Unity3D that support this sort of behaviour out of the box.

Thoughts?

RubyNova:
This PR resolves #72 and partially adds some features for #56 .

This adds a basic sprite animator framework and some GPU instancing changes (albeit primitive ones). This should help for the demo and with texture buffer abuse.


include/NovelRT.h
```diff
@@ -20,9 +20,11 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <tuple>
 #include <typeinfo>
 #include <type_traits>
 #include <vector>
+#include <queue>
```
tannergooding: nit: order (unless we otherwise have a need)

include/NovelRT/Animation/SpriteAnimatorFrame.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace NovelRT::Animation {
+  class SpriteAnimatorFrame {
+  public:
+    Utilities::Event<> FrameEnter;
+    Utilities::Event<> FrameExit;
+
+  private:
+    std::shared_ptr<Graphics::Texture> _texture;
+    float _duration;
```
tannergooding: We should eventually change these to be some kind of `Timestamp` or `Timespan` like type, which tracks this in normalized ticks. Same I think for the `accumulatedDelta` in the `SpriteAnimator`.
Doing so helps avoid issues with floating-point over many frames.

include/NovelRT/Animation/SpriteAnimatorState.h
```diff
@@ -0,0 +1,70 @@
...
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Animation {
+  class SpriteAnimatorState {
+  private:
+    std::vector<SpriteAnimatorFrame> _frames;
+    bool _shouldLoop;
+    std::vector<std::tuple<std::shared_ptr<SpriteAnimatorState>, std::vector<std::function<bool()>>>> _transitions;
```
tannergooding: if we renamed `Event` to `Delegate`, we could use it in more places (like the `std::vector<std::function<bool()>>` here).

tannergooding: (just a note for a potential future cleanup)

src/NovelRT/Animation/SpriteAnimator.cpp
```diff
@@ -0,0 +1,73 @@
...
+        auto newFrame = _currentState->getFrames()->at(_currentFrameIndex);
+        newFrame.FrameEnter();
+        _rect->setTexture(newFrame.getTexture());
+      }
+
+
+      _accumulatedDelta += delta;
+      break;
+    }
+    case AnimatorPlayState::Paused:
```
tannergooding: Log an issue to implement this?

RubyNova: I mean, technically paused doesn't do anything :^) this is really just a case to sort of "freeze" the animator. Maybe we should just delete the case lol?

src/NovelRT/Graphics/FontSet.cpp
```diff
@@ -0,0 +1,80 @@
...
+    _id(id),
+    _fontFile("") {
+
+  }
+
+  void FontSet::loadFontAsTextureSet(const std::string& file, float fontSize) {
+    if (!_fontFile.empty()) {
+      _logger.logErrorLine("This FontSet has already been initialised with data. Please make a new FontSet!");
+      throw std::runtime_error("Unable to continue! Cannot overwrite FontSet!");
+      }
```
tannergooding: closing brace indentation is off

src/NovelRT/Graphics/RenderingService.cpp
```diff
@@ -4,9 +4,28 @@
 #define GL_GLEXT_PROTOTYPES
 
 namespace NovelRT::Graphics {
+  RenderingService::RenderingService(NovelRunner* const runner) :
+    _logger(LoggingService(Utilities::Misc::CONSOLE_LOG_GFX)),
+    _runner(runner),
+    _cameraObjectRenderUbo(std::function<GLuint()>([] {
+    GLuint tempHandle;
```
tannergooding: indentation seems off

src/NovelRT/Graphics/FontSet.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+      if (_fontCharacters.size() != 0) _fontCharacters.clear();
```
tannergooding: This and the following line are very hard to read.

I think its because the if doesn't have braces and the "contents" are on the same line. So my first assumption is that `FT_Library freeTypeLoader` is improperly indented

RubyNova:
I'll introduce some tests and clean up the simpler items.

src/NovelRT/Animation/SpriteAnimator.cpp
tannergooding: That might be good.

In the future we might consider collecting metrics for how much time a given animation is paused (since that might be used for other transitions, etc).

include/NovelRT.h
RubyNova: This include isn't actually required. It is being removed.

include/NovelRT/Animation/SpriteAnimatorState.h
```diff
@@ -11,12 +11,22 @@ namespace NovelRT::Animation {
   class SpriteAnimatorState {
   private:
     std::vector<SpriteAnimatorFrame> _frames;
-    bool _shouldLoop;
+    bool _shouldLoop = false; //this is to shut the warning up since we have no ctor
```
tannergooding: Don't we need a ctor to explicitly zero init the vectors?

RubyNova: MSVC wasn't complaining about those, only the bool. /shrug

Want me to do it in a ctor anyway?

include/NovelRT/Animation/SpriteAnimatorState.h
```diff
@@ -11,12 +11,22 @@ namespace NovelRT::Animation {
...
 
   public:
-
     inline void insertNewState(std::shared_ptr<SpriteAnimatorState> stateTarget, std::vector<std::function<bool()>> transitionConditions) {
+
+      if (stateTarget == nullptr) {
+        return;
+      }
+
+      for (size_t i = 0; i < transitionConditions.size(); i++) {
```
tannergooding: What is this for? It looks like its looking for existing `nullptr`, which should never exist?

RubyNova: This is what I asked you about on Discord. This loops checks each condition to see if someone passed in an invalid function, then it erases it if that's the case, allowing for normal operation to continue.

tests/NovelRT.Tests/Animation/SpriteAnimatorStateTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+using namespace NovelRT;
+using namespace NovelRT::Maths;
+using namespace NovelRT::Animation;
+
+class SpriteAnimatorStateTest : public testing::Test {
+protected:
+  std::vector<std::shared_ptr<SpriteAnimatorState>> _states;
+
+  void SetUp() override {
+    _states = std::vector<std::shared_ptr<SpriteAnimatorState>>{
```
tannergooding: What is the behavior if we use the constructor that takes the initial size?
It looks like it calls `std::shared_ptr<SpriteAnimatorState>()`, which would give us a `shared_ptr` with no backing allocation, in which case we could delay allocating until it is needed?

tests/NovelRT.Tests/Animation/SpriteAnimatorStateTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+      std::make_shared<SpriteAnimatorState>(),
+      std::make_shared<SpriteAnimatorState>(),
+      std::make_shared<SpriteAnimatorState>(),
+      std::make_shared<SpriteAnimatorState>(),
+    };
+  }
+};
+
+void setUpRelationships(std::vector<std::shared_ptr<SpriteAnimatorState>> targets, std::vector<std::function<bool()>> conditions = std::vector<std::function<bool()>>()) {
+  for (size_t i = 0; i < targets.size(); i++) {
```
tannergooding: Is this setting up a circular buffer of animation states?

tests/NovelRT.Tests/Animation/SpriteAnimatorStateTest.cpp
tannergooding: I missed that this was a tests file, you can ignore :)

RubyNova: I am unsure. I just did this way so it was explicitly clear what I wanted.

tests/NovelRT.Tests/Animation/SpriteAnimatorStateTest.cpp
RubyNova: Its not a circular buffer, its a state machine.

Each state holds a collection of frames and an indefinite amount of references to other states, with functions evaluating to `bool` as the "transition" conditions.

include/NovelRT/Animation/SpriteAnimatorState.h
tannergooding: oh, I see. it is removing them from `transitionConditions` the parameter.
long day, carry on.

include/NovelRT/Animation/SpriteAnimatorState.h
tannergooding: Might be worth testing in Release mode to ensure we don't get wonky behavior, but otherwise I think is fine

PR #136, opened 2020-02-17
Fixed zero-tick delta behavior (#41)

### LINKED ISSUE #41: Update the logic in StepTimer to account for scenarios where _frequency > TicksPerSecond
As called out on https://github.com/novelrt/NovelRT/pull/34#discussion_r323548711, the StepTimer can technically bug out if the hardware timer frequency is greater than the TicksPerSecond.

This is problematic because the math is using integer arithmetic and follows `(counterDelta * TicksPerSecond) / _frequency`, so if `_frequency > TicksPerSecond` you can end up getting "zero" as the tick delta.

A solution should be devised that takes this into account so that users still get the appropriate updates.

Pheubel:
Handled the edge case depicted in ["Update the logic in StepTimer to account for scenarios where _frequency > TicksPerSecond"](https://github.com/novelrt/NovelRT/issues/41).

I would like a review on the implementation.

EDIT: Resolves #41 - Ruby


src/NovelRT/Timing/StepTimer.cpp
```diff
@@ -8,7 +8,7 @@
 namespace NovelRT::Timing {
   StepTimer::StepTimer(uint32_t targetFrameRate, double maxSecondDelta) :
     _frequency(glfwGetTimerFrequency()),
-    _maxCounterDelta((uint64_t)(_frequency * maxSecondDelta)),
+    _maxCounterDelta((uint64_t)(_frequency* maxSecondDelta)),
```
tannergooding: nit: a whitespace was removed

src/NovelRT/Timing/StepTimer.cpp
```diff
@@ -77,7 +77,8 @@ namespace NovelRT::Timing {
           subscriber(secondsDelta);
         }
       }
-    } else {
+    }
+    else {
```
tannergooding: @RubyNova, most K&R variants want the `else` on the same line as the closing brace, yeah?

RubyNova: I *believe* so, however our clang format isn't set up yet so we just kind of have both everywhere at the moment.

We should uh...probably fix that....lmao.

tannergooding: Well decide which we want so I know if this comment is valid :P

RubyNova: I think I would prefer it to be on the same line, just because doing K&R with it on a new line just looks strange.

src/NovelRT/Timing/StepTimer.cpp
Pheubel: i don't get why VS's autoformat wants it connected.

tannergooding:
Thanks for the fix @Pheubel

PR #137, opened 2020-02-17
Changing events to be an explicit type with handlers so unsubscription can happen

tannergooding:
`std::function` is just a wrapper which allows consistent invocation regardless of the underlying type (be it a function pointer, lambda, bound method, etc). This means that you can't match it for equality.

This PR replaces our old event logic with a new pair of types: `Event` and `EventHandler`. The `Event` just tracks the set of `EventHandler` instances that are currently subscribed.
The latter type wraps a `std::function` but also a unique id which can be used to identify the instance and therefore will allow a given handler to be unsubscribed from the event.

There are also convenience functions so much of the existing code continues to work. So, ultimately, if you want to just attach an event for the entire lifetime of the app, you can simply `event += []() { code }`, but if you want the opportunity to unsubscribe later, you must first construct and track an `EventHandler`.


src/NovelRT/DebugService.cpp
```diff
@@ -30,7 +30,7 @@ namespace NovelRT {
     _runner(runner),
     _fpsCounter(nullptr),
     _framesPerSecond(0) {
-    runner->subscribeToSceneConstructionRequested(std::bind(&DebugService::onSceneConstruction, this));
+    runner->SceneConstructionRequested += std::bind(&DebugService::onSceneConstruction, this);
```
RubyNova: IIRC, there was some kind of lambda replacement for `std::bind` - though I can't say I quite understand what that replacement would be or look like.

Your call if you wanna investigate this further.

tannergooding: Not in this PR. If you want to log an issue, we can follow up in the future.

PR #140, opened 2020-02-17
Fixing Event<> fields to be first so they are initialized first

tannergooding:


include/NovelRT/Input/InteractionObject.h
```diff
@@ -11,15 +11,16 @@ namespace NovelRT::Input {
   class InteractionObject : public WorldObject {
     friend class InteractionService;
 
+  public:
+    Utilities::Event<> Interacted;
```
tannergooding: Apparently C++ is free to initialize fields in declaration order, rather than in initializer order, so these need to be moved up here.

That is, if you have:
```
int _x;
int _y;

constructor() :
    _y(0), _x(1) { }
```

The compiler is to run the initializer for `_x` before it runs the initializer for `_y`. You can see `Initialization order` under https://en.cppreference.com/w/cpp/language/initializer_list for more details

PR #141, opened 2020-02-17
Forward variadic arguments instead of adding on new line

### LINKED ISSUE #110: Single log calls should log on a single line
Currently calls such as `_logger.logError("Mixer Error: ", getSDLError());` will print:
```
[2020-01-09 20:21:15.317] [Audio] [error] Mixer Error: 
[2020-01-09 20:21:15.317] [Audio] [error] No such audio device`
```

When the intent was likely:
```
[2020-01-09 20:21:15.317] [Audio] [error] Mixer Error: No such audio device
```

ClxS:
Resolves #110 . Previously the method would log each argument individually. It now forwards them to sdplog.

### DIFF
--- a/include/NovelRT/LoggingService.h
+++ b/include/NovelRT/LoggingService.h
@@ -37,23 +37,19 @@ namespace NovelRT {
 
     template <typename I, typename ... IRest>
     void logInfo(I current, IRest ... next) {
-      _logger->info(current);
-      logInfo(next ...);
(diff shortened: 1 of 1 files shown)


RubyNova:
Maybe I can add some tests post-merge later?

RubyNova:
I'll log an issue for the tests and write them myself later. Thank you for fixing this!

PR #145, opened 2020-02-19
Added detection for NVIDIA Optimus (Win x86/x64 only)

capnkenny:
Adds detection for NVIDIA Optimus, and logs if it is not detected.
Linux deals with this differently so it is Windows only.

### DIFF
--- a/src/NovelRT/Windowing/WindowingService.cpp
+++ b/src/NovelRT/Windowing/WindowingService.cpp
@@ -24,6 +24,22 @@ namespace NovelRT::Windowing {
     float wData = displayData->width * 0.7f;
     float hData = displayData->height * 0.7f;
 
+#if defined (_WIN64)
+    auto optimus = LoadLibrary("nvapi64.dll");
+    if (optimus != nullptr) {
+      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
+    } else {
+      _logger.logWarningLine("NVIDIA GPU not detected. Continuing w/o Optimus support.");
+    }
+#elif defined (_WIN32)
+    auto optimus = LoadLibrary("nvapi.dll");
+    if (optimus != nullptr) {
+      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
+    } else {
+      _logger.logWarningLine("NVIDIA GPU not detected. Continuing w/o Optimus support.");
+    }
+#endif
+
     _logger.logInfoLine("Attempting to create OpenGL ES v3.0 context using EGL API");
 
     glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API);
--- a/src/NovelRT/Windowing/WindowingService.cpp
+++ b/src/NovelRT/Windowing/WindowingService.cpp
@@ -119,21 +119,15 @@ namespace NovelRT::Windowing {
   void WindowingService::checkForOptimus() {
   #if defined (_WIN64)
     auto optimus = LoadLibrary("nvapi64.dll");
-    if (optimus != nullptr) {
-      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
-    }
-    else {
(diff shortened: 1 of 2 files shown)


src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -24,6 +24,22 @@ namespace NovelRT::Windowing {
     float wData = displayData->width * 0.7f;
     float hData = displayData->height * 0.7f;
 
+#if defined (_WIN64)
+    auto optimus = LoadLibrary("nvapi64.dll");
+    if (optimus != nullptr) {
+      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
+    } else {
+      _logger.logWarningLine("NVIDIA GPU not detected. Continuing w/o Optimus support.");
+    }
```
RubyNova: Maybe this should be a method? You kinda just have this same snippet twice.

capnkenny: Done.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -113,6 +115,21 @@ namespace NovelRT::Windowing {
...
+  void WindowingService::checkForOptimus() {
+  #if defined (_WIN64)
+    auto optimus = LoadLibrary("nvapi64.dll");
+  #elif defined (_WIN32)
+    auto optimus = LoadLibrary("nvapi.dll");
+  #endif
+    if (optimus != nullptr) {
+      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
+    }
+    else {
```
RubyNova: nit: Put this after the closing brace of the if statement.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -113,6 +115,20 @@ namespace NovelRT::Windowing {
 
 
   }
+
+  void WindowingService::checkForOptimus() {
```
tannergooding: nit: This should probably be called `tryLoadOptimus`. It would also be nice to not "leak" the handle.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -113,6 +115,20 @@ namespace NovelRT::Windowing {
...
+  #if defined (_WIN64)
+    auto optimus = LoadLibrary("nvapi64.dll");
+  #elif defined (_WIN32)
+    auto optimus = LoadLibrary("nvapi.dll");
+  #endif
+    if (optimus != nullptr) {
+      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
+    } else {
+      _logger.logWarningLine("NVIDIA GPU not detected. Continuing w/o Optimus support.");
```
tannergooding: nit: This shouldn't be a warning. It isn't invalid to have an AMD GPU :P

src/NovelRT/Windowing/WindowingService.cpp
tannergooding: Is there an equivalent on Linux? What about systems that have both AMD and Nvidia?

RubyNova: Linux equivalent is apparently related to the NVIDIA FOSS driver, which hasn't been figured out yet.

src/NovelRT/Windowing/WindowingService.cpp
RubyNova: Good spot, lmao

capnkenny: Nobody said its invalid :P I'll change it anyways 🙂

tannergooding: I would generally interpret a warning as a "recoverable failure".
I wouldn't think this case is a failure, and was more just checking for optional features, in which case info is probably better.

capnkenny: That makes sense

PR #147, opened 2020-02-20
Improve CMakeLists and support for other IDEs

FiniteReality:
I rewrote the CMakeLists to include vcpkg into the CMake configuration process. This makes it easier to support other IDEs, like KDevelop.

Hopefully the CI likes my changes to the CI build scripts, to remove the vcpkg stuff. It should in theory be fine, though.

PR #149, opened 2020-02-20
Make sure std::filesystem::path is correct

FiniteReality:
This should (in theory) fix the issues you were having where some paths were incorrect. Haven't tested if this builds yet since I'm waiting on #147 :sweat_smile:


RubyNova:
Well if it doesn't build, CI will yell, so lets find out :^)

PR #150, opened 2020-02-21
Added support for checking the state of keys.

RubyNova:
Title.

### DIFF
--- a/samples/main.cpp
+++ b/samples/main.cpp
@@ -157,7 +157,7 @@ int main(int argc, char* argv[])
 
   runner.getDebugService().lock()->setIsFpsCounterVisible(true);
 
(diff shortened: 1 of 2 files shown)


samples/main.cpp
```diff
@@ -170,6 +170,10 @@ int main(int argc, char* argv[])
     }
 
     novelChanRect->getTransform().setRotation(rotation);
+
+    if (runner.getInteractionService().lock()->getKeyState(NovelRT::Input::KeyCode::W) != NovelRT::Input::KeyState::Idle) {
+      console.logInfoLine("W is not idle!");
```
tannergooding: Seems like this will be a lot of logging. We might want to revisit this when we clean up the sample in a couple weeks.

PR #152, opened 2020-02-22
Adding a basic scene graph

tannergooding:
This adds support for a basic scene graph including node insertion and traversal.

### DIFF
--- a/include/NovelRT/SceneGraph/SceneNode.h
+++ b/include/NovelRT/SceneGraph/SceneNode.h
@@ -1,0 +1,255 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+
+#ifndef NOVELRT_SCENENODE_H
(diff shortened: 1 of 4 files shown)


include/NovelRT/SceneGraph/SceneNode.h
```diff
@@ -0,0 +1,255 @@
...
+        if (_visitedNodes.find(child) == _visitedNodes.cend()) {
+          _visitedNodes.insert(child);
+          _pendingNodes.push(child);
+        }
+      }
+
+      return *this;
+    }
+
+    breadth_first_traversal_result_iterator operator++(int) {
```
RubyNova: What's with the naming convention here (and similar places)?

tannergooding: is just matching the convention for stdlib iterator names
I couldn't come up with a better one (just maybe `BreadthFirstTraversalResultIterator` or something)
the iterators in particular, users shouldn't be creating themselves

PR #153, opened 2020-02-22
Cleaning up the warnings given by MSVC and Clang

tannergooding:


include/NovelRT.h
```diff
@@ -3,13 +3,12 @@
 #ifndef NOVELRT_H
 #define NOVELRT_H
 
-//cstdlib
-#include <cmath>
-#include <cstdint>
-
 //stdlib
+#define __STDC_WANT_LIB_EXT1__
```
RubyNova: What does this do?

tannergooding: It exposes the "secure" variants of certain C stdlib functions (like `fopen_s`).

RubyNova: secure in what sense?

tannergooding: Basically the original C functions didn't take into account things like buffer overrun or certain types of errors.
Microsoft exposed "secure versions" of these functions (generally ending in `_s`) which started taking these into account.
These were later standardized as an official extension known as `__STDC_WANT_LIB_EXT1__` (Annex K - Bounds-checking interfaces in the current C language specification)
You can see https://docs.microsoft.com/en-us/cpp/c-runtime-library/security-features-in-the-crt?view=vs-2019 (and the C language spec) for more details.

RubyNova: That's cool! Thanks for explaining that :D

include/NovelRT.h
```diff
@@ -28,6 +27,11 @@
 #include <type_traits>
 #include <vector>
 
+#if defined(_WIN32) || defined(_WIN64)
+#define WIN32_LEAN_AND_MEAN
```
RubyNova: This is a very uhh, creative define? could you tell me why this name and why were including Windows.h for windows builds?

include/NovelRT/Animation/SpriteAnimatorState.h
```diff
@@ -23,7 +23,7 @@ namespace NovelRT::Animation {
 
       for (size_t i = 0; i < transitionConditions.size(); i++) {
         if (transitionConditions.at(i) == nullptr) {
-          transitionConditions.erase(transitionConditions.begin() + i);
+          transitionConditions.erase(transitionConditions.begin() + static_cast<std::vector<SpriteAnimatorFrame>::iterator::difference_type>(i));
```
RubyNova: What's this do?

include/NovelRT/Timing/StepTimer.h
```diff
@@ -83,11 +83,11 @@ namespace NovelRT::Timing {
     }
 
     static double TicksToSeconds(uint64_t ticks) {
```
RubyNova: looks like were violating our naming convention in a couple places - not for this PR, but for the clang format review.

include/NovelRT.h
tannergooding: `WIN32_LEAN_AND_MEAN` is just a special define that tells `<Windows.h>` to not include the world

We need to define it ourselves due to how glad defines some of its own macros if `Windows` was not included yet.

include/NovelRT/Utilities/Misc.h
```diff
@@ -11,13 +11,13 @@ namespace NovelRT::Utilities {
...
-    static inline const std::string CONSOLE_LOG_AUDIO = "Audio";
-    static inline const std::string CONSOLE_LOG_INPUT = "Input";
-    static inline const std::string CONSOLE_LOG_WINDOWING = "WindowManager";
+    static inline const char* CONSOLE_LOG_GENERIC = "NovelRT";
+    static inline const char* CONSOLE_LOG_APP = "Application";
+    static inline const char* CONSOLE_LOG_DOTNET = ".NET";
+    static inline const char* CONSOLE_LOG_GFX = "GFX";
+    static inline const char* CONSOLE_LOG_AUDIO = "Audio";
+    static inline const char* CONSOLE_LOG_INPUT = "Input";
+    static inline const char* CONSOLE_LOG_WINDOWING = "WindowManager";
```
RubyNova: Why the change in type here?

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -319,7 +327,7 @@ std::string AudioService::getALError() {
       return std::string("The requested operation resulted in OpenAL running out of memory.");
     }
     default: {
-      return nullptr;
+      return std::string("");
```
RubyNova: any particular reason you didn't just do `return "";` here?

include/NovelRT/Animation/SpriteAnimatorState.h
tannergooding: We have a `size_t` which is unsigned, while the difference between two pointers is signed, so we need to convert for this operation to be valid.
* MSVC and Clang warn here because you can overflow going from `unsigned` to `signed` and get potentially undesired behavior

The stdlib has usings in their templated types that tell you what is used for various operations. In this case `iterator::difference_type` which is the equivalent to `ptrdiff_t` for the type contained by the vector.
So we cast our `size_t` to that type.

include/NovelRT/Utilities/Misc.h
tannergooding: statics for non primitive types require global constructors and destructors, which Clang warns about (it increases startup and shutdown cost, among other things).
defining it as the primitive type instead makes it a true constant with no startup overhead and users can still get a `std::string` if required

src/NovelRT/Audio/AudioService.cpp
tannergooding: To match us doing `std::string("...")` in the other cases

PR #160, opened 2020-02-25
Reworked sound sources for OpenAL

capnkenny:
Can now create up to 256 sources theoretically (OpenAL's limit) - must be managed manually as sources are handled, not buffers.

### DIFF
--- a/include/NovelRT/Audio/AudioService.h
+++ b/include/NovelRT/Audio/AudioService.h
@@ -28,10 +28,6 @@ namespace NovelRT::Audio {
     ALint _soundSourceState;
     SoundBank _soundStorage;
     SoundBank _bufferStorage;
-    
-
-
-
 
     ALuint readFile(std::string input);
     std::string getALError();
@@ -53,14 +49,10 @@ namespace NovelRT::Audio {
     void stopMusic();
     void setMusicVolume(float value);
     void checkSources();
-
-    //For touhou sample
-    //Returns source
     ALuint loadSound(std::string input);
     void unload(ALuint handle);
     void playSound(ALuint handle, int loops);
     void stopSound(ALuint handle);
-
 
   };
 }
--- a/include/NovelRT/Audio/AudioService.h
+++ b/include/NovelRT/Audio/AudioService.h
@@ -15,18 +15,19 @@ namespace NovelRT::Audio {
     const ALuint _noBuffer = 0;
     const ALfloat _pitch = 1.0f;
 
+    std::vector<ALuint> _soundStorage;
     Utilities::Lazy<std::unique_ptr<ALCdevice, void(*)(ALCdevice*)>> _device;
     Utilities::Lazy<std::unique_ptr<ALCcontext, void(*)(ALCcontext*)>> _context;
     LoggingService _logger;
     ALuint _musicSource;
     ALint _musicSourceState;
     ALint _musicLoopAmount;
-    ALuint _soundSource;
     ALint _soundSourceState;
     ALint _soundLoopAmount;
     SoundBank _sounds;
     MusicBank _music;
     std::string _deviceName;
+    bool _manualLoad;
 
     ALuint readFile(std::string input);
     std::string getALError();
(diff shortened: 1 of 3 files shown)


include/NovelRT/Audio/AudioService.h
```diff
@@ -17,16 +17,21 @@ namespace NovelRT::Audio {
...
-    ALint _soundSourceState;
     ALint _soundLoopAmount;
-    SoundBank _sounds;
-    MusicBank _music;
-    std::string _deviceName;
+    ALint _soundSourceState;
+    SoundBank _soundStorage;
+    SoundBank _bufferStorage;
+    
+
```
tannergooding: Nit: several stray newlines

include/NovelRT/Audio/AudioService.h
```diff
@@ -38,18 +43,25 @@ namespace NovelRT::Audio {
...
     void checkSources();
+
+    //For touhou sample
+    //Returns source
+    ALuint loadSound(std::string input);
+    void unload(ALuint handle);
+    void playSound(ALuint handle, int loops);
+    void stopSound(ALuint handle);
+
+
```
tannergooding: Same nit

samples/main.cpp
```diff
@@ -69,8 +69,11 @@ int main(int /*argc*/, char* /*argv*/[])
   auto console = NovelRT::LoggingService(NovelRT::Utilities::Misc::CONSOLE_LOG_APP);
   auto audio = runner.getAudioService();
   audio.lock()->initializeAudio();
-  //auto bgm = audio->load((soundsDirPath / "running.ogg").string(), true);
-  //auto jojo = audio->load((soundsDirPath / "jojo.ogg").string(), false);
+  auto bgm = audio.lock()->loadMusic((soundsDirPath / "marisa.ogg").string());
```
tannergooding: What is the behavior if the file cant be found?

samples/main.cpp
```diff
@@ -69,8 +69,11 @@ int main(int /*argc*/, char* /*argv*/[])
...
+  auto jojo = audio.lock()->loadSound((soundsDirPath / "caution.wav").string());
```
tannergooding: Also, as a note: this is synchronous and causes delays for large files. Something well want to consider how to handle eventually

samples/main.cpp
```diff
@@ -181,14 +184,18 @@ int main(int /*argc*/, char* /*argv*/[])
...
-  memeInteractionRect->Interacted += [&console] {
+  memeInteractionRect->Interacted += [&] {
     console.logDebug("WAHEYYY");
+    audio.lock()->playSound(jojo, 0);
+    audio.lock()->playSound(jojo2, 0);
+    audio.lock()->playSound(jojo3, 0);
+    audio.lock()->playSound(jojo4, 0);
   };
 
   //If uncommenting the call, pass &jojo to the subscription next to &audio.
```
tannergooding: Nit: comment no longer relevant

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -332,24 +260,79 @@ std::string AudioService::getALError() {
   }
 }
 
+//for touhou project
```
tannergooding: Is comment still necessary?

samples/main.cpp
RubyNova: This needs resolving before I'll approve.

src/NovelRT/Audio/AudioService.cpp
RubyNova: Please remove this comment. This implementation is legacy and is getting nuked anyway, so I don't really need to know why were doing things the way we are in here.

samples/main.cpp
capnkenny: Previously it would throw an error - I've since corrected this to specify the file that cannot be found and to also let the user know _why_ the file couldn't be loaded (in a case of i/o error, etc).

Also, added support so that if the service detects the buffer wasn't created, it'll warn and return instead of killing the engine.

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -84,7 +85,7 @@ std::vector<ALuint>::iterator AudioService::loadMusic(std::string input) {
 
   //Sorry Matt, nullptr types are incompatible to ALuint according to VS.
   if (newBuffer == _noBuffer) {
-    _logger.logError("Could not load audio file: ", getALError());
+    _logger.logWarning("Could not load audio file: " + input);
```
capnkenny: Just wanted to point out here - for some reason the LoggingService iterator does not appear to work. FWIW, `getALError()` was incorrect in this context and was always just producing a blank string, however adding a string such as `_logger.logWarning(string1, string2);` would produce a log message only with `string1`. For now, I resulted to concatenation so that it would work.

PR #165, opened 2020-03-04
Added new Timestamp type.

RubyNova:
Title.

### DIFF
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -143,7 +143,7 @@ namespace NovelRT::Windowing {
 
 //value types
 #include "NovelRT/Atom.h"
-#include "NovelRT/Timing/DeltaTimestamp.h"
+#include "NovelRT/Timing/Timestamp.h"
 #include "NovelRT/Utilities/Event.h" //these have to exist up here due to inclue order issues
 #include "NovelRT/Utilities/Lazy.h"
 #include "NovelRT/Utilities/Misc.h"
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -127,6 +127,8 @@ namespace NovelRT::SceneGraph {
 }
 
 namespace NovelRT::Timing {
+  // Windows and some other platforms use 100ns ticks
+  static const uint64_t TicksPerSecond = 10'000'000;
   typedef class StepTimer StepTimer;
 }
 
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -141,6 +141,7 @@ namespace NovelRT::Windowing {
 
 //value types
 #include "NovelRT/Atom.h"
+#include "NovelRT/Timing/DeltaTimestamp.h"
 #include "NovelRT/Utilities/Event.h" //these have to exist up here due to inclue order issues
 #include "NovelRT/Utilities/Lazy.h"
 #include "NovelRT/Utilities/Misc.h"
--- a/include/NovelRT/Timing/DeltaTimestamp.h
+++ b/include/NovelRT/Timing/DeltaTimestamp.h
@@ -1,0 +1,34 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+
+#ifndef NOVELRT_TIMING_DELTATIMESTAMP_H
+#define NOVELRT_TIMING_DELTATIMESTAMP_H
+
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Timing {
+  class DeltaTimestamp {
+  private:
+    const uint64_t _tickDelta;
+    const double _doubleDelta;
+    const float _floatDelta;
+  public:
+    DeltaTimestamp(uint64_t tickDelta, double doubleDelta, float floatDelta);
+
+    inline uint64_t getTickDelta() {
+      return _tickDelta;
+    }
+
+    inline double getDoubleDelta() {
+      return _doubleDelta;
+    }
+
+    inline float getFloatDelta() {
+      return _floatDelta;
+    }
+
+  };
+}
+
+#endif //!NOVELRT_TIMING_DELTATIMESTAMP_H
--- a/include/NovelRT/Timing/DeltaTimestamp.h
+++ b/include/NovelRT/Timing/DeltaTimestamp.h
@@ -8,24 +8,23 @@
 #endif
 
 namespace NovelRT::Timing {
-  class DeltaTimestamp {
+  class Timestamp {
   private:
     const uint64_t _tickDelta;
-    const double _doubleDelta;
-    const float _floatDelta;
-  public:
-    DeltaTimestamp(uint64_t tickDelta, double doubleDelta, float floatDelta);
 
-    inline uint64_t getTickDelta() {
+  public:
+    Timestamp(uint64_t ticks);
+
+    inline uint64_t getTickDelta() const {
       return _tickDelta;
     }
 
-    inline double getDoubleDelta() {
-      return _doubleDelta;
+    inline double getDoubleDelta() const {
+      return static_cast<double>(_tickDelta) / TicksPerSecond;
     }
 
-    inline float getFloatDelta() {
-      return _floatDelta;
+    inline float getFloatDelta() const {
+      return static_cast<float>(_tickDelta) / TicksPerSecond;
     }
 
   };
--- a/include/NovelRT/Timing/StepTimer.h
+++ b/include/NovelRT/Timing/StepTimer.h
@@ -91,7 +91,7 @@ namespace NovelRT::Timing {
     }
 
     void resetElapsedTime();
-    void tick(const Utilities::Event<double>& update);
+    void tick(const Utilities::Event<DeltaTimestamp>& update);
   };
 }
 
--- a/include/NovelRT/Timing/StepTimer.h
+++ b/include/NovelRT/Timing/StepTimer.h
@@ -57,11 +57,11 @@ namespace NovelRT::Timing {
       _targetElapsedTicks = value;
     }
 
-    inline Timestamp getTargetElapsedSeconds() const {
+    inline Timestamp getTargetElapsedTime() const {
       auto targetElapsedTicks = getTargetElapsedTicks();
       return Timestamp(targetElapsedTicks);
     }
-    inline void setTargetElapsedSeconds(double value) {
+    inline void setTargetElapsedTime(double value) {
       auto targetElapsedTicks = SecondsToTicks(value);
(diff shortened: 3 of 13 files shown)


RubyNova:
I do wonder if we should move the construction of the double and float deltas inside the type itself.

Are we happy with the current implementation or should I move stuff around?

include/NovelRT/Timing/DeltaTimestamp.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef NOVELRT_TIMING_DELTATIMESTAMP_H
+#define NOVELRT_TIMING_DELTATIMESTAMP_H
+
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Timing {
+  class DeltaTimestamp {
```
tannergooding: nit: This should probably be called just `Timestamp` or `TimeSpan` or something.

It doesn't strictly speaking have to be a delta, it just represents some amount of time in 100ns ticks.

include/NovelRT/Timing/DeltaTimestamp.h
```diff
@@ -0,0 +1,34 @@
...
+  private:
+    const uint64_t _tickDelta;
+    const double _doubleDelta;
+    const float _floatDelta;
```
tannergooding: nit: `_floatDelta` and `_doubleDelta` should be dynamically computed.

It isn't expensive to do and should happen only a handful of times per frame at most.

include/NovelRT/Timing/DeltaTimestamp.h
```diff
@@ -0,0 +1,34 @@
...
+
+    inline uint64_t getTickDelta() {
+      return _tickDelta;
+    }
+
+    inline double getDoubleDelta() {
+      return _doubleDelta;
+    }
+
+    inline float getFloatDelta() {
```
tannergooding: rather than `getFloat` it should probably be called `getSeconds` or something similar (to make it clear the unit the result represents, etc).

src/NovelRT/Timing/DeltaTimestamp.cpp
```diff
@@ -0,0 +1,9 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+
+#include <NovelRT.h>
+
+namespace NovelRT::Timing {
+
+  DeltaTimestamp::DeltaTimestamp(uint64_t tickDelta, double doubleDelta, float floatDelta) : _tickDelta(tickDelta), _doubleDelta(doubleDelta), _floatDelta(floatDelta) {}
```
tannergooding: nit: The actual implementation is trivial, I don't see any reason why it needs to be part of the resulting library rather than just in the header file.

RubyNova: Because, by convention, weve always put constructors into a source file.

This would become the exception to change that now, no?

include/NovelRT.h
```diff
@@ -127,6 +127,8 @@ namespace NovelRT::SceneGraph {
 }
 
 namespace NovelRT::Timing {
+  // Windows and some other platforms use 100ns ticks
+  static const uint64_t TicksPerSecond = 10'000'000;
```
RubyNova: I moved it to here because of include order issues and having the same constant twice seemed silly.

src/NovelRT/Timing/DeltaTimestamp.cpp
tannergooding: No, there are a few places where it is pure header, namely around simple types like this.

We probably want to do it for other simple types as well as it will help codegen.

RubyNova: Alright I shall move it then. :)

tests/NovelRT.Tests/Timing/TimestampTest.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+class TimestampTest : public testing::Test {
+protected:
+  Timestamp _timeStamp;
+
+  void SetUp() override {
+    _timeStamp = Timestamp(TicksPerSecond);
+  }
+public:
+  TimestampTest::TimestampTest() : _timeStamp(Timestamp(0)) {}
```
RubyNova: This is needed because `Timestamp` does not have a parameterless ctor.

include/NovelRT/Timing/Timestamp.h
```diff
@@ -0,0 +1,33 @@
...
+#define NOVELRT_TIMING_DELTATIMESTAMP_H
+
+#ifndef NOVELRT_H
+#error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
+#endif
+
+namespace NovelRT::Timing {
+  class Timestamp {
+  private:
+    uint64_t _ticks;
```
RubyNova: I removed the `const` qualifier from here due to a deleted function which meant I couldn't assign it to a private field, lol. Woops.

tests/NovelRT.Tests/Timing/TimestampTest.cpp
tannergooding: Why not add one which initializes the ticks to zero?

include/NovelRT/Timing/Timestamp.h
tannergooding: Could you elaborate as to which function is preventing this from being effectively readonly?

tests/NovelRT.Tests/Timing/TimestampTest.cpp
RubyNova: Because, given the current API, it can't be reassigned a new tick value later. Do you want me to support this?

include/NovelRT/Timing/Timestamp.h
RubyNova: in the tests, `_timeStamp = TimeStamp(0)` means that I'd need to define a custom `=` operator, which I didn't feel like doing, especially when the API surface effectively makes this `const` anyway.

include/NovelRT/Timing/StepTimer.h
```diff
@@ -43,13 +41,13 @@ namespace NovelRT::Timing {
       return _totalTicks;
     }
 
-    inline double getElapsedSeconds() const {
+    inline Timestamp getElapsedSeconds() const {
```
tannergooding: This (and the below) should probably be called `getElapsedTime()` and `getTotalTime()` now.

tannergooding: Same with `getTargetElapsedSeconds` to `getTargetElapsedTime`

include/NovelRT/Timing/Timestamp.h
tannergooding: Ah right, because C++ assignment for `const` isn't the same as C# with `readonly`.

I don't remember the right syntax for this...

RubyNova: I don't either! 😄

src/NovelRT/Timing/Timestamp.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+
+#include <NovelRT.h>
```
tannergooding: File can be deleted?

RubyNova: Uhhh...I already deleted this. Why is it still here? Let me double check.

RubyNova: Ok, somehow this re-added itself. Thanks Git. 😒

include/NovelRT/Timing/Timestamp.h
```diff
@@ -0,0 +1,33 @@
...
+    inline uint64_t getTicks() const {
+      return _ticks;
+    }
+
+    inline double getSecondsDouble() const {
+      return static_cast<double>(_ticks) / TicksPerSecond;
+    }
+
+    inline float getSecondsFloat() const {
+      return static_cast<float>(_ticks) / TicksPerSecond;
```
tannergooding: This is going to need to be `static_cast<float>(getSecondsDouble())`. The former should likely also be called just `getSeconds()`.

It may also be good to assert the `float` result is not zero if the `double` result is not zero.

tannergooding: For reference... 

The issue here is that `float` can only represent integers up to 2^24 accurately which is only `1.6777216 seconds` if you are using 100ns ticks.
After that, you start getting more and more off in the computed result.

`double` on the other hand can represent integers up to 2^53, which is `10424.999137431703 days`. Once you scale those down by `TicksPerSecond` and then convert to float, they will generally be within a range where the minor precision loss doesn't matter.

include/NovelRT/Timing/StepTimer.h
```diff
@@ -59,11 +57,11 @@ namespace NovelRT::Timing {
...
     }
 
-    inline double getTargetElapsedSeconds() const {
+    inline Timestamp getTargetElapsedTime() const {
       auto targetElapsedTicks = getTargetElapsedTicks();
-      return TicksToSeconds(targetElapsedTicks);
+      return Timestamp(targetElapsedTicks);
     }
-    inline void setTargetElapsedSeconds(double value) {
+    inline void setTargetElapsedTime(double value) {
```
tannergooding: This should take in a `Timestamp` not a `double`, so it correctly mirrors `getTargetElapsedTime`.

`SecondsToTicks` and `TicksToSeconds` should then likely be moved onto `Timestamp` instead of being here.

include/NovelRT/Timing/StepTimer.h
```diff
@@ -59,11 +57,11 @@ namespace NovelRT::Timing {
       _targetElapsedTicks = value;
     }
 
-    inline double getTargetElapsedSeconds() const {
+    inline Timestamp getTargetElapsedTime() const {
       auto targetElapsedTicks = getTargetElapsedTicks();
```
tannergooding: Maybe should be part of the API review, but it might be better if `getTargetElapsedTicks` wasn't exposed to the user
and instead users just get the timestamp and call `getTicks` on that
The compiler will optimize it the same, but it encourages users to work with `Timestamp` instead, which is likely the "more correct" thing

tannergooding: (same with the other `get*Ticks` methods in here)

include/NovelRT/Timing/StepTimer.h
RubyNova: Resolved as per the Discord discussion with new ctor.

include/NovelRT/Timing/StepTimer.h
RubyNova: Marked it for discussion for the review.

PR #176, opened 2020-03-26
Add some documentation.

jeuxjeux20:
It's just great to have some helpful (or not) text in a popup when hovering...

### DIFF
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -81,27 +81,39 @@
 //libpng
 #include <png.h>
 
+/**
+ * NovelRT is a cross-platform, flexible Visual Novel and 2D game engine.
+ * It is aimed at designers and developers alike, however many of the designer tools and features we have on our roadmap have yet to be implemented.
+ */
 namespace NovelRT {
   typedef class Atom Atom;
   typedef class DebugService DebugService;
   typedef class LoggingService LoggingService;
   typedef class NovelRunner NovelRunner;
 }
-
+/**
+ * Contains animation features.
+ */
 namespace NovelRT::Animation {
   typedef class SpriteAnimator SpriteAnimator;
 }
-
+/***
+ * Contains audio features, such as playing audio, and managing audio resources.
+ */
 namespace NovelRT::Audio {
   typedef std::vector<ALuint> SoundBank;
   typedef std::vector<ALuint> MusicBank;
   typedef class AudioService AudioService;
 }
-
+/**
+ * Contains features to interop with the .NET Core runtime.
+ */
 namespace NovelRT::DotNet {
   typedef class RuntimeService RuntimeService;
 }
-
+/**
+ * Contains graphics features, such as rendering, textures, cameras, etc...
+ */
 namespace NovelRT::Graphics {
   typedef class Texture Texture;
   typedef class BasicFillRect BasicFillRect;
@@ -111,13 +123,17 @@ namespace NovelRT::Graphics {
   typedef class RenderObject RenderObject;
   typedef class TextRect TextRect;
 }
-
+/**
+ * Contains input features, such as keyboard and mouse interactivity.
+ */
 namespace NovelRT::Input {
   typedef class BasicInteractionRect BasicInteractionRect;
   typedef class InteractionObject InteractionObject;
   typedef class InteractionService InteractionService;
 }
-
+/**
+ * Contains scene graph features.
+ */
 namespace NovelRT::SceneGraph {
   typedef class QuadTreeNode QuadTreeNode;
   typedef class QuadTreeScenePoint QuadTreeScenePoint;
@@ -125,13 +141,17 @@ namespace NovelRT::SceneGraph {
   typedef class Scene Scene;
   typedef class SceneNode SceneNode;
 }
-
+/**
+ * Contains timing features, such as timers.
+ */
 namespace NovelRT::Timing {
   // Windows and some other platforms use 100ns ticks
   static const uint64_t TicksPerSecond = 10'000'000;
   typedef class StepTimer StepTimer;
 }
-
+/**
+ * Contains windowing features.
+ */
 namespace NovelRT::Windowing {
   typedef class WindowingService WindowingService;
 }
@@ -144,7 +164,7 @@ namespace NovelRT::Windowing {
 //value types
 #include "NovelRT/Atom.h"
 #include "NovelRT/Timing/Timestamp.h"
-#include "NovelRT/Utilities/Event.h" //these have to exist up here due to inclue order issues
+#include "NovelRT/Utilities/Event.h" //these have to exist up here due to include order issues
 #include "NovelRT/Utilities/Lazy.h"
 #include "NovelRT/Utilities/Misc.h"
 
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -112,7 +112,7 @@ namespace NovelRT::DotNet {
   typedef class RuntimeService RuntimeService;
 }
 /**
- * Contains graphics features, such as rendering, textures, cameras, etc...
+ * Contains graphics features, such as rendering, textures, cameras, etc.
  */
 namespace NovelRT::Graphics {
   typedef class Texture Texture;
--- a/include/NovelRT/Audio/AudioService.h
+++ b/include/NovelRT/Audio/AudioService.h
@@ -7,8 +7,8 @@
 #error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
 #endif
 
-//TODO: This won't exist after Kenny's rewrite, not too bothered about this class.
 namespace NovelRT::Audio {
+  // TODO: This won't exist after Kenny's rewrite, not too bothered about this class.
   class AudioService {
   private:
     const size_t _bufferSize = 2048;
--- a/include/NovelRT/Input/KeyCode.h
+++ b/include/NovelRT/Input/KeyCode.h
@@ -9,7 +9,7 @@
 
 namespace NovelRT::Input {
   /**
-   * A set of key codes, with their OpenGL values.
+   * A set of key codes.
    */
   enum class KeyCode : int32_t {
     Nothing = GLFW_KEY_UNKNOWN,
--- a/include/NovelRT/Input/KeyCode.h
+++ b/include/NovelRT/Input/KeyCode.h
@@ -8,6 +8,9 @@
 #endif
 
 namespace NovelRT::Input {
+  /**
+   * A set of key codes, with their OpenGL values.
+   */
   enum class KeyCode : int32_t {
     Nothing = GLFW_KEY_UNKNOWN,
     Enter = GLFW_KEY_ENTER,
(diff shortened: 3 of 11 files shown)


include/NovelRT/Input/KeyCode.h
```diff
@@ -8,6 +8,9 @@
 #endif
 
 namespace NovelRT::Input {
+  /**
+   * A set of key codes, with their OpenGL values.
```
RubyNova: Incorrect. These are the values assigned by GLFW, but that's an implementation detail and doesn't need to be shared.

include/NovelRT/Transform.h
```diff
@@ -16,42 +19,79 @@ namespace NovelRT {
     bool _active;
 
   public:
+    /**
+     * Creates a new Transform, with the specified position, rotation and scale.
+     * @param position The position.
+     * @param rotation The rotation.
+     * @param scale The scale.
```
RubyNova: Maybe describe these a little more? i.e. `scale` represents the size, for example.

include/NovelRT/Transform.h
```diff
@@ -16,42 +19,79 @@ namespace NovelRT {
...
 
+    /**
+     * Gets the bounds of this transform.
+     */
     inline Maths::GeoBounds getBounds() const {
       return Maths::GeoBounds(_position, _scale, _rotation);
     }
 
+    /**
+     * Gets the constant position of this transform.
```
RubyNova: Im unsure if "constant" is the right word here, but this is something we could discuss.

include/NovelRT/Transform.h
```diff
@@ -16,42 +19,79 @@ namespace NovelRT {
...
 
+    /**
+     * Gets the position of this transform.
+     */
     inline Maths::GeoVector2<float>& position() {
       return _position;
     }
 
+    /**
+     * Gets the constant rotation of this transform.
```
RubyNova: Same here with the use of the word "constant". just doesn't...sound right?

include/NovelRT/Transform.h
```diff
@@ -16,42 +19,79 @@ namespace NovelRT {
...
 
+    /**
+     * Gets the rotation of this transform.
+     */
     inline float& rotation() {
       return _rotation;
     }
 
+    /**
+     * Gets the constant scale of this transform.
```
RubyNova: Same here with the use of the word "constant". just doesn't...sound right?

include/NovelRT/Transform.h
```diff
@@ -16,42 +19,79 @@ namespace NovelRT {
...
 
+    /**
+     * Gets the scale of this transform.
+     */
     inline Maths::GeoVector2<float>& getScale() {
       return _scale;
     }
 
+    /**
+     * Sets the scale of this transform to the specified value.
```
RubyNova: Wow how does this method even still exist? You can remove the docs on this one, this should've been nuked.

include/NovelRT/Input/KeyCode.h
jeuxjeux20: Oh literally just read "GL" and though that was OpenGL, i'm just going to remove that

include/NovelRT/Transform.h
jeuxjeux20: Actually it's just because the whole method returned a `const`, and because for some awkward reason resharper does not like me to do `@copyDoc Transform::position`, I wanted to make it a bit... different? Anyway, `constant` is going away (and this wasn't an intended pun I swear)

include/NovelRT/Transform.h
jeuxjeux20: I'll try

include/NovelRT/Transform.h
jeuxjeux20: Going away as well

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
   class Transform {
   protected:
     Maths::GeoVector2<float> _position;
     Maths::GeoVector2<float> _scale;
     float _rotation;
 
   public:
+    /**
+     * Creates a new Transform, with the specified position, rotation and scale.
+     * @param position The position, which represents the position where the transform will be rendered.
```
RubyNova: Transforms can be used for other things besides rendering. Perhaps `which represents the position of the transform in a given space such as screen or world space.` or something similar would be more correct? Not sure I like the wording of my example, but hopefully my point makes sense.

jeuxjeux20: I saw that when looking at Unity docs too, just wasn't sure if there was still a world/screen space thing so i left that out. Just put that.

include/NovelRT/Audio/AudioService.h
```diff
@@ -7,8 +7,8 @@
 #error Please do not include this directly. Use the centralised header (NovelRT.h) instead!
 #endif
 
-//TODO: This won't exist after Kenny's rewrite, not too bothered about this class.
 namespace NovelRT::Audio {
+  // TODO: This won't exist after Kenny's rewrite, not too bothered about this class.
```
jeuxjeux20: Because **this** happened:

image

include/NovelRT.h
```diff
@@ -81,27 +81,39 @@
...
-
+/**
+ * Contains features to interop with the .NET Core runtime.
+ */
 namespace NovelRT::DotNet {
   typedef class RuntimeService RuntimeService;
 }
-
+/**
+ * Contains graphics features, such as rendering, textures, cameras, etc...
```
RubyNova: Nit: that can just be `etc.` since `etc...` reads a bit weirdly for documentation.

include/NovelRT/NovelRunner.h
```diff
@@ -14,7 +14,13 @@ namespace NovelRT {
   class NovelRunner {
 
   public:
+    /**
+     * An event that occurs when the scene has to be (re)constructed.
```
RubyNova: I think better wording might be `This event is used for constructing the rendered scene. This is called once per frame update.`

include/NovelRT/NovelRunner.h
```diff
@@ -14,7 +14,13 @@ namespace NovelRT {
...
+     */
     Utilities::Event<> SceneConstructionRequested;
+    /**
+     * An event that occurs on a frame update, with the delta Timestamp between the last frame update and the new one.
```
RubyNova: You could probably add on here, `Use this to define game behaviour that is required to be done per-frame.`, just to make it completely clear.

include/NovelRT/NovelRunner.h
```diff
@@ -41,18 +47,24 @@ namespace NovelRT {
      * Starts the visual novel.
```
RubyNova: lol I've just seen this - could you update this to specify that it launches NovelRT/the NovelRT game loop/the game instead? Much love :D

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
+   */
   class Transform {
   protected:
     Maths::GeoVector2<float> _position;
     Maths::GeoVector2<float> _scale;
     float _rotation;
 
   public:
+    /**
+     * Creates a new Transform, with the specified position, rotation and scale.
```
RubyNova: Grammar nit: Don't need the first comma here.

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
+     * @param position The position, which represents the position of the transform in a given space such as screen or world space.
+     * @param rotation The rotation, in degrees, which represents the rotation of the transform.
```
RubyNova: Grammar nit: Don't need the first comma here.

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
+    /**
+     * Gets the Axis Aligned Bounding Box based on the position and scale of this transform.
+     */
     inline Maths::GeoBounds getAABB() const {
       auto scale = fmaxf(_scale.getX(), _scale.getY());
       return Maths::GeoBounds(_position, Maths::GeoVector2(scale, scale), 0);
     }
 
+    /**
+     * Gets the bounds, based on the position, scale and rotation of this transform.
```
RubyNova: Grammar nit: Don't need the first comma here.

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
+     */
     inline Maths::GeoBounds getBounds() const {
       return Maths::GeoBounds(_position, _scale, _rotation);
     }
 
+    /**
+     * Gets the position of this transform.
```
RubyNova: Maybe add to this, `This is the read-only version.`? Or something to indicate this is the version that gets used in `const` contexts. :D

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
+     */
     inline Maths::GeoVector2<float>& position() {
       return _position;
     }
 
+    /**
+     * Gets the rotation, in degrees, of this transform.
```
RubyNova: Maybe add to this, `This is the read-only version.`? Or something to indicate this is the version that gets used in `const` contexts. :D

include/NovelRT/Transform.h
```diff
@@ -8,41 +8,74 @@
...
     inline float& rotation() {
       return _rotation;
     }
 
+    /**
+     * Gets the scale of this transform.
+     */
     inline const Maths::GeoVector2<float>& scale() const {
       return _scale;
     }
```
RubyNova: To clarify - all of these are basically the way were handling auto properties. Not sure if this changes much.

Since you're here, could you also add a non-const version of `scale()`? I forgot to add that back in after nuking `setScale()` and don't wanna cause merge hell for you again.

include/NovelRT.h
jeuxjeux20: Hm, I'm used to doing `etc...` but makes sense.

include/NovelRT/NovelRunner.h
jeuxjeux20: Will change

include/NovelRT/NovelRunner.h
jeuxjeux20: Will add 👌

include/NovelRT/NovelRunner.h
jeuxjeux20: Sure 👌👌

include/NovelRT/Transform.h
jeuxjeux20: Forgot to delete it

include/NovelRT/Transform.h
jeuxjeux20: Forgot to delete it 2

include/NovelRT/Transform.h
jeuxjeux20: Forgot to delete it 3

include/NovelRT/Transform.h
jeuxjeux20: Ah yeah, my previous attempt was with "constant" 😅

include/NovelRT/Transform.h
jeuxjeux20: Ah yeah, my previous attempt was with "constant" 😅 2

include/NovelRT/Transform.h
jeuxjeux20: Ah yeah, my pre... 

Yeah I know, like std does with `vector`.

PR #179, opened 2020-04-02
macOS Support

capnkenny:
Merging this PR would introduce basic support for building NovelRT, its tests and samples for macOS 10.15 (Catalina).

- Based on OpenGL 4.1 Core
- Earlier macOS compatibility may/may not be feasible due to our inclusion of `std::filesystem` - this is only supported on 10.15 due to it being the first release that macOS includes `std::filesystem` in `libc++` (There are Mac-specific alternatives though that would allow earlier ones to work)
- Requires separate shaders due to version compatibility (more details in comments)

### DIFF
--- a/include/NovelRT.h
+++ b/include/NovelRT.h
@@ -38,6 +38,10 @@
 #include <Windows.h>
 #endif
 
+#if defined(__APPLE__)
+#include <mach-o/dyld.h>
+#endif
+
 //Freetype
 #include <ft2build.h>
 #include FT_FREETYPE_H
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -19,9 +19,14 @@ set(NOVELRT_LIB_LINK_LIBRARIES
   glfw
   glm
   OpenAL::OpenAL
-  png
   spdlog::spdlog_header_only
 )
+
+if (APPLE)
+  set(NOVELRT_LIB_LINK_LIBRARIES ${NOVELRT_LIB_LINK_LIBRARIES} png_static)
+else()
+  set(NOVELRT_LIB_LINK_LIBRARIES ${NOVELRT_LIB_LINK_LIBRARIES} png)
+endif()
 
 if (WIN32)
   set(NOVELRT_LIB_LINK_LIBRARIES ${NOVELRT_LIB_LINK_LIBRARIES} sndfile-shared)
@@ -47,9 +52,11 @@ link_libraries(${NETHOST_LIBRARY})
 
 if (WIN32)
   set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
-else()
+elseif(UNIX AND NOT APPLE)
   link_libraries(m)
   link_libraries(stdc++fs)
+else()
+  link_libraries(m)
 endif()
 
 add_library(NovelRT SHARED ${NOVELRT_LIB_HEADERS} ${NOVELRT_LIB_SOURCES})
--- a/src/NovelRT/Graphics/RenderingService.cpp
+++ b/src/NovelRT/Graphics/RenderingService.cpp
@@ -57,10 +57,15 @@ namespace NovelRT::Graphics {
 
       glEnable(GL_BLEND);
       glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-
+#if defined(__APPLE__)
+        _basicFillRectProgram = loadShaders("BasicVertexShader_Apple.glsl", "BasicFragmentShader_Apple.glsl");
+        _texturedRectProgram = loadShaders("TexturedVertexShader_Apple.glsl", "TexturedFragmentShader_Apple.glsl");
+        _fontProgram = loadShaders("FontVertexShader_Apple.glsl", "FontFragmentShader_Apple.glsl");
+#else
       _basicFillRectProgram = loadShaders("BasicVertexShader.glsl", "BasicFragmentShader.glsl");
       _texturedRectProgram = loadShaders("TexturedVertexShader.glsl", "TexturedFragmentShader.glsl");
       _fontProgram = loadShaders("FontVertexShader.glsl", "FontFragmentShader.glsl");
+#endif
     }
     else {
       _camera->forceResize(windowSize);
--- a/src/NovelRT/Utilities/Misc.cpp
+++ b/src/NovelRT/Utilities/Misc.cpp
@@ -26,6 +26,25 @@ namespace NovelRT::Utilities {
 
       path[pathLength] = L'\0';
       return path;
+#elif defined(__APPLE__)
+      char *path;
+      uint32_t pathLength = 0;
+      _NSGetExecutablePath(nullptr, &pathLength);
+      path = new char[pathLength];
+      if ((pathLength <= 0) || (pathLength >= PATH_MAX))
+      {
+          return std::filesystem::current_path();
+      }
+      
+      if (_NSGetExecutablePath(path, &pathLength) < 0)
+      {
+          return std::filesystem::current_path();
+      }
+      
+      char *actualPath = realpath(path, nullptr);
+      free(path);
+      
+      return std::string(actualPath).c_str();
 #else
       char path[PATH_MAX + 1];
       auto pathLength = readlink("/proc/self/exe", path, PATH_MAX);
--- a/src/NovelRT/Windowing/WindowingService.cpp
+++ b/src/NovelRT/Windowing/WindowingService.cpp
@@ -36,12 +36,27 @@ namespace NovelRT::Windowing {
       checkForOptimus(OptimusLibraryName);
 #endif
 
+      
+#if defined(__APPLE__)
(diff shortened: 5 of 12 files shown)


src/CMakeLists.txt
```diff
@@ -47,9 +52,11 @@ link_libraries(${NETHOST_LIBRARY})
 
 if (WIN32)
   set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
-else()
+elseif(UNIX AND NOT APPLE)
   link_libraries(m)
   link_libraries(stdc++fs)
+else()
+  link_libraries(m)
```
capnkenny, line 59: Apple includes `std::filesystem` in `libc++` so there's no external linking needed here.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -36,13 +36,28 @@ namespace NovelRT::Windowing {
...
+      {
+          _logger.throwIfNullPtr(window, "Failed to create OpenGL v4.1 context");
+      }
+#else
     _logger.logInfoLine("Attempting to create OpenGL ES v3.0 context using EGL API");
 
     glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API);
     glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
-    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 0);
+    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
```
capnkenny, line 57: Oops, need to revert this line - mistake on my end.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -36,13 +36,28 @@ namespace NovelRT::Windowing {
...
+      glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 1);
+      glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GLFW_TRUE);
+      glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
+      
+      auto window = glfwCreateWindow(wData, hData, windowTitle.c_str(), nullptr, nullptr);
+      if (window == nullptr)
+      {
+          _logger.throwIfNullPtr(window, "Failed to create OpenGL v4.1 context");
+      }
+#else
```
capnkenny, line 52: If-def'ed a much smaller context creation method for macOS - this is the only context that works with our current calls and shaders and needs to be called this way, so if the target macOS device does not support OGL 4.1 for some reason it will fail straight away. Other OSes will continue to follow the same pattern we normally use.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -66,9 +81,7 @@ namespace NovelRT::Windowing {
         glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
         glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
         glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GLFW_TRUE);
-        glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
-        glfwWindowHint(GLFW_CONTEXT_CREATION_API, GLFW_NATIVE_CONTEXT_API);
```
capnkenny, line 70: I'll fix this too - got confused if this was actually part of `master` or not.

include/NovelRT.h
```diff
@@ -38,6 +38,10 @@
 #include <Windows.h>
 #endif
 
+#if defined(__APPLE__)
+#include <mach-o/dyld.h>
+#endif
```
capnkenny, line 43: This is so that we can use Apple's native function to find the executable's file path - this _could_ be moved possibly into `NovelRT::Utilities::Misc`

src/NovelRT/Utilities/Misc.cpp
```diff
@@ -26,6 +26,25 @@ namespace NovelRT::Utilities {
...
+          return std::filesystem::current_path();
+      }
+      
+      char *actualPath = realpath(path, nullptr);
+      free(path);
+      
+      return std::string(actualPath).c_str();
 #else
       char path[PATH_MAX + 1];
       auto pathLength = readlink("/proc/self/exe", path, PATH_MAX);
```
capnkenny, line 50: When using the `readlink()` method on macOS, this returns a directory that is one folder above the actual executable's directory (i.e. "artifacts/build/Debug" instead of "artifacts/build/Debug/samples").

Using Apple's native method in this manner allows us to:
1) find the length of the path, then
2) Get the proper executable path as determined by macOS.

With that being said, there's a possibility to get a symbolic link instead of an actual path, so I used `realpath()` here to ensure an absolute path is always returned as long as `_NSGetExecutablePath()` succeeds.

src/NovelRT/Graphics/RenderingService.cpp
```diff
@@ -57,10 +57,15 @@ namespace NovelRT::Graphics {
...
-
+#if defined(__APPLE__)
+        _basicFillRectProgram = loadShaders("BasicVertexShader_Apple.glsl", "BasicFragmentShader_Apple.glsl");
+        _texturedRectProgram = loadShaders("TexturedVertexShader_Apple.glsl", "TexturedFragmentShader_Apple.glsl");
+        _fontProgram = loadShaders("FontVertexShader_Apple.glsl", "FontFragmentShader_Apple.glsl");
+#else
       _basicFillRectProgram = loadShaders("BasicVertexShader.glsl", "BasicFragmentShader.glsl");
       _texturedRectProgram = loadShaders("TexturedVertexShader.glsl", "TexturedFragmentShader.glsl");
       _fontProgram = loadShaders("FontVertexShader.glsl", "FontFragmentShader.glsl");
+#endif
```
capnkenny, line 68: Did this so that Apple will always pull its versions of the shaders instead of failing to load the normal ones and crashing.

src/Resources/Shaders/BasicFragmentShader_Apple.glsl
```diff
@@ -0,0 +1,8 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root for more information.
+#version 330
```
capnkenny, line 2: Since we use OGL 4.1 Core _and we're on Apple hardware_, GLSL version 300 ES is not supported - we move over to version 330 to maintain compatibility (specifically in the Apple-designated shaders only).

capnkenny:
There's also some newline nits I need to fix but this is mainly to see viability of supporting future builds as part of our mainstream branch.

include/NovelRT.h
RubyNova: This looks like a system include, so right under the windows.h linkage is fine.

src/CMakeLists.txt
RubyNova: 👍

src/NovelRT/Graphics/RenderingService.cpp
RubyNova: We might be able to handle this in a nicer way in the future, but I understand. Nice!

src/NovelRT/Utilities/Misc.cpp
RubyNova: Gotcha.

src/NovelRT/Windowing/WindowingService.cpp
RubyNova: How come we cannot check for other contexts, exactly?

src/Resources/Shaders/BasicFragmentShader_Apple.glsl
RubyNova: Makes sense. Not sure I understand their OGL support much, Mac is weird.

src/NovelRT/Windowing/WindowingService.cpp
capnkenny: Although we may be able to check for a OGL 3.2 context, this would set us back to GLSL version 1.40 approximately and would affect our feature set. Also, EGL Contexts as well as Compatibility contexts are not supported by Mac (from my research, I could be wrong). (And calling the wrong context _immediately_ nopes out the `glfwCreateWindow` call)

src/Resources/Shaders/BasicFragmentShader_Apple.glsl
capnkenny: Digging thru some OGL docs, it appears that Version 3.0 is specific to ES and falls between Version 3.30 (supported by OpenGL 4.1) and version 1.40 (Supported by OpenGL 3.2). Very weird indeed.

src/NovelRT/Windowing/WindowingService.cpp
RubyNova: Ah ok, makes sense.

capnkenny:
This shall be re-done, and the CI on it is borked anyways.... closing for the time being 😢

PR #184, opened 2020-04-17
Fixed issue where the input system wasn't updating keys correctly.

### LINKED ISSUE #171: InteractionService does not propagate key state correctly.
Expected behaviour:
When a key is held down, it should only remain in the `KeyDown` state for a single frame, then transition to `KeyDownHeld`

Actual behaviour:
It never goes past `KeyDown`, so `KeyDownHeld` is never produced as a state.

Steps to reproduce:
1 - set up code to check for `KeyDown` only on a specific key with a log.
2 - attach this code to `Update`.
3 - run the software.
4 - press the key you are checking for.
5 - inspect logs and observe the issue.

RubyNova:
Resolves #171


RubyNova:
So this now fully works as per @tannergooding discussing it with me with a new 'buffer based' (actually just std::maps) system for key input caching. I believe the new implementation is much cleaner than my original fix however the resulting source file is a bit messy from debugging and testing. I will clean it up then mark it as ready for review.

include/NovelRT/Input/InteractionService.h
```diff
@@ -13,22 +13,23 @@ namespace NovelRT::Input {
     friend class Windowing::WindowingService; //I get this looks weird but its because GLFW treats the window as this system as well as the window.
 
   private:
-    NovelRunner* const _runner;
+    static inline const uint32_t INPUT_BUFFER_COUNT = 2;
```
tannergooding, line 16: Why hardcode to 2?

include/NovelRT/Input/InteractionService.h
```diff
@@ -13,22 +13,23 @@ namespace NovelRT::Input {
...
 
+    uint32_t _previousBufferIndex;
```
tannergooding, line 18: Why do you need to track previous and current, rather than just current and implicitly determining previous?

include/NovelRT/Input/InteractionService.h
RubyNova: I just picked two as a standard amount. No real reason.

include/NovelRT/Input/InteractionService.h
RubyNova: because implicitly determining previous everywhere is annoying and I prefer to have an actual cached index for it.

include/NovelRT/Input/KeyState.h
```diff
@@ -12,7 +12,7 @@ namespace NovelRT::Input {
    * A set of key states.
    */
   enum class KeyState : int32_t {
-    Idle = 0,
+    Idle = -1,
```
tannergooding, line 15: Why is `Idle = -1`? What is the default key state?

samples/main.cpp
```diff
@@ -163,8 +163,7 @@ int main(int /*argc*/, char* /*argv*/[])
     rotation += rotationAmount * delta.getSecondsFloat();
     novelChanRect->transform().rotation() = rotation;
 
-    if (rotation > 360.0f)
-    {
+    if (rotation > 360.0f) {
```
tannergooding, line 166: shouldn't this be `>= 360`, since `360 == 0`?

include/NovelRT/Input/KeyState.h
RubyNova: it clashed with one of the GLFW key states returning `0`, so default is currently `GLFW_RELEASE`.

samples/main.cpp
RubyNova: No idea, I just fixed the bracing.

include/NovelRT/Input/InteractionService.h
tannergooding: Is this meant to be the same as the render buffer? Should it be configurable?

RubyNova: Maybe? However if we want an API for configuring it that's out of scope for this ticket and we'll need to open another issue, this bug has been hanging around for too long as it is.

include/NovelRT/Input/InteractionService.h
tannergooding: Why not just track a ref to the right index and update it when you need to calculate the current?
Or centralize the logic in a single place? It's really just `(current == 0) ? count - 1 : current - 1`

RubyNova: I already do provide the previous index when I calculate the current, they get cached at the same time. I'm unsure what benefit this would have.

include/NovelRT/Input/KeyState.h
tannergooding: Maybe it should be reordered and commented to make it clear what the GLFW values are:
```
Idle = -1,
KeyUp = GLFW_RELEASE, // 0
...
```

RubyNova: Okay I'll do that quickly then.

include/NovelRT/Input/InteractionService.h
tannergooding: Is it out of scope? It's just taking the count in the constructor, rather than taking a constant here.

include/NovelRT/Input/InteractionService.h
tannergooding: It avoids bloating the struct by an additional 4 bytes for something that is trivially computable.

include/NovelRT/Input/InteractionService.h
RubyNova: It's an API change, which is my point. I dislike mixing API changes within those if possible and this branch purely exists to fix the bug, but I can introduce it if you want. It's not a big deal really I suppose.

include/NovelRT/Input/InteractionService.h
RubyNova: But then I'd need to compute it multiple times in a single pass? Sorry, I'm still not understanding this. I would've thought just calculating and caching ones for the entire pass of the input system to use would have been better than calculating it once in the FSM, calculating it once in `processKeys` and calculating it wherever else its needed.

include/NovelRT/Input/InteractionService.h
tannergooding: It's not a big deal either way. I was just commenting that this is introducing a new hardcoded concept and there isn't a corresponding issue tracking that.

include/NovelRT/Input/InteractionService.h
tannergooding: Why would it need to be computed multiple times? Maybe this discussion would be better for discord?

RubyNova: Sure, I'll move over there.

PR #193, opened 2020-06-20
Fixed memory leak in Graphics::TextRect.

RubyNova:
Title.

### DIFF
--- a/src/NovelRT/Graphics/TextRect.cpp
+++ b/src/NovelRT/Graphics/TextRect.cpp
@@ -39,18 +39,21 @@ namespace NovelRT::Graphics {
   }
   void TextRect::setText(const std::string& value) {
     _text = value;
-    auto difference = _text.length() - _letterRects.size();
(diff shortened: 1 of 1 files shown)


tannergooding:
As per offline chat, using `ptrdiff_t` or checking `if (letterRects.Length < txt.Length)` is likely a better fix.

src/NovelRT/Graphics/TextRect.cpp
```diff
@@ -39,12 +39,12 @@ namespace NovelRT::Graphics {
   }
   void TextRect::setText(const std::string& value) {
     _text = value;
-    int32_t difference = int32_t(_letterRects.size()) - int32_t(_text.length());
 
-    if (difference < 0) {
+    if (_letterRects.size() < _text.size()) {
```
tannergooding, line 43: size here and length just below?

RubyNova: ...oops.

PR #199, opened 2020-07-12
Transparency and Background Colour changing support

capnkenny:
This PR should allow for users to:
1) Define a custom background colour - if not, it will use a default colour of Black (0,0,0,0) or no colour (see below for significance)

2) Pass a bool flag to `NovelRunner` which tells if Framebuffer Transparency is enabled.
- It is important to note that the framebuffer will respect Alpha values and treat an RGBA value of (0,0,0,0) as having _no colour_, hence the framebuffer will be completely transparent if enabled.

### DIFF
--- a/include/NovelRT/Graphics/RenderingService.h
+++ b/include/NovelRT/Graphics/RenderingService.h
@@ -22,14 +22,14 @@ namespace NovelRT::Graphics {
     ShaderProgram _basicFillRectProgram;
     ShaderProgram _texturedRectProgram;
     ShaderProgram _fontProgram;
-
-    RGBAConfig _framebufferColour;
     
     Utilities::Lazy<GLuint> _cameraObjectRenderUbo;
     std::shared_ptr<Camera> _camera;
 
     std::map<Atom, std::weak_ptr<Texture>> _textureCache;
     std::map<Atom, std::weak_ptr<FontSet>> _fontCache;
+
+    RGBAConfig _framebufferColour;
 
     void bindCameraUboForProgram(GLuint shaderProgramId);
 
@@ -39,7 +39,6 @@ namespace NovelRT::Graphics {
   public:
     RenderingService(NovelRunner* const runner);
     int initialiseRendering();
-    
     void tearDown() const;
 
     std::unique_ptr<ImageRect> createImageRect(const Transform& transform, int layer, const std::string& filePath, const RGBAConfig& colourTint = RGBAConfig(255, 255, 255, 255));
@@ -55,7 +54,7 @@ namespace NovelRT::Graphics {
     void beginFrame() const;
     void endFrame() const;
 
-    void setBackgroundColour(const RGBAConfig colour);
+    void setBackgroundColour(const RGBAConfig& colour);
 
     std::shared_ptr<Texture> getTexture(const std::string& fileTarget = "");
     std::shared_ptr<FontSet> getFontSet(const std::string& fileTarget, float fontSize);
--- a/include/NovelRT/Graphics/RenderingService.h
+++ b/include/NovelRT/Graphics/RenderingService.h
@@ -23,6 +23,8 @@ namespace NovelRT::Graphics {
     ShaderProgram _texturedRectProgram;
     ShaderProgram _fontProgram;
 
+    RGBAConfig _framebufferColour;
+    
     Utilities::Lazy<GLuint> _cameraObjectRenderUbo;
     std::shared_ptr<Camera> _camera;
 
@@ -37,7 +39,7 @@ namespace NovelRT::Graphics {
   public:
     RenderingService(NovelRunner* const runner);
     int initialiseRendering();
-
+    
     void tearDown() const;
 
     std::unique_ptr<ImageRect> createImageRect(const Transform& transform, int layer, const std::string& filePath, const RGBAConfig& colourTint = RGBAConfig(255, 255, 255, 255));
@@ -52,6 +54,8 @@ namespace NovelRT::Graphics {
 
     void beginFrame() const;
     void endFrame() const;
+
+    void setBackgroundColour(const RGBAConfig colour);
 
     std::shared_ptr<Texture> getTexture(const std::string& fileTarget = "");
     std::shared_ptr<FontSet> getFontSet(const std::string& fileTarget, float fontSize);
--- a/include/NovelRT/NovelRunner.h
+++ b/include/NovelRT/NovelRunner.h
@@ -23,7 +23,6 @@ namespace NovelRT {
      * Use this to define game behaviour that is required to be done per-frame.
      */
     Utilities::Event<Timing::Timestamp> Update;
-    
   private:
     int _exitCode;
     Utilities::Lazy<std::unique_ptr<Timing::StepTimer>> _stepTimer;
--- a/include/NovelRT/NovelRunner.h
+++ b/include/NovelRT/NovelRunner.h
@@ -23,7 +23,7 @@ namespace NovelRT {
      * Use this to define game behaviour that is required to be done per-frame.
      */
     Utilities::Event<Timing::Timestamp> Update;
-
+    
   private:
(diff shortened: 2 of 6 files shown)


capnkenny:
Note that since the transparency flag was appended to the ctor for `NovelRunner`, it is required to specify `targetFrameRate` at the moment. This can be changed, however if not, specifying `NovelRunner(0, "WindowTitle", true)` will lead to an unintended consequence of `targetFrameRate` being equal to 1.

include/NovelRT/NovelRunner.h
```diff
@@ -43,7 +43,7 @@ namespace NovelRT {
      * @param windowTitle The title of the window created for NovelRunner.
      * @param targetFrameRate The framerate that should be targeted and capped.
      */
-    explicit NovelRunner(int displayNumber, const std::string& windowTitle = "NovelRTTest", uint32_t targetFrameRate = 0);
+    explicit NovelRunner(int displayNumber, const std::string& windowTitle = "NovelRTTest", uint32_t targetFrameRate = 0, bool transparency = false);
```
capnkenny, line 46: As long as transparency is not specified, we assume that we don't want it enabled (which would mostly be the case).

src/NovelRT/Graphics/RenderingService.cpp
```diff
@@ -15,7 +15,8 @@ namespace NovelRT::Graphics {
       glBindBufferRange(GL_UNIFORM_BUFFER, 0, tempHandle, 0, sizeof(Maths::GeoMatrix4x4<float>));
       return tempHandle;
     })),
-    _camera(nullptr) {
+    _camera(nullptr),
+    _framebufferColour(RGBAConfig(0,0,0,0)) {
```
capnkenny, line 19: This is where our default colour is determined so that we always have _something_ passing into our `glClearColor` call.

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -36,6 +36,11 @@ namespace NovelRT::Windowing {
       checkForOptimus(OptimusLibraryName);
 #endif
 
+    //Set Framebuffer Transparency
+   if (transparencyEnabled)
+   {
+     glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_TRUE);
+   }
```
capnkenny, line 43: Sets the framebuffer to be transparent. Must be done on the Window level before the OpenGL context is initialized.

src/NovelRT/Graphics/RenderingService.cpp
RubyNova: Shouldn't the default colour be black (or the blue we originally used)?

src/NovelRT/Windowing/WindowingService.cpp
RubyNova: Does this account for all possible attempts to create the window?

Also your bracing is off.

include/NovelRT/Graphics/RenderingService.h
```diff
@@ -37,7 +39,7 @@ namespace NovelRT::Graphics {
   public:
     RenderingService(NovelRunner* const runner);
     int initialiseRendering();
-
+
```
RubyNova, line 42: What happened here exactly?

include/NovelRT/Graphics/RenderingService.h
```diff
@@ -53,6 +55,8 @@ namespace NovelRT::Graphics {
     void beginFrame() const;
     void endFrame() const;
 
+    void setBackgroundColour(const RGBAConfig colour);
```
RubyNova, line 58: Don't we pass `RGBAConfig` by reference? Not that it matters too much I don't think in this case, but its marked `const` when you're passing by value, which is odd.

include/NovelRT/NovelRunner.h
```diff
@@ -23,7 +23,7 @@ namespace NovelRT {
      * Use this to define game behaviour that is required to be done per-frame.
      */
     Utilities::Event<Timing::Timestamp> Update;
-
+
```
RubyNova, line 26: What happened here exactly?

src/NovelRT/Windowing/WindowingService.cpp
capnkenny: We have different types of bracing throughout our code - can you remind me which style we're to use going forward?
And yes, since this is called before OpenGL context initialization, it is cached for all attempts.

src/NovelRT/Graphics/RenderingService.cpp
```diff
@@ -279,4 +280,9 @@ namespace NovelRT::Graphics {
     returnValue->loadFontAsTextureSet(fileTarget, fontSize);
     return returnValue;
   }
+
+  void RenderingService::setBackgroundColour(const RGBAConfig colour)
+  {
+    _framebufferColour = colour;
+  }
```
RubyNova, line 287: Bracing is screwy here.

src/NovelRT/Graphics/RenderingService.cpp
capnkenny: The color specified here _is_ black (technically speaking it's no color, but it produces black since Red, Green, and Blue are not introduced).

If you'd like the default to be the original blue, I can change this to (0,0,102,0) (this would effectively reproduce our original call of `glColor4f(0.0f, 0.0f, 0.4f, 0.0f)`)

include/NovelRT/Graphics/RenderingService.h
```diff
@@ -23,6 +23,8 @@ namespace NovelRT::Graphics {
     ShaderProgram _texturedRectProgram;
     ShaderProgram _fontProgram;
 
+    RGBAConfig _framebufferColour;
+
```
RubyNova, line 27: This ordering of fields doesn't match the ordering initialised in the ctor I don't think? Could you check this please?

capnkenny:
Looks like I have some re-ordering to do to make Ubuntu happy 😅 I'll do this and make the styling corrections when I get back

src/NovelRT/Graphics/RenderingService.cpp
RubyNova: But, alpha is `0` here. How is this working?

src/NovelRT/Windowing/WindowingService.cpp
RubyNova: For the time being were still using K&R. We agreed to switch to Allman once we enforced the .NET style everywhere via `clang-format`, but that hasn't been done yet, so its best to just stick to the old style until we go over.

src/NovelRT/Graphics/RenderingService.cpp
capnkenny: Had to look this one up lol. It works because normal windows that are non-transparent do not carry the alpha value to the framebuffer, so the `0` for alpha is just ignored - we never had to worry before so 0 worked for us. However, when transparency is enabled, it will _not_ be ignored.

I'll set alpha to 1 by default.

include/NovelRT/NovelRunner.h
capnkenny: _Nothing to see here...just an extra space is all 👍 _

include/NovelRT/Graphics/RenderingService.h
capnkenny: Switched to pass by reference.

include/NovelRT/Graphics/RenderingService.h
capnkenny: Another extra space thats been removed now, tis all

src/NovelRT/Windowing/WindowingService.cpp
capnkenny: Fixed to K&R style. And to answer the previous question (as a reminder from Discord), this accounts for all window creation attempts :D

PR #205, opened 2020-07-24
System decoupling and refactor.

### LINKED ISSUE #204: Service decoupling and refactoring.
As it stands, the systems are all tightly coupled in a very specific manner, mainly due to the fact that we don't guarantee lifetime for any of the services amongst other things.

As such, a refactoring needs to take place to move away from the current setup so that decoupling can be done more smoothly. This will also massively aid in making `NovelRunner` more of an optional helper type.

RubyNova:
This PR refactors the system services so that the runner is more optional. However, this means we have moved away from "the runner controlling the lifetime" to a purely shared_ptr model. It will make the UX of the API a lot nicer in the long run despite being a breaking change (its alpha after all, this stuff happens). Resolves #204 .


capnkenny:
I want to do some brief testing first, however this looks pretty good imo.

2 questions:

1) Is there a chance of circular dependency now that we're purely on shared pointers? (I didn't see any explicitly in the PR but I'd like to know if you happen to have identified any)

2) For dtors, we'd want to move away from destructing explicitly and instead call a `tearDown()` method of some sort, correct?

RubyNova:
> Is there a chance of circular dependency now that we're purely on shared pointers? (I didn't see any explicitly in the PR but I'd like to know if you happen to have identified any)

No. All the systems were decoupled in terms of circular dependencies long ago. This change is purely a lifetime management change. 😄 

> For dtors, we'd want to move away from destructing explicitly and instead call a tearDown() method of some sort, correct?

IMHO, the best way is to have both i.e. have dtor call `tearDown()` in situations where its needed. This way if we don't call it ourselves and the service is discarded, it still cleans up after itself.

PR #207, opened 2020-07-25
Fix all broken logging statements.

jeuxjeux20:
Fixed a bug where it would write `Screen size: `  with no screen size after that. That was caused by `logError("something", argument)` with no `{}` placeholders in the string that prints `argument`. There were also others statements that have the same problem. And now it's fixed. Nice.


capnkenny:
Nice 😁 No complaints here, just gonna test this before merging

RubyNova:
I'd want @jeuxjeux20 to do it ideally, as I've no idea if this PR is even done yet. 😄

jeuxjeux20:
I'll try to merge as soon as the others PRs are merged, no worries!

src/NovelRT/Windowing/WindowingService.cpp
```diff
@@ -135,9 +135,9 @@ namespace NovelRT::Windowing {
   void WindowingService::checkForOptimus(const char* library) {
     _optimus = LoadLibrary(reinterpret_cast<LPCSTR>(library));
     if (_optimus != nullptr) {
-      _logger.logInfoLine("NVIDIA GPU detected. Enabling...");
+      _logger.logInfo("NVIDIA GPU detected. Enabling...");
     } else {
-      _logger.logInfoLine("NVIDIA GPU not detected. Continuing w/o Optimus support.");
+      _logger.logInfo("NVIDIA GPU not detected. Continuing w/o Optimus support.");
```
RubyNova, line 140: Since were here, you could change that `w/o` to `without`?

jeuxjeux20: Done!

PR #214, opened 2020-08-01
Build tools: Write infrastructure to copy outputs between dependencies

### LINKED ISSUE #208: [Windows] Nethost is either not copying correctly or not being loaded correctly under the new build system.
To clarify - it is copied into the `dotnet` subfolder, but on windows it expects `nethost.dll` to be in the same folder as the executable. I have no idea why.

### LINKED ISSUE #209: GTest and GMock isn't correctly identifying the build type on Windows
Workaround: specify `-Dgtest_USE_DEBUG_BUILD=ON -Dgmock_USE_DEBUG_BUILD=ON` as additional arguments when building Debug or RelWithDebInfo.

In summary: Fuck vcpkg.

FiniteReality:
AKA: the missing piece

This should replace the weird vcpkg dependency script that's built in, and fix #209. In addition, this should also fix #208 as that was part of the same issue.

Not tested on Windows yet. As such, I'm expecting CI to fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,11 @@ option(NOVELRT_BUILD_SAMPLES "Build NovelRT samples" ON)
 option(NOVELRT_BUILD_DOCUMENTATION "Build NovelRT documentation" ON)
 option(NOVELRT_BUILD_TESTS "Build NovelRT tests" ON)
 
+include(OutputCopyLocal)
+
 add_subdirectory(src)
+
+add_subdirectory(resources)
 
 if(NOVELRT_BUILD_SAMPLES)
   add_subdirectory(samples)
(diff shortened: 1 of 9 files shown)


FiniteReality:
I've let this go stale since libraries on windows are ridiculous and CMake doesn't do stuff right. I'm guessing the solution is going to end up being "write our own CMake Ffnd scripts to actually do things the correct way"...

Which is going to take absolutely forever so I'm going to leave our workarounds in #209 and #208 in place for now.

@RubyNova We should probably have a wiki page for a FAQ or some such. That way I can list these workarounds there.

RubyNova:
OK.

This sorta means getting our "website" (aka github pages) up and running correctly, so we'll need to do that first.

FiniteReality:
Closing as this was obsoleted by #385

Issue #103, opened 2020-01-03
Add factory method for default perspective view in NovelRT::Graphics::Camera.


RubyNova:
Title, mostly, but this is basically a ticket for supporting a perspective counterpart to the orthographic factory method effectively. Not urgent, but definitely on the TODO.

RubyNova:
Out of date version, and duplicate, of #91. Closing.

PR #216, opened 2020-08-05
Add the perspective camera factory method

### LINKED ISSUE #91: Introduce built-in camera modes for NovelRT::Graphics::Camera?
As per #73 I have introduced the groundwork for supporting multiple camera view/projection view types.

Its a bit messy and needs cleaning up, but once its done we can look into this in further detail for Novel3D probably, I can't see them getting too much use out of it in the current state of the engine.

We can either remove it or try to implement it however I'm a bit "on the fence" about it, and would like other people's opinions.

Shidesu:
Edit from Ruby: Resolves #91 .

### DIFF
--- a/include/NovelRT/Graphics/Camera.h
+++ b/include/NovelRT/Graphics/Camera.h
@@ -52,8 +52,14 @@ namespace NovelRT::Graphics {
       return _cameraFrameState;
     }
 
+    /**
+     * Creates an orthographic camera using default settings and the given window size.
+     */
     static std::unique_ptr<Camera> createDefaultOrthographicProjection(const Maths::GeoVector2<float>& windowSize);
 
+    /**
+     * Creates a perspective camera using default settings and the given window size.
+     */
     static std::unique_ptr<Camera> createDefaultPerspectiveProjection(const Maths::GeoVector2<float>& windowSize);
   };
 }
--- a/include/NovelRT/Graphics/Camera.h
+++ b/include/NovelRT/Graphics/Camera.h
@@ -54,7 +54,7 @@ namespace NovelRT::Graphics {
 
     static std::unique_ptr<Camera> createDefaultOrthographicProjection(const Maths::GeoVector2<float>& windowSize);
 
-    static std::unique_ptr<Camera> createPerspectiveProjection(const Maths::GeoVector2<float>& windowSize);
+    static std::unique_ptr<Camera> createDefaultPerspectiveProjection(const Maths::GeoVector2<float>& windowSize);
   };
 }
 
--- a/include/NovelRT/Graphics/Camera.h
+++ b/include/NovelRT/Graphics/Camera.h
@@ -17,6 +17,7 @@ namespace NovelRT::Graphics {
     Utilities::Lazy<Maths::GeoMatrix4x4<float>> _cameraUboMatrix;
     Maths::GeoMatrix4x4<float> generateUboMatrix();
     CameraFrameState _cameraFrameState;
+    std::function<void(Camera*, const Maths::GeoVector2<float>&)> _forceResizeCallback;
 
     void initialiseCameraForFrame();
     void forceResize(const Maths::GeoVector2<float>& windowSize);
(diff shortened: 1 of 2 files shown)


include/NovelRT/Graphics/Camera.h
```diff
@@ -53,6 +53,8 @@ namespace NovelRT::Graphics {
     }
 
     static std::unique_ptr<Camera> createDefaultOrthographicProjection(const Maths::GeoVector2<float>& windowSize);
+
+    static std::unique_ptr<Camera> createPerspectiveProjection(const Maths::GeoVector2<float>& windowSize);
```
RubyNova, line 57: can we rename this to `createDefaultPerspectiveProjection`?

RubyNova:
Resolves #103 .

src/NovelRT/Graphics/Camera.cpp
```diff
@@ -15,6 +15,13 @@ namespace NovelRT::Graphics {
     return returnVal;
   }
 
+  std::unique_ptr<Camera> Camera::createDefaultPerspectiveProjection(const Maths::GeoVector2<float>& windowSize) {
+    auto returnVal = std::make_unique<Camera>();
+    returnVal->setProjectionMatrix(Maths::GeoMatrix4x4<float>(glm::perspective<float>(glm::radians(90.0f), windowSize.getX() / windowSize.getY(), 0.1f, 100.0f)));
+    returnVal->setViewMatrix(Maths::GeoMatrix4x4<float>(glm::lookAt(glm::vec3(0.0f, 0.0f, 3.0f), glm::vec3(0.0f, 0.0f, 0.0f), glm::vec3(0.0f, 1.0f, 0.0f))));
```
Shidesu, line 21: Do you want me to comment the role of each vec3?

RubyNova: What do you mean?

Shidesu: In case what they are for is unclear?

RubyNova: ok maybe im being dumb - are you talking about commenting API surface or implementation details?

Shidesu: Implementation details, for other contributors, don't know if that's a thing

RubyNova: In this case, GLM is its own documented API. I know what that does, the developers know what that does, and I'd expect anyone wanting to touch the graphics internals to know what it does or to at least go read up about it.

_However._

It might not hurt to add any missing Doxygen documentation to the _public_ API surface while you are here to the camera, which describes what the API surface does. :D

Shidesu:
So a kind of strategy pattern right? And we set the delegate in the factory method.

RubyNova:
Yes, internally the factory methods will set the delegate (or `std::function<T>` in C++'s case, but you get the idea), then people can provide custom ones should they choose to make a custom camera without our factory stuff. :D

Shidesu:
Sounds good to me, I'll add that tomorrow I think.

src/NovelRT/Graphics/Camera.cpp
```diff
@@ -33,7 +57,6 @@ namespace NovelRT::Graphics {
   }
 
   void Camera::forceResize(const Maths::GeoVector2<float>& windowSize) {
-    setProjectionMatrix(Maths::GeoMatrix4x4<float>(glm::ortho<float>(0, windowSize.getX(), windowSize.getY(), 0, 0, 65535)));
-    setViewMatrix(Maths::GeoMatrix4x4<float>(glm::scale(glm::vec3(windowSize.getX() / 1920.0f, windowSize.getY() / 1080.0f, -1.0f))));
+    forceResizeCallback()(this, windowSize);
```
RubyNova, line 60: Have you confirmed this provides the correct behaviour at runtime?

Shidesu: I ran the sample and personally saw no difference. I ran it with perspective and orth camera.
I might be wrong so a double/triple/quadruple check is welcome.

RubyNova: All you need to check is resize the window at runtime. Did you do that and did it work?

Shidesu: Yeah the resize worked. Only time I had a failure it was on the persepctive camera when there was no check for Y == 0.
Now there's a check and stuff is correctly resizing.

RubyNova: Ok awesome.

PR #219, opened 2020-08-11
Improved Event and EventHandler

jeuxjeux20:
Introduced move semantics and a `remove` method to remove `EventHandler` by their ids.
The API is a bit dirty (i think) so this will be a draft for now.

* All `operator+=` of `Event` make use of perfect forwarding.
* `EventHandler` has also been modified to support perfect forwarding.
* A `remove(Atom)` method has been added to remove event handlers by their Atom, so you don't need to hold a reference to them.

### DIFF
--- a/include/NovelRT/Utilities/Event.h
+++ b/include/NovelRT/Utilities/Event.h
@@ -58,19 +58,19 @@ namespace NovelRT::Utilities {
       return _handlers.size();
     }
 
-    const EventHandler<TArgs...>& operator+=(const EventHandler<TArgs...>& handler) {
+    Atom operator+=(const EventHandler<TArgs...>& handler) {
       if (handler.getId() != Atom()) {
-        return _handlers.emplace_back(handler);
+        _handlers.emplace_back(handler);
       }
-      return handler;
+      return handler.getId();
     }
 
-    const EventHandler<TArgs...>& operator+=(const std::function<void(TArgs...)>& function) {
+    Atom operator+=(const std::function<void(TArgs...)>& function) {
       auto handler = EventHandler<TArgs...>(function);
       return *this += handler;
     }
 
-    const EventHandler<TArgs...>& operator+=(std::function<void(TArgs...)>&& function) {
+    Atom operator+=(std::function<void(TArgs...)>&& function) {
       auto handler = EventHandler<TArgs...>(function);
       return *this += handler;
     }
--- a/include/NovelRT/Utilities/Event.h
+++ b/include/NovelRT/Utilities/Event.h
@@ -18,9 +18,14 @@ namespace NovelRT::Utilities {
     EventHandler() : EventHandler(nullptr) {
     }
 
+    explicit EventHandler(std::function<void(TArgs...)>&& function) :
+            _id(Atom::getNextEventHandlerId()),
+            _function(std::move(function)) {
+    }
+
     explicit EventHandler(const std::function<void(TArgs...)>& function) :
-      _id((function != nullptr) ? Atom::getNextEventHandlerId() : Atom()),
-      _function(function) {
+            _id((function != nullptr) ? Atom::getNextEventHandlerId() : Atom()),
+            _function(function) {
     }
 
     void operator()(TArgs... args) const {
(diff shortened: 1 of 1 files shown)


include/NovelRT/Utilities/Event.h
```diff
@@ -18,9 +18,14 @@ namespace NovelRT::Utilities {
...
+    explicit EventHandler(std::function<void(TArgs...)>&& function) :
+            _id(Atom::getNextEventHandlerId()),
+            _function(std::move(function)) {
+    }
+
     explicit EventHandler(const std::function<void(TArgs...)>& function) :
-      _id((function != nullptr) ? Atom::getNextEventHandlerId() : Atom()),
-      _function(function) {
+            _id((function != nullptr) ? Atom::getNextEventHandlerId() : Atom()),
+            _function(function) {
```
RubyNova, line 28: random line ending changes or something else?

include/NovelRT/Utilities/Event.h
```diff
@@ -53,25 +58,37 @@ namespace NovelRT::Utilities {
       return _handlers.size();
     }
 
-    void operator+=(const EventHandler<TArgs...>& handler) {
+    Atom operator+=(const EventHandler<TArgs...>& handler) {
```
RubyNova, line 61: I'm a bit concerned this may alienate some people - traditionally += doesn't return anything I don't think? so I'm not sure if this would make for good API UX. I'll wait and see on any sample code that may end up in the proposal though.

include/NovelRT/Utilities/Event.h
```diff
@@ -53,25 +58,37 @@ namespace NovelRT::Utilities {
...
       if (handler.getId() != Atom()) {
         _handlers.emplace_back(handler);
       }
+      return handler.getId();
+    }
+
+    Atom operator+=(const std::function<void(TArgs...)>& function) {
```
RubyNova, line 68: Same here, += doesn't normally return anything.

include/NovelRT/Utilities/Event.h
```diff
@@ -53,25 +58,37 @@ namespace NovelRT::Utilities {
...
     void operator-=(const EventHandler<TArgs...>& handler) {
-      if (handler.getId() == Atom())
-        return;
+      remove(handler.getId());
+    }
 
-      auto match = std::find(_handlers.cbegin(), _handlers.cend(), handler);
+    void remove(Atom eventHandlerId) {
+      if (eventHandlerId == Atom())
+        return;
```
RubyNova, line 84: I'm not entirely sure I understand this?

RubyNova:
Overall, I'm just wondering if there's a more explicit way we could deal with this. What is the current reasoning behind not using the event handler object itself?

include/NovelRT/Utilities/Event.h
jeuxjeux20: this was an adaptation of the same code before, so event handlers that have the default Atom (because they are empty) don't do anything

include/NovelRT/Utilities/Event.h
jeuxjeux20: CLion did some sneaky stuff I guess

include/NovelRT/Utilities/Event.h
jeuxjeux20: I agree that's pretty weird, I'll put it back to void, however, there will still be no way of removing handlers created using the `operator+=` with `std::function`, which is something I would like to "fix", or we can maybe leave it as it is by design.

jeuxjeux20:
Not using the `EventHandler` object directly makes a bit more sense to me, as you can add a lambda and have the `Event` *hold* the `EventHandler`, which may contain resources in its callback function.

I made new changes, the `operator+=` returns `void` and is completely perfect-forwarded.
However, the `std::function`'s constructor that takes a callable object requires to be **`CopyConstructible`**, which makes move-only objects (example: `std::unique_ptr`) impossible to use.

Considering that sadly this doesn't fix what I wanted to, which is, have event handlers hold move-only objects, I don't really know if this PR is even worth continuing.

jeuxjeux20:
actually no screw that

PR #221, opened 2020-08-14
Doxy

QuantumlyTangled:


FiniteReality:
I believe that the better fix for this would be to update the CI images to use Doxygen 1.8.19, where this Doxygen bug seems to be fixed.

We could also mention this in the "FAQ" document I mentioned in #222 until Doxygen 1.8.19 lands in more locations

FiniteReality:
Update:

Doxygen 1.8.18 appears to generate the correct documentation with the following changes necessary:

```diff
diff --git a/doxygen/CMakeLists.txt b/doxygen/CMakeLists.txt
index edce0af..20c3150 100644
--- a/doxygen/CMakeLists.txt
+++ b/doxygen/CMakeLists.txt
@@ -19,8 +19,12 @@ set(DOXYGEN_INTERACTIVE_SVG YES)
 
 set(DOXYGEN_USE_MDFILE_AS_MAINPAGE README.md)
 
+set(DOXYGEN_STRIP_FROM_PATH ${CMAKE_SOURCE_DIR})
+
 doxygen_add_docs(Doxygen ALL
   COMMENT "Generating documentation"
   ${CMAKE_SOURCE_DIR}/README.md
+  ${CMAKE_SOURCE_DIR}/LICENCE.md
+  ${CMAKE_SOURCE_DIR}/CODE_OF_CONDUCT.md
   ${CMAKE_SOURCE_DIR}/include
 )
```

RubyNova:
Then can you post that summary here? Discord is a great way to lose things

QuantumlyTangled:
Comment pulled from Discord (15/08/2020)
```
In short latest has relative path support, but many older versions dont.
And the pre shipped one on Ubuntu doesnt
So either we can ignore the Ubuntu LTS users and go for latest.
Or provide backwards support
```

FiniteReality:
We're already ignoring LTS users older than 20.04 since the CI environment only targets 20.04 - for end users there's no real reason to be on anything older than this. However, the diff I mentioned earlier does not work in 1.8.17, which is what 20.04 uses.

Because of this, I'm going to tentatively approve this as a workaround, with a heavy note that as soon as we can, we should revert back to the patch I mentioned in https://github.com/novelrt/NovelRT/pull/221#issuecomment-674341322.

FiniteReality (approved):

PR #227, opened 2020-11-11
Adding basic ink support

tannergooding:
This adds basic ink support to NovelRT.


include/NovelRT/DotNet/RuntimeService.h
```diff
@@ -8,41 +8,36 @@
...
-  class RuntimeService {
+  class RuntimeService : public std::enable_shared_from_this<RuntimeService> {
   private:
+    struct Exports {
+      void(*Initialise)();
+      void(*Teardown)();
+      void(*FreeObject)(intptr_t obj);
+      void(*FreeString)(const char* str);
+      void(*GetInkServiceExports)(struct Ink::InkService::Exports* exports);
+    };
```
RubyNova, line 19: Minor nit: Could we not use `std::add_pointer<T>::type` here to make these fnptrs easier to read?

include/NovelRT/DotNet/RuntimeService.h
```diff
@@ -8,41 +8,36 @@
...
-
-      if (result != 0)
-      {
-        _logger.logError("Failed to locate the specified managed function: {}", result);
-        throw std::runtime_error("Failed to locate the specified managed function");
-      }
-
-      return std::function<_Fty>(reinterpret_cast<_Fty*>(delegate));
-    }
-
```
RubyNova, line 43: What happened here? I thought the ink stuff was just gonna sit on top of the runtime servie in our plan to standardise the runtime service as an interface that other runtimes such as JVM can have bindings written for? Was this method just redesigned?

include/NovelRT/Ink/InkService.h
```diff
@@ -0,0 +1,36 @@
...
+#endif
+
+namespace NovelRT::Ink {
+  class InkService : public std::enable_shared_from_this<InkService> {
+  private:
+    struct Exports {
+      void(*Initialise)();
+      void(*Teardown)();
+      void(*GetStoryExports)(struct Story::Exports* exports);
+    };
```
RubyNova, line 17: Minor nit, same as previous: Could we not use std::add_pointer<T>::type here to make these fnptrs easier to read?

include/NovelRT/Ink/Story.h
```diff
@@ -0,0 +1,42 @@
...
+  class Story {
+  private:
+    struct Exports {
+      intptr_t(*CreateFromJsonString)(const char* jsonString);
+      void(*ChooseChoiceIndex)(intptr_t storyHandle, int choiceIdx);
+      const char*(*Continue)(intptr_t storyHandle);
+      const char*(*ContinueMaximally)(intptr_t storyHandle);
+      bool(*GetCanContinue)(intptr_t storyHandle);
+      void(*ResetState)(intptr_t storyHandle);
+    };
```
RubyNova, line 20: Minor nit, same as previous: Could we not use std::add_pointer<T>::type here to make these fnptrs easier to read?

include/NovelRT/Ink/Story.h
```diff
@@ -0,0 +1,42 @@
...
+    Story(std::shared_ptr<InkService> inkService, void(*getExports)(Exports* exports), const char* jsonString) noexcept;
+
+    bool CanContinue();
+
+    void ChooseChoiceIndex(int choiceIdx);
+
+    const char* Continue();
+    const char* ContinueMaximally();
+
+    void ResetState();
```
RubyNova, line 36: Are the pascal case functions intentional as per our new planned style?

include/NovelRT/Ink/Story.h
```diff
@@ -0,0 +1,42 @@
...
+
+    friend class InkService;
+  };
+}
+
+#endif // NOVELRT_TIMING_STEPTIMER_H
```
RubyNova, line 42: This comment seems to have the wrong name.

resources/Scripts/story.json
```diff
@@ -0,0 +1,18 @@
...
+    "\n",
+    "^And then it should loop back to the beginning.",
+    "\n",
+    "done",
+    {
+      "#f": 3
+    }
+  ],
+  "listDefs": {}
+}
```
RubyNova, line 18: Any chance we could push up the ink source for this?

RubyNova (commented):
I have a few questions before approving - overall looking good though. 👍

include/NovelRT/DotNet/RuntimeService.h
tannergooding: Possibly, but I think it would be better to just keep them like this as we will need these signatures anyways for use in the C wrapper (if we made it not a nested type, its definition could even be shared).

include/NovelRT/DotNet/RuntimeService.h
tannergooding: This was exposing the 3.1 `getFunction` which required matching delegate types. The new functionality exposes the 5.0 `getFunction` which just returns a straight function pointer to a `UnmanagedCallersOnly` static method.

include/NovelRT/Ink/Story.h
tannergooding: Yes and because we can't name it `continue` since that is a C/C++ keyword.

include/NovelRT/Ink/Story.h
tannergooding: Yeah, copy/paste error.
```suggestion
#endif // NOVELRT_INK_STORY_H
```

resources/Scripts/story.json
tannergooding: You mean the raw ink file? I don't see a "lot" of benefit given that you can import/export the json into inkle.

include/NovelRT/DotNet/RuntimeService.h
RubyNova: Ahhh so it's done on the C# side effectively?

tannergooding: Not quite. Its that function pointers didn't exist in .NET Core 3.1 but they do in .NET 5 so the API we need in NovelRT changed.
This reflects that by using the newer/better/cheaper function pointer API instead and removing the older one that required a delegate and extra allocation.

RubyNova: I see. Thanks!

RubyNova (approved):

PR #233, opened 2020-11-28
Update Readme for custom vcpkg install

### LINKED ISSUE #232: Readme does not specify to use the vcpkg fork
When following the readme you will be directed to the instalation guide of vcpkg, however by following that users will end up installing from the microsoft repository, which novelrt uses it's own branch.

capnkenny:
Changed to custom version of vcpkg for now.

Fixes #232

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,13 +69,16 @@ make -j
 ```
 
 #### Windows
-First, you must set up (our version of*) vcpkg and Python 3 to install the dependencies:
+First, you must set up [(our fork of*) vcpkg](https://github.com/capnkenny/vcpkg) as shown below:
 ```
-python -m pip install setuptools
-python -m pip install glad
 git clone https://github.com/capnkenny/vcpkg
 cd vcpkg
 ./bootstrap-vcpkg.bat
+```
+Next, use vcpkg and Python 3 to install the dependencies:
+```
+python -m pip install setuptools
+python -m pip install glad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,11 +69,15 @@ make -j
 ```
 
 #### Windows
-First, you must [setup vcpkg](https://docs.microsoft.com/en-us/cpp/build/vcpkg?view=vs-2019#installation) and Python 3 to install the dependencies:
+First, you must set up (our version of*) vcpkg and Python 3 to install the dependencies:
```
RubyNova, line 72: We should probably clarify here the instructions remain the same, just with your fork. Can you also link it?

RubyNova (changes requested):

README.md
capnkenny: Wouldn't the additional lines afterwards cope for this, since the instructions now state the address to clone from?

My fear is that just stating the directions are the same from the original site will cause them to pull from the MSFT branch and not ours.

RubyNova: Ah I did not see those - However I am still concerned about the "completeness" here. Any way we could improve it?

capnkenny: I've split the directions up and added a specific link to our fork - pls re-review and let me know if this was more what you were expecting?

RubyNova (approved):
I am still concerned someone will ask us what to do with the custom vcpkg but this will do for now as I've just seen another PR affecting things.

Issue #35, opened 2019-09-12
Implement clang-format stylesheet.


RubyNova:
We need to get up and running on `clang-format` based on our current code style. Some stuff keeps passing in PRs when it shouldn't and really it'd be great to automate this with `clang-format`.

jatin69:
Hey @RubyNova 
I use [this clang-format file](https://github.com/jatin69/FTP/blob/master/.clang-format) in my c++ projects. Kindly review this file and associated codebase formatting, and let me know if this one works for you. I'll send a PR to add the file.

RubyNova:
Thanks, I'll check it out.

capnkenny:
Picking this up next as formatting is currently hellish 🗡️ 

Based on Microsoft (.NET) formatting with a few tweaks as we go
-Option to run formatter via CMake
-Option to check formatting via CI (which will fail the build if anything is found)

PR #236, opened 2020-12-02
Implementing Clang Format

capnkenny:


capnkenny:
Hey @FiniteReality, since I'm going to be adding the clang-format checks into CI, do you think it needs to be added to all of the runs? Currently I've only implemented it for Windows, and was curious to know your thoughts as the Windows runs should produce a diff file that's applicable in either OS (I'd think...)

FiniteReality:
I'd personally make a separate job for it, rather than patching it into an existing job.

cmake/clang-format.cmake
```diff
@@ -0,0 +1,45 @@
...
+     ALL_SRC_FILES
+      src/*/*.c src/*/*.cpp samples/*/*.cpp samples/*/*.c include/*/*.h tests/*/*.cpp
+     )
+
+
+find_program(CLANG_FORMAT "clang-format")
+if(CLANG_FORMAT)
+  message("Found clang-format at ${CLANG_FORMAT}")
+  if(WIN32)
+    add_custom_target(
```
FiniteReality, line 11: clang-format should probably be done as a step in a target, e.g.:
```cmake
add_custom_command(TARGET <target> PRE_BUILD
    command clang::format $<TARGET_PROPERTY:SOURCES>
)
```

cmake/clang-format.cmake
```diff
@@ -0,0 +1,45 @@
+file(GLOB_RECURSE
+     ALL_SRC_FILES
+      src/*/*.c src/*/*.cpp samples/*/*.cpp samples/*/*.c include/*/*.h tests/*/*.cpp
+     )
+
+
+find_program(CLANG_FORMAT "clang-format")
```
FiniteReality, line 7: This should probably be done as part of a `find_package` script.

CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ project(NovelRT
   LANGUAGES C CXX
 )
 
+# Added Clang-format custom target so that we can run it locally if needed.
+include(cmake/clang-format.cmake)
```
FiniteReality, line 11: Should probably be an option for this?

capnkenny:
ahhhh its broken... I'll fix it later - gonna switch to Ubuntu-based clang-format checking

cmake/clang-format.cmake
```diff
@@ -0,0 +1,45 @@
...
+  if(WIN32)
+    add_custom_target(
+        clang-format
+        COMMAND clang-format
+        -i
+        -style=file
+        ${ALL_SRC_FILES}
+    )
+    add_custom_target(
+        clang-format-ci
```
FiniteReality, line 19: I don't think we should include CI-specific things in our CMake build script; it should be done separately.

scripts/ci-check-format.ps1
```diff
@@ -0,0 +1,9 @@
+Param(
+  [ValidateSet("Debug", "MinSizeRel", "Release", "RelWithDebInfo")][string] $configuration = "Debug"
+)
+
+$RepoRoot = Join-Path -Path $PSScriptRoot -ChildPath ".."
+$ArtifactsDir = Join-Path -Path $RepoRoot -ChildPath "artifacts"
+$BuildDir = Join-Path -Path $ArtifactsDir -ChildPath "build/$configuration"
+
+cmake --build $BuildDir --target clang-format-ci
```
FiniteReality, line 9: We probably shouldn't even need cmake here; a glob should be fine.

FiniteReality (commented):
Style changes seem alright, but I'm not sure the CMake changes are ideal...

My thoughts are that checking the format of the code should likely stay out of the build system; we should simply have a CI check for it and encourage users to use the appropriate plugins/support in their development environments.

FiniteReality:
Possibly could use [this](https://github.com/marketplace/actions/clang-format-lint) instead of a bunch of CMake stuff?

capnkenny:
Thanks for the feedback @FiniteReality!

Based on that, I made the adjustments necessary to only have CI check the formatting via clang-format, and to output a patch if there's changes that are required.

This PR does _not_ cover how to apply said patch or how to run locally as there are numerous ways - this can probably be added to our documentation at a later date.

The style guide used is based on Microsoft's style guide (as per @RubyNova) with a few modifications. As such, I've applied it against all headers and source files (*.c[pp]) - note that C# isn't formatted at this time (afaik).

capnkenny:
Resolves #35

include/NovelRT.Interop/Animation/NrtSpriteAnimatorFrame.h
```diff
@@ -1,27 +1,28 @@
...
-NrtTimestamp Nrt_SpriteAnimatorFrame_getDuration(NrtSpriteAnimatorFrame frame);
-NrtResult Nrt_SpriteAnimatorFrame_setDuration(NrtSpriteAnimatorFrame frame, NrtTimestamp timestamp);
+    NrtResult Nrt_SpriteAnimatorFrame_getTexture(NrtSpriteAnimatorFrame frame, NrtTexture* outputTexture);
+    NrtResult Nrt_SpriteAnimatorFrame_setTexture(NrtSpriteAnimatorFrame frame, NrtTexture texture);
+    NrtTimestamp Nrt_SpriteAnimatorFrame_getDuration(NrtSpriteAnimatorFrame frame);
+    NrtResult Nrt_SpriteAnimatorFrame_setDuration(NrtSpriteAnimatorFrame frame, NrtTimestamp timestamp);
 
-NrtResult Nrt_SpriteAnimatorFrame_addFrameEnter(NrtSpriteAnimatorFrame frame, void(*func)());
-NrtResult Nrt_SpriteAnimatorFrame_addFrameExit(NrtSpriteAnimatorFrame frame, void(*func)());
+    NrtResult Nrt_SpriteAnimatorFrame_addFrameEnter(NrtSpriteAnimatorFrame frame, void (*func)());
```
RubyNova, line 24: Do we want to keep these sorts of spaces in fnptrs?

RubyNova:
Oh god what no I only put the last commit rEeEeEeEeE

include/NovelRT.Interop/Animation/NrtSpriteAnimatorFrame.h
capnkenny: Based on my testing, `PointerAlignment` is already set to `Left` and clang-format is not picking this up as a C-Style cast (since it really isn't...) so this will be unavoidable unfortunately.

At the moment, I'm using clang-format from Clang 10 since that is what our CI is targeting iirc. If that changes, I can certainly check the newer style guides as it updates to see if they have anything to address this in the future.

capnkenny:
Granted that CI passes, I've alphabetized the `.clang-format` options a bit so it aligns with the style guide a bit more. That, and reformatting what's in master for the billionth time   : ^ )

include/NovelRT.Interop/Animation/NrtSpriteAnimatorState.h
```diff
@@ -1,29 +1,36 @@
...
-NrtResult Nrt_SpriteAnimatorState_setShouldLoop(NrtSpriteAnimatorState state, int32_t loop);
-NrtResult Nrt_SpriteAnimatorState_getFrames(NrtSpriteAnimatorState state, NrtSpriteAnimatorFrameVector* outputFramess);
-NrtResult Nrt_SpriteAnimatorState_setFrames(NrtSpriteAnimatorState state, NrtSpriteAnimatorFrameVector frames);
-NrtResult Nrt_SpriteAnimatorState_tryFindValidTransition(NrtSpriteAnimatorState state, NrtSpriteAnimatorState* outputTransitionState);
+    NrtSpriteAnimatorState Nrt_SpriteAnimatorState_create();
+    NrtResult Nrt_SpriteAnimatorState_insertNewState(NrtSpriteAnimatorState state, NrtSpriteAnimatorState stateTarget,
+                                                     NrtSpriteAnimatorStateConditionFunctions vector);
+    NrtResult Nrt_SpriteAnimatorState_removeStateAtIndex(NrtSpriteAnimatorState state, size_t index);
+    int32_t Nrt_SpriteAnimatorState_getShouldLoop(NrtSpriteAnimatorState state);
+    NrtResult Nrt_SpriteAnimatorState_setShouldLoop(NrtSpriteAnimatorState state, int32_t loop);
```
RubyNova, line 21: TODO for myself: update this to NrtBool

include/NovelRT.Interop/Graphics/NrtRenderingService.h
```diff
@@ -1,35 +1,52 @@
...
+    NrtResult Nrt_RenderingService_create(NrtWindowingService windowingService,
+                                          NrtRenderingService* outputRenderingService);
+    NrtResult Nrt_RenderingService_initialiseRendering(
+        NrtRenderingService renderingService); // this function has its own return codes, so in theory we can combine
+                                               // them both into the main return code
+    NrtResult Nrt_RenderingService_tearDown(NrtRenderingService renderingService);
+    NrtResult Nrt_RenderingService_createImageRectWithFile(NrtRenderingService renderingService,
+                                                           NrtImageRect* outputImageRect, NrtTransform transform,
+                                                           int32_t layer, const char* filePath,
+                                                           NrtRGBAConfig colourTint);
```
RubyNova, line 26: Is this how we want to handle method parameters? It reads a bit odd to me but would appreciate further input.

capnkenny: I can attempt to binpack the params or allow them to each be on their own line if it won't fit on one, however this will probably cause some of our more param heavy methods to expand a bit space-wise.

@FiniteReality / @tannergooding any thoughts?

include/NovelRT/Ecs/ComponentCache.h
```diff
@@ -1,92 +1,95 @@
...
-     * @brief Stores all ComponentBuffer instances currently initialised for this instance of the ECS. You should not be instantiating this yourself in a regular setup.
-     * 
+     * @brief Stores all ComponentBuffer instances currently initialised for this instance of the ECS. You should not be
+     * instantiating this yourself in a regular setup.
+     *
      */
     class ComponentCache
     {
-        private:
+      private:
```
RubyNova, line 24: now THIS is interesting - I would like additional input on this indentation specifically. Do we want to keep this behaviour?

include/NovelRT.Interop/Graphics/NrtRenderingService.h
tannergooding: I think I'd prefer not to hav emultiple on the same line when they are split like this, it makes things like `transform` and `filePath` easy to miss

include/NovelRT/Ecs/ComponentCache.h
tannergooding: I'm fine with it being dedented one, to be even with the brace for the class

include/NovelRT.Interop/Animation/NrtSpriteAnimatorFrame.h
tannergooding: I think we should keep it and we should prefer typedefs rather than inline fnptr declarations.

FiniteReality: Function pointers are pretty hard to understand as is, so I totally approve of extracting them into typedefs.

include/NovelRT.Interop/Graphics/NrtRenderingService.h
FiniteReality: For C/C++, I personally prefer this:
```
void MyLongMethod(
    LongParameter1 myLongParameter,
    LongParameter2 myOtherLongParameter,
    LongParameter3 andSoOn);

File: include/NovelRT.Interop/Animation/NrtSpriteAnimatorFrame.h (hunk 6 above)
Comment (tannergooding) [reply include/NovelRT.Interop/Animation/NrtSpriteAnimatorFrame.h:24]: In before:
```cpp
template<typename... TArgs>
typedef void (*Action)(TArgs... args);

template<typename... TArgs, typename TResult>
typedef TResult (*Func)(TArgs... args);
```

FiniteReality: At that point, just use std::function :stuck_out_tongue:

include/NovelRT/Ecs/ComponentCache.h
capnkenny: Resolved - I've dedented access modifiers to be even with the bracing for the class.

include/NovelRT.Interop/Graphics/NrtRenderingService.h
capnkenny: Should be resolved - I've split all parameters onto separate lines.

include/NovelRT.Interop/Animation/NrtSpriteAnimatorFrame.h
capnkenny: TODO for me: Update this to typedefs (outside of this PR)

include/NovelRT.Interop/Animation/NrtAnimationTypedefs.h
```diff
@@ -1,19 +1,21 @@
-// Copyright © Matt Jones and Contributors. Licensed under the MIT License (MIT). See LICENCE.md in the repository root for more information.
+// Copyright © Matt Jones and Contributors. Licensed under the MIT License (MIT). See LICENCE.md in the repository root
+// for more information.
```
RubyNova, line 2: Is this intentional?

capnkenny: Yes - MS-style clang-format sets a column limit of 120, which works rather well with our parameter changes (it helps keep each param on one line when breaking them apart like we have).

On that note, it also appears to not effect Doxy in a negative way so we should be alright with this.

RubyNova (approved):
LGTM! (famous last words)

Issue #218, opened 2020-08-09
CI: investigate moving to Github Actions.


Tylertron1998:
Github Actions is a CI/CD pipeline that is scriptable with an API, and has a neat feature called [problem matchers](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md) - which is super great for developer experience. We could also consider using it to push release builds of the engine, ready to be used by consumers.

capnkenny:
I am aware of GH Actions being able to support our pipeline, so this wouldn't be so bad - especially since it would no longer require AzDO access so we could keep everything here. With that said, a few concerns:
1) (Not that AzDO has this too but) Is there support on Actions to have a UK English linter? (This was an ask by @RubyNova since we're working on a UK codebase.)

2) How do we proceed to log historical test results? As of now, it looks like we are able to log the test results from our unit tests into AzDO Test Plans, however I don't believe GH has an alternative to this.

Tylertron1998:
Actions is completely scriptable - so it's more a question of 'does a UK English linter exist?'; if so definitely.

Actions do save output from any workflow based on commit hash - however, they [have a time limit and do expire after 90 days.](https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#downloading-and-deleting-artifacts-after-a-workflow-run-is-complete). 

IMO this isn't a massive issue, I am not sure why we'd need to look into test logs from 90+ days ago -- and the code is still there so we can rerun tests manually if that's the case.

RubyNova:
We found a very...erm....**patriotic** linter shall we say? for US english, so we might just be able to fork that and rework it for UK english.

Tylertron1998:
awesome! If you can drop it to me either here, or in discord, I'll be happy to pull up a fork of novelrt and give it a test run!

Tylertron1998:
I've found this: https://github.com/nuspell/nuspell which seems to be our best option. It supports Hunspell libraries (yay for UK English!) but doesn't differentiate between source and non source files. I'll play about and see if I can make it only operate over files we care about (i.e. source & cmake files), ignore any language keywords, and go from there.

capnkenny:
> IMO this isn't a massive issue, I am not sure why we'd need to look into test logs from 90+ days ago -- and the code is still there so we can rerun tests manually if that's the case.

[...]

PR #237, opened 2020-12-02
ci: add github workflow config

Favna:
I guess my name is now Percy the Lightning Thief because @Tylertron1998 had his ⚡ stolen

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,45 @@
+name: Continuous Integration
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  CI_Ubuntu:
+    name: ubuntu_${{ matrix.configuration }}_x64
+    runs-on: ubuntu-latest
+    container:
+      image: novelrt/novelrt-build:latest
+    strategy:
+      matrix:
+        configuration: [DEBUG, RELEASE]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Ubuntu - Run CI Build Script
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,45 @@
+name: Continuous Integration
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
```
capnkenny, line 8: So at the moment, we have an AzDO job that runs every time a push to a PR is made, and then once every 24 hours on master...

Based on that, this looks to fulfill the PR change, however I'm not sure that this would cover the 24-hour requirement. Also, it appears that it will re-run when pushed to master.

Can you confirm if this is the case? And if so, would we be able to change this to only PRs and a 24-hour schedule?
(If more information is wanted on schedules for workflows, you can look [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events) 😄 )

capnkenny (commented):
Looks good overall to me - nice work! I just have one question which wasn't really defined by our requirements very well 😅 Sorry about that.

capnkenny:
Solves #218

.github/workflows/continuous-integration.yml
Favna: Done in 24ebb16

capnkenny (approved):
LGTM! Once AzDO passes ( >.>) we'll be good to turn off AzDO pipelines I think @RubyNova

RubyNova:
Awesome stuff. 😄

Tylertron1998 (approved):
👍 thank you again for this man! Appreciate the help

PR #243, opened 2020-12-15
Add a warning incase GCC is used.

Tylertron1998:
Ubuntu seems to default to GCC, so we should warn the user in case this happens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ If you wish to attempt to build a basic visual novel with the existing C++ API, 
 - OpenAL 1.19.1
 - spdlog 1.4.2
 
+> if you are compiling on Linux, please note - we do not support GCC at this time.
+
 ### Build instructions
 
 We provide build scripts in the root of the repository which may work for you.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,7 +64,7 @@ Then, clone and build NovelRT:
 git clone https://github.com/NovelRT/NovelRT.git
 mkdir -p NovelRT/build
 cd NovelRT/build
-cmake ..
```
RubyNova, line 67: This is not what I meant.

This is also technically wrong.

Just note we don't support GCC at this time in plain english somewhere appropriate?

RubyNova (changes requested):

README.md
Tylertron1998: Can you clarify what you mean by this? It ran perfectly fine on my machine - also yeah, adding some explanation is probably a good idea.

RubyNova: the official CXX compiler flag value is `clang++` and this is somewhat a hack on Linux. You'd usually change those env vars fully if you want to use clang.

README.md
```diff
@@ -60,6 +60,7 @@ sudo apt install clang cmake doxygen graphviz g++ libfreetype-dev libglfw3-dev \
     libspdlog-dev python3-glad
 ```
 Then, clone and build NovelRT:
+> **WARNING:** we currently do not support GCC as a compiler option. Expect errors/bugs if you try to use it - consider clang instead.
```
RubyNova, line 63: maybe instead of "consider" we could say "supported compilers are MSVC and clang for their respective platforms."?

Tylertron1998:
after some internal discussion on the discord server - we have decided instead to just note what the supported route should be. While giving exact instructions would be nice, that can be difficult given the vast variety of distros/operating systems/setups.

RubyNova (approved):
LGTM

PR #257, opened 2021-02-02
Adding Doxygen CI workflow to update documentation site

capnkenny:
This should allow us to automatically update the Doxygen documentation based on whatever is currently live in the master branch.

- Only will update as a PR is merged into master
- Will only push the updated documentation into the `gh-pages` branch as `Github Actions`


.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,46 @@
+name: Doxygen
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  CI_Doxygen:
```
FiniteReality, line 9: iirc jobs have a name too, so you could just name this `Build and publish documentation`

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: ubuntu_debug_x64
+    runs-on: ubuntu-latest
+    container:
+      image: novelrt/novelrt-build:latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Run CI Build Script (Engine & Docs only)
+        run: scripts/cibuild.sh --configuration Debug --generate -DNOVELRT_BUILD_TESTS OFF -DNOVELRT_BUILD_SAMPLES OFF
```
FiniteReality, line 19: IMHO we should make a dedicated parameter for cibuild.sh/build.sh which *only* builds the documentation.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Generate Documentation
```
FiniteReality, line 21: If we do the above, we shouldn't need this step, which makes debugging CI faster and easier

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,46 @@
...
+        uses: mattnotmitt/doxygen-action@v1
+        id: generate
+        with:
+            working-directory: '$GITHUB_WORKSPACE/novelrt/artifacts/build/Debug'
+            doxyfile-path: 'doxygen/Doxyfile.doxygen'
+
+      - name: Copy Documentation
+        id: copy
+        if: ${{ success() }}
+        run: mkdir -p $GITHUB_WORKSPACE/novelrt/docs && cp -r -f -v $GITHUB_WORKSPACE/novelrt/artifacts/build/Debug/doxygen/html/* $GITHUB_WORKSPACE/novelrt/docs
```
FiniteReality, line 31: I believe we can set `DOXYGEN_OUTPUT_DIRECTORY` as a CMake variable to `artifacts/docs` or some such (as part of cibuild, most likely) to avoid this step entirely.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Get current date
+        id: date
+        run: echo "name=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
```
FiniteReality, line 35: We probably don't actually need the date here since the `git commit` step includes the date in the git log anyway. What would be more useful, IMO, is the commit hash that triggered the build, which you can get using `$GITHUB_SHA`

FiniteReality (changes requested):
Looks good on paper, but I think we can do better for debugging purposes.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: novelrt/novelrt-build:latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Generate Documentation
+        run: scripts/cibuild.sh --documentation -DDOXYGEN_OUTPUT_DIR=$GITHUB_WORKSPACE/novelrt/docs
+
```
capnkenny, line 20: I added a flag to only generate documentation... this should work with and without `cibuild.sh` as the change is rooted in `build.sh`.

Should we add this for the Windows variant as well?

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Commit Documentation Updates
+        if: ${{ success() }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git checkout gh-pages
+          git add ./docs/*
+          git commit -m "Updated documentation - $GITHUB_SHA"
```
capnkenny, line 28: Updated to use the SHA of the current commit on master (that triggers this workflow) if I've understood the docs correctly.

doxygen/CMakeLists.txt
```diff
@@ -10,7 +10,14 @@ function(append_string_to_first_list_element APPEND_ARGS_INPUT APPEND_ARGS_APPEN
...
 
-set(DOXYGEN_OUTPUT_DIR ${CMAKE_BINARY_DIR}/doxygen)
+if (NOT DEFINED DOXYGEN_OUTPUT_DIR)
+  set(DOXYGEN_OUTPUT_DIR ${CMAKE_BINARY_DIR}/doxygen)
+else()
+  message("Doxygen output directory defined as: ${DOXYGEN_OUTPUT_DIR}")
+  set(DOXYGEN_OUTPUT_DIRECTORY ${DOXYGEN_OUTPUT_DIR})
+  set(DOXYGEN_HTML_OUTPUT ${DOXYGEN_OUTPUT_DIR})
+endif()
+
```
capnkenny, line 20: Funnily enough, we use `DOXYGEN_OUTPUT_DIR` as our own variable, but we never used to set `DOXYGEN_OUTPUT_DIRECTORY` which is what CMake uses to define where everything should go. Was this intended?

(Logic behind this is that this now allows controlling where the files go - for us, I can specify the `/docs` folder so once pushed, the gh-pages branch should be happy with the docs folder)

scripts/build.sh
```diff
@@ -189,6 +212,15 @@ if $generate; then
...
 
+if $documentation; then
+  Generate
+  GenerateDocumentationOnly
+
+  if [ "$LASTEXITCODE" != 0 ]; then
+    return "$LASTEXITCODE"
+  fi
+fi
+
```
capnkenny, line 223: If the documentation flag is passed, it assumes that Generate was never called to generate the Doxyfile so we call it manually here.

.github/workflows/doxygen.yml
capnkenny: This _should_ be resolved.

.github/workflows/doxygen.yml
capnkenny: Yup! I wrote another comment wondering about that CMake var. Resolved.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,30 @@
+name: Doxygen
+
+on:
+  push:
+    branches:
+      - master
```
FiniteReality, line 6: Should be `main`

doxygen/CMakeLists.txt
FiniteReality: It probably should have been `DOXYGEN_OUTPUT_DIRECTORY` in the first place - I'm personally fine with doing a rename instead of anything strange like this

FiniteReality (changes requested):

.github/workflows/doxygen.yml
tannergooding: This should hopefully get resolved simply with rebasing or merging the PR with latest

tannergooding: oh wait, nevermind, new file 😄

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Commit Documentation Updates
+        if: ${{ success() }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git checkout gh-pages
+          git add ./docs/*
+          git commit -m "Updated documentation - $GITHUB_SHA"
+          git push
```
tannergooding, line 29: I feel like this should be creating a PR to the pages repo and applying auto-merge once CI/sign-off completes

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/checkout@v2
+
+      - name: Generate Documentation
+        run: scripts/cibuild.sh --documentation -DDOXYGEN_OUTPUT_DIR=$GITHUB_WORKSPACE/novelrt/docs
+
+      - name: Commit Documentation Updates
+        if: ${{ success() }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
```
tannergooding, line 25: Is there docs on this e-mail or is there some "official-ness" to it?

scripts/build.sh
```diff
@@ -27,6 +28,7 @@ while [[ $# -gt 0 ]]; do
       ;;
     --ci)
       ci=true
+      #AHHHH we need to fix this at some point.
```
tannergooding, line 31: What's AHHHH here?

scripts/build.sh
tannergooding: Why not just `generate=true` if `documentation` is `true`, just like we do for `CI` so it flows naturally and works with the rest of the script?

.github/workflows/doxygen.yml
tannergooding: I think yes. Ideally we can do any build step on either system.

scripts/build.sh
capnkenny: Ah, I don't remember if this was even me or not, but this was left as part of a reminder for the xports below it. Previously, we would use the default compiler, but since CI would default to GCC 9 at the time and `clang` defaulted to clang-9, we added the two export lines to ensure that we always use Clang 10 instead.

.github/workflows/doxygen.yml
capnkenny: For this, I followed the example provided from the Checkout v2 Actions Repo (from GH) - https://github.com/actions/checkout

.github/workflows/doxygen.yml
capnkenny: Would you be able to assist with setting up auto-merge? This is something I'm very unfamiliar with 😅

Also, so I understand the flow desired:
Checkout to new feature branch
Add the docs folder that was generated
Commit to the feature branch
Push (to remote)
Request PR between the feature branch and the `gh-pages` branch

Does this seem correct?

.github/workflows/doxygen.yml
capnkenny: Yep, gotta change this to main now

scripts/build.sh
capnkenny: I think it's really just an oversight I had from the first implementation - the thought was to not build since it was unnecessary, we only needed to generate and then build the `Doxygen` target as needed.

I'll correct this

doxygen/CMakeLists.txt
capnkenny: Moved all to `DOXYGEN_OUTPUT_DIRECTORY` - just kept the step that reports if it is already defined (it will print it so we know where the output is)

scripts/build.sh
capnkenny: Rewrote this to follow the general flow for CI

.github/workflows/doxygen.yml
capnkenny: Changed this to main

.github/workflows/doxygen.yml
capnkenny: Tried to implement this (GH doesn't provide a stock "Create Pull Request" action so I had to use one from the marketplace instead.)

scripts/build.sh
capnkenny: This is removed.

.github/workflows/doxygen.yml
capnkenny: This has been replaced - an e-mail was found as per https://github.community/t/github-actions-bot-email-address/17204/3 that will actually utilize the github-actions bot account for commits, PRs, etc.

.github/workflows/doxygen.yml
```diff
@@ -15,16 +15,33 @@ jobs:
...
+
       - name: Generate Documentation
         run: scripts/cibuild.sh --documentation -DDOXYGEN_OUTPUT_DIR=$GITHUB_WORKSPACE/novelrt/docs
 
+      #Mandatory step it appears
+      #Username and Email found from https://github.community/t/github-actions-bot-email-address/17204/3
+      - name: Configure Git
+      - run: |
+          git config user.name "github-actions"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
capnkenny: In reference to this convo: https://github.com/novelrt/NovelRT/pull/257/files/046abc10878bc4185964794d304d061ccd4bda8a#r610989328

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: scripts/cibuild.sh --documentation -DDOXYGEN_OUTPUT_DIR=$GITHUB_WORKSPACE/novelrt/docs
+
+      - name: Commit Documentation Updates
+        if: ${{ success() }}
+        run: |
+          git add ./docs/*
+          git commit -m "Updated documentation - $GITHUB_SHA"
+          git push origin documentation/$GITHUB_SHA
+
+      - name: Create Pull Request
```
tannergooding, line 38: nit: If we can trivially acquire https://github.com/cli/cli, that would be better than depending on a third party CI leg
-- It may even be default on the GH Actions machines

That should just be downloading the zip from the release URL and then running `gh auth logic` (which should pick up `with: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`)
and then doing `gh pr create --title "title" --body "body" --base gh-pages`

I imagine the `create-pull-request@1.0.0` action is doing basically that itself anyways

tannergooding: https://github.blog/2021-03-11-scripting-with-github-cli/#using-gh-in-github-actions

It does indeed come built in

FiniteReality (changes requested):
LGTM after you make the changes Tanner mentioned

.github/workflows/doxygen.yml
capnkenny: Finally got around to playing with GH CLI and testing this. Latest commit should include the change iirc

capnkenny:
Hey @tannergooding - if you could re-review when you get a chance? Thanks!

FiniteReality (approved):
LGTM

doxygen/CMakeLists.txt
```diff
@@ -20,7 +26,7 @@ set(DOCS_MAINPAGE
 foreach(docFile ${DOCS_MAINPAGE})
```
FiniteReality, line 26: Before the merge though, could somebody quickly explain why we need this again? I'm skimming through the logic and I'm not 100% sure what it's doing

capnkenny: Hey @FiniteReality - this is setting the main page to be the contents of `README.md`, and links together the Code of Conduct and the License to the doxygen output (this was added prior to me, but removing it removed all traces of the License, and the main page became blank).

FiniteReality: Weird. I'll have to check the git blame, I guess :sweat_smile:

PR #270, opened 2021-03-11
hotfix: fix dotnet service compilation by explicitly casting ints to bytes

Drevoed:
since https://github.com/dotnet/roslyn/pull/50755 got merged, the changed code now properly infers the resulting expression type to be of type int, so that the explicit conversion is required.


RubyNova (approved):
LGTM, will merge when CI passes. Thanks!

tannergooding (approved):

PR #272, opened 2021-03-20
[WIP] Add additional diagnostics

RyadaProductions:
This PR is to solve issue #30

Things to do for this PR:
- [x] Fix the current frame diagnostics not updating properly.
- [x] Add basic count of loaded shaders to diagnostics.
- [ ]  Add basic count of loaded objects in the scene.

Previously the fps counter wasn't updating if the FPS doesn't change, so if you permanently hit the vsync cap it would never update the total frames rendered/avg
The first commit in this PR takes care of this problem by changing the location of the updateFpsCounter call.

The second commit adds a way to trace the amount of shaders loaded into the system.
It is implemented in a push from the RenderingService to the DebugService to make sure that if we load a shader halfway through the runtime it will still be reflected and updated and not force us to update the shaderload text every frame.
This should save some if not a minor amount of processing power but make it responsive and update on the frame the new shader is loaded.

### DIFF
--- a/include/NovelRT/DebugService.h
+++ b/include/NovelRT/DebugService.h
@@ -16,6 +16,7 @@ namespace NovelRT
     private:
         std::shared_ptr<Graphics::RenderingService> _renderingService;
         std::unique_ptr<Graphics::TextRect> _fpsCounter;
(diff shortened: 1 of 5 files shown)


RubyNova:
PR is dead. Closing.

PR #273, opened 2021-03-25
Generating C# bindings for NovelRT

tannergooding:
This is just the raw 1-to-1 bindings, a proper wrapper can be done in a separate PR.


generation/NovelRT/Audio/include-NrtAudio.h
```diff
@@ -0,0 +1 @@
+#include <NovelRT.h>
```
RubyNova, line 1: Could you clarify what these headers are with only one include line in them? The NovelRT interop headers are self-isolated and shouldn't need `NovelRT.h` unless this is your centralised header solution for interop (for now)?

tannergooding: These (`include-Nrt*.h`) headers are for ClangSharp so the C bindings can be auto-generated. It simplifies the process overall and allows for any tiny fixups if they're ever required.

`NovelRT.h` is the "centralized" header and simply exposes the C API if you reference it via a C compiler or if you explicitly define `NOVELRT_C_API`. This makes it so that I don't have to include all the individual header files here and users don't need to think about what they do or don't want. Instead, they get NovelRT and the linker takes care of trimming anything unused.

RubyNova: I see, sounds good to me!

RubyNova (approved):
I think this PR is definitely Gooding.

generation/remap.rsp
```diff
@@ -0,0 +1,55 @@
+--remap
+AudioServiceHandle*=IntPtr
```
tannergooding, line 2: @capnkenny, I don't like that this is `AudioServiceHandle*`. It implies, to me, that this is a pointer to a handle rather than being a handle itself.

A better name might be `NrtAudioService_t*`, etc. This makes it so you have the "friendly name" as `NrtAudioService` and then the backing hidden struct `NrtAudioService_t`, so bindings (like C#) see `NrtAudioservice_t*`. This matches what things like Vulkan do themselves.

Alternatively, this could be a proper:
```cpp
typedef struct
{
    struct NrtAudioService_t* Handle;
} NrtAudioService;
```

The benefit then is that it isn't a typedef but rather a "proper" wrapper type, which makes it simpler for bindings to just use `NrtAudioService` directly and to "ignore" the underlying handle.

include/NovelRT.Interop/Input/NrtBasicInteractionRect.h
```diff
@@ -23,7 +23,6 @@ extern "C"
     NrtResult Nrt_Input_BasicInteractionRect_setLayer(NrtBasicInteractionRect object, int32_t value);
     int32_t Nrt_Input_BasicInteractionRect_getActive(NrtBasicInteractionRect object);
     NrtResult Nrt_Input_BasicInteractionRect_setActive(NrtBasicInteractionRect object, int32_t value);
-    NrtResult Nrt_Input_BasicInteractionRect_executeObjectBehaviour(NrtBasicInteractionRect object);
```
tannergooding, line 26: There were a couple duplicate methods in NovelRT.Interop (here and below in `NrtStepTimer.h`). I'd like confirmation from @capnkenny that this were proper dupes and not something that was accidentally misnamed.

generation/remap.rsp
RubyNova: we can't use the _t suffix as that's reserved. It's why we're not using it.

tannergooding: We could use `_T` like Vulkan does: https://github.com/KhronosGroup/Vulkan-Headers/blob/master/include/vulkan/vulkan_core.h#L25? There is also `_type`, `HWND__` like Windows does, and several other ways it could be disambiguated.

It's also notably not reserved by C or C++, it is only reserved by POSIX (and therefore "Unix", but I also am skeptical that they would release a type called `NrtAudioService_t` 😉).

image
image

RubyNova: Sure, I just wonder how much it is worth it - also you might want to talk with Monica about that since she explicitly asked us not to do this (and it took us 7 months to get to this point)

RubyNova: cc @FiniteReality

tannergooding: Not to use `*_t` or not to use some explicit suffix (e.g. `*_T`, `*_type`, `*__`, etc) to indicate that the underlying thing is an opaque type and not a handle?

RubyNova: not to use `_t` - I don't mind changing naming conventions if you want to do so, however it is low priority for me because:
- I have graphics work to be doing at the moment myself, and Kenny is working on Audio
- It took us 7 months to even get the C API stable so I'm quite burnt out after 6 rewrites

tannergooding: I'm also fine with not using `_t`, provided one of the other options is picked. Was just calling out that the term `NrtAudioServiceHandle*` will be confusing to external languages as it implies its a pointer to a handle, where in practice the pointer itself is the handle and the backing type is "opaque".

include/NovelRT.h
```diff
@@ -81,167 +84,285 @@
...
+
+#if defined(__cplusplus) && !defined(NOVELRT_C_API)
+  /**
+   * @brief Contains animation features.
+   */
+  namespace NovelRT::Animation {
+    typedef class SpriteAnimator SpriteAnimator;
+  }
+#else // !__cplusplus
+  #include "NovelRT.Interop/Animation/NrtAnimationTypedefs.h"
```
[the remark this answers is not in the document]
capnkenny: IIRC, this was done due to some of the Interop headers requiring a definition for a type that was used in multiple areas, causing redefinitions to occur.

If this is how I'm imagining it, it _should_ be fine to make single file headers for Interop except the fact that we'll have multiple classes' interop versions falling into one file.

Would you be able to confirm? And if true, then @RubyNova do you have objections? (I don't, since it actually makes more sense in my head now than it did before lol)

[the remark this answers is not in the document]
tannergooding: That sounds about right. Vulkan handles this by declaring the handles (typedefs) all up front at the top of `vulkan_core.h` (https://github.com/KhronosGroup/Vulkan-Headers/blob/master/include/vulkan/vulkan_core.h#L62) and then defining everything else below it.

That's basically what these `#include` in `NovelRT.h` are doing. It's ensuring all typedefs are defined before the main headers are included and then the headers defining the methods and such are below that.
It's basically required for something like ClangSharp because otherwise the header passed into ClangSharp has to itself ensure its doing the right includes and typedefs in the right order; vs just including `NovelRT.h` and traversing the files that define the types you want.

generation/remap.rsp
FiniteReality: I was saying to not use `_t` prefixes, as they are reserved by POSIX. In my interop code for animation, I've done this:

```c
typedef struct KeyframeSequenceHandle* NrtKeyframeSequence;
typedef float (*NrtEasingFunc)(float time, float begin, float change, float duration, void* userdata);
```

And that lets me do this:
```cpp
struct KeyframeSequence {
    float current_time;
    float last_value;
    size_t current_index;

    std::vector<Keyframe> keyframes;
};

extern "C" {

NrtKeyframeSequence Nrt_KeyframeSequence_create()
{
    return reinterpret_cast<NrtKeyframeSequence>(new KeyframeSequence());
}

// etc.
}
```

The name of the struct in the typedef isn't super important; since it's opaque and we're using it for its unique identity in the pointer, to avoid things like passing an NrtAnimationStateMachine to a function expecting an NrtKeyframeSequence

tannergooding: > The name of the struct in the typedef isn't super important; since it's opaque and we're using it for its unique identity in the pointer, to avoid things like passing an NrtAnimationStateMachine to a function expecting an NrtKeyframeSequence

I agree with the latter half but not the first. The name is important because its what other systems like C# will end up seeing and exposing by default since they don't have typedefs and since struct wrappers aren't equivalent.

Languages like C# have basically three options:
* use `void*`
* use `IntPtr`
* use `NrtAudioServiceHandle*`

The name of the last implies its a pointer to a handle, when in fact the names are effectively reversed. That is in `typedef struct KeyframeSequenceHandle* NrtKeyframeSequence;`, the RHS (`NrtKeyframeSequence`) is the handle, while the LHS (`KeyframeSequenceHandle*`) is the underlying struct pointer and so seeing `typedef struct NrtKeyframeSequence* NrtKeyframeSequenceHandle;` makes "more" sense to me. Most other libraries do something like `typedef struct NrtKeyframeSequence_T* NrtKeyframeSequence;` however, often via some `DECLARE_HANDLE(NrtKeyFrameSequence)` helper.

FiniteReality: In that case, we should probably just do `typedef struct KeyframeSequence* NrtKeyframeSequence`, since as long as there's no `struct KeyframeSequence { ... };` in the header files, C code will see an opaque type

tannergooding: I think we should continue prefixing all the types with `Nrt` to avoid ambiguity. The `_T` isn't "new" and is used by other libraries. Since it is uppercase, it is also POSIX compatible.

FiniteReality: I'm fine with `typedef struct NrtKeyframeSequence* NrtKeyframeSequence;` - that makes both of these declarations valid:

```c
void doThings()
{
    struct NrtKeyframeSequence* sequence; // referring to the tag of the struct
    NrtKeyframeSequence sequence2; // referring to the type name
}
```

capnkenny: Would this change need to be done prior to this being merged? I'm not sure if we'd want to handle this in the same or a separate PR (my gut tells me it'd be a separate one that would block this from being merged)

include/NovelRT.Interop/Input/NrtBasicInteractionRect.h
capnkenny: These were dupes, my bad 😅

include/NovelRT.h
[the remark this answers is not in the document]
capnkenny: Like the other convo - handle in separate PR? Or handle in this one and adjust the bindings appropriately?

capnkenny (commented):
LGTM - pending results of ongoing convos 👍

generation/remap.rsp
tannergooding: No, it would be a separate PR and the bindings would simply be regenerated after that goes in.

tannergooding:
Merging. We settled on `typedef struct NrtKeyframeSequence* NrtKeyframeSequenceHandle` in Discord chat.

PR #276, opened 2021-03-27
Consistently use `typedef struct NrtName* NrtNameHandle`

tannergooding:
This simplifies and makes the C API more consistent. It also makes it easier for tools such as ClangSharp to traverse the files


include/NovelRT.Interop/DotNet/NrtDotNetTypedefs.h
```diff
@@ -1,19 +1,18 @@
 // Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
 // for more information.
 
-#include <NovelRT.Interop/NrtInteropErrorHandlingInternal.h>
-#include <NovelRT.Interop/NrtInteropUtils.h>
```
RubyNova, line 5: What happened here? how did includes get replaced with an include guard?

include/NovelRT.Interop/Ecs/NrtEcsTypedefs.h
```diff
@@ -0,0 +1,41 @@
...
+    typedef void (*NrtSystemUpdateFnPtr)(NrtTimestamp, NrtCatalogueHandle);
+
+    typedef NrtAtom NrtEntityId;
+    typedef NrtAtom NrtComponentTypeId;
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // NOVELRT_INTEROP_ECS_ECSTYPEDEFS_H
```
RubyNova, line 41: How come these have all been extracted from their relevant headers?

include/NovelRT.Interop/Input/NrtInteractionService.h
```diff
@@ -4,32 +4,24 @@
 #ifndef NOVELRT_INTEROP_INPUT_INTERACTIONSERVICE_H
 #define NOVELRT_INTEROP_INPUT_INTERACTIONSERVICE_H
 
-#include "../Maths/NrtGeoVector2F.h"
-#include "../NrtInteropUtils.h"
-#include "../NrtTransform.h"
-#include "../Windowing/NrtWindowingService.h"
-#include "NrtBasicInteractionRect.h"
-#include "NrtKeyCode.h"
-#include "NrtKeyStateFrameChangeLog.h"
```
RubyNova, line 13: How come these are suddenly not needed? This seems wrong to me.

include/NovelRT.Interop/Maths/NrtQuadTree.h
```diff
@@ -1,44 +1,39 @@
...
 // for more information.
 
-#ifndef NOVELRT_MATHS_QUADTREE_C_H
-#define NOVELRT_MATHS_QUADTREE_C_H
+#ifndef NOVELRT_INTEROP_MATHS_QUADTREE_H
+#define NOVELRT_INTEROP_MATHS_QUADTREE_H
 
-#include "NrtGeoBounds.h"
-#include "NrtQuadTreePoint.h"
-#include <stddef.h>
```
RubyNova, line 9: All these missing includes are worrying me. I explicitly added them so that if you had access to the handle, you'd have access to the relevant functionality associated with that handle.

include/NovelRT.Interop/NrtDebugService.h
```diff
@@ -4,23 +4,20 @@
 #ifndef NOVELRT_INTEROP_DEBUGSERVICE_H
 #define NOVELRT_INTEROP_DEBUGSERVICE_H
 
-#include "Graphics/NrtRenderingService.h"
-#include "Utilities/NrtCommonEvents.h"
```
RubyNova, line 8: More missing includes. See previous comment. I'll stop commenting on them after this one if I see any others.

RubyNova (commented):
I have some questions and also some concerns about breaking changes here.

I also noticed we now have an additional reliance on NovelRT.h which I thought we were trying to move away from to help with compile time issues.

I need some clarification here. sorry. :(

include/NovelRT.Interop/DotNet/NrtDotNetTypedefs.h
tannergooding: This is GitHub seeing a delete and file addition where the contents between the two files is similar enough it wants to show the diff as a rename.

The old file was removed (and merged into `NrtTypedefs.h`) and the new file is completely different/unrelated.

include/NovelRT.Interop/Ecs/NrtEcsTypedefs.h
tannergooding: Because there was an inconsistency overall across the headers.

This allows all typedefs to be defined up front and simplifies the dependency graph between files. It allows someone working with bindings to see all typedefs/opaque handles in one location so they can easily create the relevant renames/remappings.

(Many libraries, including Vulkan, Xlib, DirectX, etc centralize their own forward declarations at the top of the central include file as well).

include/NovelRT.Interop/Input/NrtInteractionService.h
tannergooding: Because the enums and opaque handles were moved to a namespace specific `Nrt*Typedefs.h` folder, which helps with generating bindings and simplifies the file dependency chain.

include/NovelRT.Interop/Maths/NrtQuadTree.h
tannergooding: The C files will primarily be used by interop bindings, not by actual C code, in which case it is more beneficial (in my experience) to have the typedefs centrally managed and defined.

As mentioned above, many many native libraries centralize their opaque types and definitions. Part of this is because they are often needed by many things below that point, but also because it allows you to easily see all the types at once: https://github.com/KhronosGroup/Vulkan-Headers/blob/master/include/vulkan/vulkan_core.h#L57. See also: `X11/X.h`, most of the Windows SDK (particularly DirectX and friends), `gl.h`, many of our own dependencies, etc.

RubyNova (approved):

PR #278, opened 2021-03-28
Porting the CAPI sample to the C# bindings side

tannergooding:
This adds the CAPI sample for the C# bindings and adjusts the `Nrt_append*` APIs to not use varargs, as that is not compatible with .NET.

This also cleans up the .NET cmakelist files to build and perform their up to date checks based on the correct set of inputs.

### DIFF
--- a/include/NovelRT.Interop/Utilities/NrtMisc.h
+++ b/include/NovelRT.Interop/Utilities/NrtMisc.h
@@ -13,8 +13,8 @@ extern "C"
 
     const char* Nrt_getExecutablePath();
     const char* Nrt_getExecutableDirPath();
-    const char* Nrt_appendFilePath(int32_t numberOfArgs, ...);
-    const char* Nrt_appendText(int32_t numberOfArgs, ...);
+    const char* Nrt_appendFilePath(int32_t numberOfArgs, const char* const* args);
+    const char* Nrt_appendText(int32_t numberOfArgs, const char* const* args);
 
 #ifdef __cplusplus
 }
--- a/samples/NovelRT.DotNet.Sample/CMakeLists.txt
+++ b/samples/NovelRT.DotNet.Sample/CMakeLists.txt
@@ -1,0 +1,86 @@
+find_package(Dotnet ${NOVELRT_DOTNET_VERSION} REQUIRED)
+
+set(DOTNET_SAMPLE_CSPROJ ${CMAKE_CURRENT_LIST_DIR}/NovelRT.DotNet.Sample.csproj)
+
+set(DOTNET_SAMPLE_SOURCES
+  NovelRT.DotNet.Sample.runtimeconfig.json
+  Program.cs
+)
+
+set(DOTNET_SAMPLE_OPTIONS
+  /nologo
+  /p:IntermediateOutputPath=${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/Dotnet_Sample_Build.dir/
+  /p:MSBuildProjectExtensionsPath=${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/Dotnet_Sample_Build.dir/
+  /p:OutDir=${CMAKE_CURRENT_BINARY_DIR}/
+  /p:PublishDir=${CMAKE_CURRENT_BINARY_DIR}/publish/
+)
+
+execute_process(
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:Restore
+)
+
+execute_process(
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:GetCMakeOutputAssembly
+  OUTPUT_VARIABLE DOTNET_SAMPLE_OUTPUT_ASSEMBLY
+)
+string(STRIP "${DOTNET_SAMPLE_OUTPUT_ASSEMBLY}" DOTNET_SAMPLE_OUTPUT_ASSEMBLY)
+
+execute_process(
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:GetCMakeOutputByproducts
+  OUTPUT_VARIABLE DOTNET_SAMPLE_OUTPUT_BYPRODUCTS
+)
+string(STRIP "${DOTNET_SAMPLE_OUTPUT_BYPRODUCTS}" DOTNET_SAMPLE_OUTPUT_BYPRODUCTS)
+
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/publish)
+
+add_custom_command(
+  OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${DOTNET_SAMPLE_OUTPUT_ASSEMBLY}
+  DEPENDS ${DOTNET_SAMPLE_SOURCES} ${DOTNET_SAMPLE_CSPROJ}
+  BYPRODUCTS ${DOTNET_SAMPLE_OUTPUT_BYPRODUCTS}
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:Build,Publish
+  COMMENT "Building ${DOTNET_SAMPLE_CSPROJ}"
+)
+
+add_custom_target(Dotnet_Sample_Build ALL
+  DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/${DOTNET_SAMPLE_OUTPUT_ASSEMBLY}
+)
+
+add_executable(Dotnet_Sample IMPORTED GLOBAL)
+add_dependencies(Dotnet_Sample Dotnet_Build Dotnet_Sample_Build)
+
+set_target_properties(Dotnet_Sample
+  PROPERTIES
+  IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/publish/IF_YOU_SEE_THIS_YOU_TRIED_TO_RUN_THE_DOTNET_TARGET
+)
+
+add_custom_command(
+  TARGET Dotnet_Sample_Build POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+    $<TARGET_FILE_DIR:Engine>
+    ${CMAKE_CURRENT_BINARY_DIR}
+)
+
+add_custom_command(
+  TARGET Dotnet_Sample_Build POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+    $<TARGET_FILE_DIR:Engine>
+    ${CMAKE_CURRENT_BINARY_DIR}/publish
+)
+
+add_custom_command(
+  TARGET Dotnet_Sample_Build POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+    $<TARGET_FILE_DIR:Interop>
+    ${CMAKE_CURRENT_BINARY_DIR}
+)
+
+add_custom_command(
+  TARGET Dotnet_Sample_Build POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+    $<TARGET_FILE_DIR:Interop>
+    ${CMAKE_CURRENT_BINARY_DIR}/publish
+)
--- a/samples/NovelRT.DotNet.Sample/NovelRT.DotNet.Sample.csproj
+++ b/samples/NovelRT.DotNet.Sample/NovelRT.DotNet.Sample.csproj
@@ -1,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
+    <SelfContained>true</SelfContained>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../src/NovelRT.DotNet/NovelRT.DotNet.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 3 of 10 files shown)


include/NovelRT.Interop/Utilities/NrtMisc.h
```diff
@@ -13,8 +13,8 @@ extern "C"
 
     const char* Nrt_getExecutablePath();
     const char* Nrt_getExecutableDirPath();
-    const char* Nrt_appendFilePath(int32_t numberOfArgs, ...);
-    const char* Nrt_appendText(int32_t numberOfArgs, ...);
+    const char* Nrt_appendFilePath(int32_t numberOfArgs, const char* const* args);
```
tannergooding, line 16: `varargs` are not compatible with .NET and so we need explicit lists instead.

samples/NovelRT.DotNet.Sample/Program.cs
```diff
@@ -0,0 +1,391 @@
...
+
+            if (error != null)
+            {
+                Nrt_LoggingService_logErrorLine(console, error);
+                return -1;
+            }
+
+            sbyte* path;
+
+            fixed (byte* pathPart = Encoding.UTF8.GetBytes("Resources"))
```
tannergooding, line 69: The C# bindings don't have any of the "nice" surface area or helpers that take `string` yet, so need to explicitly pin and convert in the sample.

This will be improved in a later PR.

samples/NovelRT.DotNet.Sample/Program.cs
```diff
@@ -0,0 +1,391 @@
...
+using System;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Text;
+using NovelRT.Interop;
+using static System.Net.Mime.MediaTypeNames;
+using static NovelRT.Interop.NovelRT;
+using static NovelRT.Interop.NrtBool;
+using static NovelRT.Interop.NrtResult;
```
RubyNova, line 12: Not a complaint, just commenting how this confused me at first, It all makes sense now I see these two using statics ahahahahaha

RubyNova (approved):
LGTM!

samples/NovelRT.DotNet.Sample/Program.cs
tannergooding: They just exist so you can do `NRT_SUCCESS` and not `NrtResult.NRT_SUCCESS`, makes it simpler to read and port the logic 😄

RubyNova: yep figured that out after a few minutes of me going "wut?"lmao

include/NovelRT.Interop/Utilities/NrtMisc.h
RubyNova: I see....is there no way to just have both? Or too much overhead do you thnk?

tannergooding: I think its more overhead than its worth. It buys very little over putting the types in a stack allocated buffer, basically just saves a line of code per call.

I think it would be better to define common "overloads" for taking in 2-3 parts. They just need unique names because "C".

RubyNova: The underlying impl is variadic though, so ultimately I don't think having multiple "overloads" would really work out too well.

This is probably the best we can do.

tannergooding: It works out the same? It's basically a question of forcing the user to do:
```cpp
const char* const pathParts[2] = { part1, part2 };
Nrt_appendFilePath(2, pathParts);
```

or allowing them to do the following, where it does (at the minimum) the above (or possibly something more efficient:
```cpp
Nrt_appendFilePath2(part1, part2);
```

`Nrt_appendText` and `Nrt_appendFilePath` both use `string::append` which is non-variadic: https://www.cplusplus.com/reference/string/string/append/
The same would be true if we used `filesystem::path` for concatenating.

This is similar to how .NET has `string.Format(string format, object arg)`, `string.Format(string format, object arg1, object arg2)`, `string.Format(string format, object arg1, object arg2, object arg3)` and `string.Format(string format, params object[] args)`, because the first three can be more efficient

RubyNova: Wait really? Then why are we using variadic for the API surface itself?

tannergooding: > Then why are we using variadic for the API surface itself?

Could you clarify?

tannergooding:
@FiniteReality, could you review the CMake changes and @capnkenny could you review the CAPI changes?

samples/NovelRT.DotNet.Sample/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+find_package(Dotnet ${NOVELRT_DOTNET_VERSION} REQUIRED)
+
+set(DOTNET_SAMPLE_CSPROJ ${CMAKE_CURRENT_LIST_DIR}/NovelRT.DotNet.Sample.csproj)
+
+set(DOTNET_SAMPLE_SOURCES
```
FiniteReality, line 5: This can just be `SOURCES`

samples/NovelRT.DotNet.Sample/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+find_package(Dotnet ${NOVELRT_DOTNET_VERSION} REQUIRED)
+
+set(DOTNET_SAMPLE_CSPROJ ${CMAKE_CURRENT_LIST_DIR}/NovelRT.DotNet.Sample.csproj)
```
FiniteReality, line 3: This can just be `CSPROJ`

samples/NovelRT.DotNet.Sample/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+
+set(DOTNET_SAMPLE_SOURCES
+  NovelRT.DotNet.Sample.runtimeconfig.json
+  Program.cs
+)
+
+set(DOTNET_SAMPLE_OPTIONS
```
FiniteReality, line 10: This can just be `OPTIONS`

samples/NovelRT.DotNet.Sample/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+
+execute_process(
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:Restore
+)
+
+execute_process(
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:GetCMakeOutputAssembly
+  OUTPUT_VARIABLE DOTNET_SAMPLE_OUTPUT_ASSEMBLY
```
FiniteReality, line 26: This could realistically just be `OUTPUT_ASSEMBLY`

samples/NovelRT.DotNet.Sample/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+)
+string(STRIP "${DOTNET_SAMPLE_OUTPUT_ASSEMBLY}" DOTNET_SAMPLE_OUTPUT_ASSEMBLY)
+
+execute_process(
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_SAMPLE_CSPROJ} ${DOTNET_SAMPLE_OPTIONS}
+  /t:GetCMakeOutputByproducts
+  OUTPUT_VARIABLE DOTNET_SAMPLE_OUTPUT_BYPRODUCTS
```
FiniteReality, line 33: This could realistically just be `OUTPUT_BYPRODUCTS`

samples/NovelRT.DotNet.Sample/NovelRT.DotNet.Sample.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <TargetFramework>net5.0</TargetFramework>
+    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
+    <SelfContained>true</SelfContained>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../src/NovelRT.DotNet/NovelRT.DotNet.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
```
FiniteReality, line 15: We should probably just rely on the CMake targets for copying these resources for now; just to be consistent with the rest of the project - and also for making sure CMake knows about any dependencies between the targets.

src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -1,8 +1,59 @@
-find_package(Dotnet 5.0.100 REQUIRED)
+find_package(Dotnet ${NOVELRT_DOTNET_VERSION} REQUIRED)
 
-set(CSPROJ ${CMAKE_CURRENT_LIST_DIR}/NovelRT.DotNet.csproj)
+set(DOTNET_CSPROJ ${CMAKE_CURRENT_LIST_DIR}/NovelRT.DotNet.csproj)
```
FiniteReality, line 3: This change is likely not needed.

src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -1,8 +1,59 @@
...
 
-set(OPTIONS
+set(DOTNET_SOURCES
```
FiniteReality, line 5: This change is likely not needed.

src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -1,8 +1,59 @@
...
+
+  Interop/Timing/NovelRT.cs
+
+  Interop/Utilities/NovelRT.cs
+
+  Interop/Windowing/NovelRT.cs
+  Interop/Windowing/NrtWindowMode.cs
+)
+
+set(DOTNET_OPTIONS
```
FiniteReality, line 56: This change is likely not needed.

src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -11,44 +62,45 @@ set(OPTIONS
...
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_CSPROJ} ${DOTNET_OPTIONS}
   /t:Restore
 )
 
 execute_process(
-  COMMAND ${Dotnet_PROGRAM} msbuild ${CSPROJ} ${OPTIONS}
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_CSPROJ} ${DOTNET_OPTIONS}
   /t:GetCMakeOutputAssembly
-  OUTPUT_VARIABLE OUTPUT_ASSEMBLY
+  OUTPUT_VARIABLE DOTNET_OUTPUT_ASSEMBLY
```
FiniteReality, line 72: This change is likely not needed.

src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -11,44 +62,45 @@ set(OPTIONS
...
 )
-string(STRIP "${OUTPUT_ASSEMBLY}" OUTPUT_ASSEMBLY)
+string(STRIP "${DOTNET_OUTPUT_ASSEMBLY}" DOTNET_OUTPUT_ASSEMBLY)
 
 execute_process(
-  COMMAND ${Dotnet_PROGRAM} msbuild ${CSPROJ} ${OPTIONS}
+  COMMAND ${Dotnet_PROGRAM} msbuild ${DOTNET_CSPROJ} ${DOTNET_OPTIONS}
   /t:GetCMakeOutputByproducts
-  OUTPUT_VARIABLE OUTPUT_BYPRODUCTS
+  OUTPUT_VARIABLE DOTNET_OUTPUT_BYPRODUCTS
```
FiniteReality, line 79: This change is likely not needed.

samples/NovelRT.DotNet.Sample/NovelRT.DotNet.Sample.csproj
FiniteReality: To clarify, this is in case we add any pre-processing steps to our resources which may need special handling (e.g. precompiling shaders to SPIR-V)

FiniteReality (dismissed):
Seems fine for the most part, but a lot of names could likely be simplified, as the CMake `set()` command only operates within the current scope (unless `PARENT_SCOPE` is specified), and each CMakeLists is in its own scope IIRC.

samples/NovelRT.DotNet.Sample/CMakeLists.txt
tannergooding: We weren't doing this elsewhere (see `ENGINE_SOURCES` for example) and that's why I mirrored it here.

samples/NovelRT.DotNet.Sample/NovelRT.DotNet.Sample.csproj
tannergooding: Relying on CMake doesn't get it to the publish folder, so we need two steps. Likewise, its problematic for anyone trying to use `dotnet build`, which otherwise "works".

We likely need a good setup to ensure that `dotnet build` and `cmake` ultimately do the same thing.

samples/NovelRT.DotNet.Sample/CMakeLists.txt
FiniteReality: Yeah; those places are actually unnecessarily prefixed too, it's just that "if it ain't broke, don't fix it" applied :stuck_out_tongue:

samples/NovelRT.DotNet.Sample/NovelRT.DotNet.Sample.csproj
FiniteReality: Since we're controlling the publish directory from CMake, we can just manually copy to it. It's not the *greatest* solution, and I agree we need a better solution, but that's not going to be fun to implement.

samples/NovelRT.DotNet.Sample/CMakeLists.txt
tannergooding: Happy to simplify, just to be clear, just wanted to call the above out since we should likely do it everywhere.

FiniteReality: Yeah, we should.

include/NovelRT.Interop/Utilities/NrtMisc.h
capnkenny: I think Matt was asking why were even using variadic params in the API in the first place. If so, then the answer is - "I didn't know about this at the time 😃"

If this is working as expected, then I'm okay with it. It seems like a simple enough change - varargs can get dicey IIRC anyways.

capnkenny (commented):
LGTM on the CAPI side of things 👍

tannergooding:
Merging. As per the discussion with Monica in Discord, I'll be submitting a follow up PR to address the CMake requests.

PR #280, opened 2021-04-03
Set OUTPUT_NAME and regenerate the C# bindings

tannergooding:


src/NovelRT.DotNet/Interop/Animation/NovelRT.cs
```diff
@@ -7,108 +7,87 @@ namespace NovelRT.Interop
 {
     public static unsafe partial class NovelRT
     {
-        [DllImport("Interop", ExactSpelling = true)]
-        [return: NativeTypeName("NrtResult")]
-        public static extern int Nrt_SpriteAnimator_create([NativeTypeName("NrtNovelRunnerHandle")] IntPtr runner, [NativeTypeName("NrtImageRectHandle")] IntPtr rect, [NativeTypeName("NrtSpriteAnimatorHandle *")] IntPtr* outputAnimator);
+        [DllImport("NovelRT.Interop", ExactSpelling = true)]
+        public static extern NrtResult Nrt_SpriteAnimator_create([NativeTypeName("NrtNovelRunnerHandle")] IntPtr runner, [NativeTypeName("NrtImageRectHandle")] IntPtr rect, [NativeTypeName("NrtSpriteAnimatorHandle *")] IntPtr* outputAnimator);
```
tannergooding, line 11: There are three changes from regenerating the bindings:
1. The `DllImport` has changed from `Interop` to `NovelRT.Interop`
2. Where an enum was being taken or returned, we now take return the actual `enum` rather than the `int` it was typedefd to
3. Where `NativeTypeName` and the managed type name differed only in whitespace, the attribute was removed

src/NovelRT.Interop/CMakeLists.txt
```diff
@@ -64,6 +64,9 @@ set(SOURCES
 
 add_library(Interop SHARED ${SOURCES})
 add_dependencies(Interop Engine)
+
+set_property(TARGET Interop PROPERTY OUTPUT_NAME "NovelRT.Interop")
```
tannergooding, line 68: CC. @FiniteReality for the cmake change

FiniteReality (approved):
Seems good to me. I'm not sure about the name `NovelRT.Interop` but I can't think of anything better, honestly.

PR #281, opened 2021-04-09
Updating Atom and GetComponentTypeId to work across multiple libraries

tannergooding:
This updates the various `Atom::getNext*Id()` and the `template<typename TComponent> ComponentTypeId GetComponentTypeId()` methods to work as expected when two otherwise independent libraries use the methods from the same application.


include/NovelRT/Atom.h
```diff
@@ -55,54 +55,24 @@ namespace NovelRT
             return _value >= other._value;
         }
 
-        operator uintptr_t() noexcept
+        operator uintptr_t() const noexcept
```
tannergooding, line 58: This not being `const` meant that C/C++ thought it was mutating the value and so it couldn't be used with a `const Atom`.

include/NovelRT/Atom.h
```diff
@@ -55,54 +55,24 @@ namespace NovelRT
...
         {
             return _value;
         }
 
         // TODO: These should be internal to NovelRT
 
-        static Atom getNextEventHandlerId() noexcept
-        {
-            static std::atomic_uintptr_t _nextEventHandlerId(0);
```
tannergooding, line 67: These being in the header meant that `A.dll` and `B.exe` would each have their own storage for the static local here, resulting in them not having consistent world views of the "next id"

include/NovelRT/Ecs/EcsUtils.h
```diff
@@ -20,11 +24,11 @@ namespace NovelRT::Ecs
      * @tparam TComponent The component type to get the ID for.
      * @return ComponentTypeId The numerical ID of the component type as an Atom.
      */
-    template<typename TComponent>[[nodiscard]] ComponentTypeId GetComponentTypeId() noexcept
+    template<typename TComponent> [[nodiscard]] ComponentTypeId GetComponentTypeId() noexcept
     {
-        static const Atom id = Atom::getNextComponentTypeId();
-        return id;
+        auto componentTypeId = GetComponentTypeIds()[typeid(TComponent)];
```
tannergooding, line 29: MSVC and Clang differ in behavior here. For Clang, this was working as expected and `A.dll` and `B.exe` would share a storage slot for `T` if the method was marked `extern`.

For MSVC, however, it doesn't look for exports if `B.exe` directly contains a definition itself (which makes sense IMO, prefer local over external) but that causes `A.dll` and `B.exe` to have inconsistent world views for a given `T`.

This resolves that issue by using `RTTI`, specifically it uses an `unordered_map<type_index, ComponentTypeId>`. `typeid(T)` for a statically known, non-polymorphic `T` is a constant time lookup and so we don't expect any issues here around perf.

include/NovelRT/Ecs/EcsUtils.h
```diff
@@ -20,11 +24,19 @@ namespace NovelRT::Ecs
...
-    template<typename TComponent>[[nodiscard]] ComponentTypeId GetComponentTypeId() noexcept
+    template<typename TComponent> [[nodiscard]] ComponentTypeId GetComponentTypeId() noexcept
     {
-        static const Atom id = Atom::getNextComponentTypeId();
-        return id;
+        auto& componentTypeId = GetComponentTypeIds()[typeid(TComponent)];
+
+        if (componentTypeId == Atom(0))
+        {
+            // TODO: Two threads each trying to get a component type ID for T
```
tannergooding, line 33: The simplest thing here is probably to jus do a:
```
if (componentTypeId == Atom(0))
{
    if (Interlocked.CompareExchange(&componentTypeId, expected: 0, desired: nuint.MaxValue) == 0)
    {
        componentTypeId = Atom::getNextComponentTypeId();
    } 
    else
    {
        while (Volatile.Read(&componentTypeId) == nuint.MaxValue) {
            // spin
        }
    }
}
```

In the common case, this will never be hit or the thread will have free access to update the memory
and in the uncommon case, one thread will likely do a single spin

include/NovelRT/Ecs/EcsUtils.h
RubyNova: Could we add the `std::is_trivially_copyable` trait check to this? Would make it consistent with `NovelRT::Ecs:SparseSet<T>` :D

tannergooding: Done

Directory.Build.props
```diff
@@ -22,7 +22,6 @@
     <BaseIntermediateOutputPath>$(BaseArtifactsPath)obj</BaseIntermediateOutputPath>
     <DebugType>embedded</DebugType>
     <EnableDefaultNoneItems>false</EnableDefaultNoneItems>
-    <Nullable>enable</Nullable>
```
RubyNova: I've just seen this in the commit history - we aren't using NRTs for C#?

tannergooding: Not right now. `Ink` isn't using them itself and I don't want to block this PR on figuring out the correct manual bindings.

RubyNova: Ah okay, understandable.

tests/Directory.Build.targets
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+
+  <!--
+    Directory.Build.targets is automatically picked up and imported by
+    Microsoft.Common.targets. This file needs to exist, even if empty so that
+    files in the parent directory tree, with the same name, are not imported
+    instead. The import fairly late and most other props/targets will have
```
RubyNova, line 8: Typo? The -> They, no?

RubyNova (approved):
LGTM

PR #283, opened 2021-04-23
Removed usage of std::byte.

RubyNova:
As per the title, this is a result of the discussion of the C++ specification and the precise behaviour of `std::byte` over `uint8_t`. This change is also being introduced to make the types across C and C++ a tad more consistent.


capnkenny (approved):
LGTM

PR #292, opened 2021-04-27
Ecs: Catalogue & ShimObject + skeletal ecs classes

Statharas:
Fully implemented Catalogue bindings and provided ShimObject as a base Ecs class.

Created skeletal classes for referencing.


RubyNova (changes requested):
Overall a good start - however, since this is the idiomatic C# API wrappers, it probably shouldn't live in the Interop folder. Subsequently, its namespace should be `NovelRT.Ecs`. :)

PR #297, opened 2021-04-30
Hotfix for .NET runtimeconfig.json.

RubyNova:


src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -94,13 +94,19 @@ add_custom_target(Dotnet_Build ALL
...
 add_executable(Dotnet IMPORTED GLOBAL)
 add_dependencies(Dotnet Dotnet_Build)
 set_target_properties(Dotnet
   PROPERTIES
   IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/IF_YOU_SEE_THIS_YOU_TRIED_TO_RUN_THE_DOTNET_TARGET
 )
 
+file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/NovelRT.DotNet.runtimeconfig.json
+        DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+
```
FiniteReality, line 108: Also unnecessary here, but again, not a blocker

FiniteReality (approved):
We should probably open an issue to investigate why we need this workaround. From everything I can see, it should be working fine.

PR #304, opened 2021-05-02
Expose opaque context parameters for interop APIs

FiniteReality:
I did a rough grep for any function pointers we expose (`egrep -r '\(\*\w+\)\(.*\)'`) and updated them to also expose opaque context parameters, which can be specified by the user or interop library, so that we can easily pass through data through our interop bindings.

Furthermore, this PR also simplifies some cases where we were using global state to store function pointers, by replacing them wholesale with uses of `std::bind`. Further cleanup can likely be done in these cases - this is just the minimum set of changes that I could easily find to enable easier interop scenarios.

### DIFF
--- a/samples/CAPI/main.c
+++ b/samples/CAPI/main.c
@@ -83,7 +83,7 @@ void moveNovelChan(NrtTimestamp delta, void* context)
 
             int size = snprintf(NULL, 0, "Flipped X axis movement. Count: %i", bounces);
             char* buf = alloca(size + 1);
-            snprintf(buf, sizeof(buf), "Flipped X axis movement. Count: %i", bounces);
+            snprintf(buf, size + 1, "Flipped X axis movement. Count: %i", bounces);
             Nrt_LoggingService_logInfoLine(console, buf);
         }
     }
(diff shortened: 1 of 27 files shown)


samples/CAPI/main.c
```diff
@@ -121,7 +147,7 @@ void moveNovelChan(NrtTimestamp delta)
 }
 
 // Function to interact with Ink
-void interactWithNovelChan()
+void interactWithNovelChan(void* context)
```
RubyNova, line 150:
```suggestion
void InteractWithNovelChan(void* context)
```

RubyNova (approved):
Bar my one nit about our new code style, I think this is fine. Just make clang-format happy and we should be good to go.

samples/CAPI/main.c
FiniteReality: I ran clang-format on the entire repo and it started changing a lot of things; especially things I didn't touch. I'm not sure what our current policy is on large formatting changes like that, so I figured I'd leave the formatting changes out for now.

RubyNova: Use our patch. Not sure why that's happening to you, but yeah. Patch is usually fine.

Also Windows ain't building lol.

PR #306, opened 2021-05-05
Nullptr return code parity

RubyNova:
This addresses an issue with our return codes. I changed some signatures when I spotted them to try and help with parity - I must admit I didn't try too hard on this front as half of these are being rewritten anyway.


capnkenny (approved):
This PR brings up some questions on how we're handling parts of the C API - as such, expect questions from me until it's cut and dry and clear :D

Until then, for the purpose of this PR - LGTM

PR #315, opened 2021-06-15
Fixed Rect getColourConfig() and added a destroy method

jeuxjeux20:
- Added a new `destroy` methods to `BasicFillRect`, `ImageRect` and `TextRect`
- Fixed `getColourConfig()` (`BasicFillRect`, `TextRect`) and `getColourTint()` (`ImageRect`) segfault bug

### DIFF
--- a/include/NovelRT.Interop/Graphics/NrtBasicFillRect.h
+++ b/include/NovelRT.Interop/Graphics/NrtBasicFillRect.h
@@ -28,7 +28,6 @@ extern "C"
                                                      NrtRenderObjectHandle* outputRenderObject);
     NrtResult Nrt_BasicFillRect_destroy(NrtBasicFillRectHandle rect);
 
-
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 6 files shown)


include/NovelRT.Interop/Graphics/NrtBasicFillRect.h
```diff
@@ -11,6 +11,9 @@ extern "C"
 {
 #endif
 
+    // destroy
+    NrtResult Nrt_BasicFillRect_destroy(NrtBasicFillRectHandle rect);
```
RubyNova, line 15: Conventionally, destroy methods are the last method of the type to be defined. Can we do that here and in the other places you have added these? And can we please make the cpp files mirror these declarations?

RubyNova (changes requested):
Just a minor change. :)

include/NovelRT.Interop/Graphics/NrtBasicFillRect.h
jeuxjeux20: Should be done now.

RubyNova (approved):
LGTM

PR #324, opened 2021-08-13
Make C# / Ink support conditional

### LINKED ISSUE #319: Split `NovelRT::Ink` off to be compiled conditionally (Temporarily)
As part of the SDK work, we will need to ensure that Ink is conditional as people using pure C++ will not necessarily need to rely and/or want to include Ink, which forces a `dotnet` dependency on end users.

For now, until we refactor the Engine to have as part of language bindings, we'll use a conditional define to include parts of the Ink codebase (including a `conan` option to by default include it)

capnkenny:
This PR adds a CMake and C++ preprocessor definition that controls whether or not any C# code is built.

Tested WITH and WITHOUT Ink - Samples and Tests should be passing at this time.

Default behaviour is to _include_ the C# / Ink support and C# bindings.

Resolves #319


capnkenny:
As an aside, this PR also removes Lua support until we can get proper first-party support built in (after discussing with @RubyNova )

RubyNova:
Does this PR prevent any Ink related code from being used or built in the sample projects? I believe in Simple we invoke the Ink runtime.

RubyNova (approved):
LGTM

PR #326, opened 2021-08-22
Ported `conanfile` to .py from .txt

### LINKED ISSUE #320: Port `conanfile` to Python version
In order to prepare for packaging to `conan`, we'll need to port our `conanfile.txt` to `conanfile.py`. Once this is done, we'll be able to package our projects properly and pull down as needed.

capnkenny:
Implemented changes necessary to have Conan use the `conanfile.py` file, opening up future changes and potential.

Also added a ease-of-use builder, which runs the CMake commands for us based on how Conan was configured - completely not necessary, but can be used by choice.

Resolves #320


FiniteReality (approved):
Looks good on paper

conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+        "flac:shared":True,
+        "fmt:shared":True,
+        "Opus:shared":True,
+        "Ogg:shared":True,
+        "Vorbis:shared":True,
+        "glad:gl_profile":"core",
+        "glad:gl_version":4.0,
+        "glad:gles2_version":3.0,
+        "spdlog:header_only":True,
+        #"vulkan-loader:shared":True,
```
RubyNova, line 48: Why is this commented out?

capnkenny: It's not currently a part of `main` branch and causes issues if the library required isn't being included (see the Vulkan deps that are commented as well). I only added anything pertaining to Vulkan as we'll eventually need it (aka when you pull main into your graphics branch 🙂 )

RubyNova: What kind of issues? That sounds wonky, but thanks for the heads up.

capnkenny: Well, if I uncomment the option but not the dep, python gets upset because the option doesn't exist (as a key) to set the value. At least that's what I was seeing on my box.

(DW, uncommenting both option and dep made everything happy w/o issue)

RubyNova: Ohhh I see. Makes sense then. Must've missed the dependency comment.

RubyNova (approved):
LGTM bar clang format complaints.

PR #329, opened 2021-09-12
Switching doxygen workflow to use Windows

capnkenny:
Since `gh` commands are not in the container we use for Linux, I've temporarily switched the workflow to use Windows only so that we can begin to produce documentation for the NovelRT website - realistically as long as this works, we wouldn't need to _switch_ to Linux either, unless we just want a faster build time for this workflow.

Also, fixed an obscure bug that prevented the `$documentation` flag from being set properly in the Windows CI script🙂 as well as an odd issue where the doxygen directory was not set properly, so nothing would get committed anyways.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -35,7 +35,10 @@ function Generate() {
   if ($ci) {
   conan config install https://github.com/novelrt/ConanConfig.git
   conan install . -if "$BuildDir" --build=missing --profile windows-vs2019-amd64
-    $remaining = ,"-DNOVELRT_BUILD_DOCUMENTATION=OFF", "-DPython_FIND_REGISTRY=NEVER", "-DPython_FIND_STRATEGY=LOCATION" + $remaining
+      $remaining = ,"-DPython_FIND_REGISTRY=NEVER", "-DPython_FIND_STRATEGY=LOCATION" + $remaining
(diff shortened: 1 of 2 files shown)


scripts/build.ps1
```diff
@@ -35,7 +35,10 @@ function Generate() {
   if ($ci) {
   conan config install https://github.com/novelrt/ConanConfig.git
   conan install . -if "$BuildDir" --build=missing --profile windows-vs2019-amd64
```
RubyNova, line 37: Shouldn't this be `..`? I know its not part of this PR but its something I just spotted.

RubyNova (approved):
LGTM

scripts/build.ps1
capnkenny: Nope, since this is working in the root directory, it's pointing accurately to `conanfile.py` 😄

RubyNova: That's wild. How is that even building then?

capnkenny: We pipe in the appropriate build/repo paths when calling the respective functions (i.e `cmake -S $RepoRoot -B $BuildDir` to configure, `cmake --build $BuildDir` to build, etc) so that everything is where it should be, even though we're calling from the root directory

PR #335, opened 2021-09-23
Fixing Doxygen workflow with tested GH Action

capnkenny:
This now should run properly using our linux container.

- Made sure that Github SHA gets pulled in properly in the `gh pr` command as it did not like using environment variables
- Fixed the documentation path and variable that was being passed in

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -42,4 +42,4 @@ jobs:
       - name: Create Pull Request
         if: ${{ success() }}
         run: |
-          gh pr create --title "${{ GITHUB_SHA }} into gh-pages" --body "Updated documentation from commit ${{ GITHUB_SHA }} in main." --base gh-pages --head documentation/${{ GITHUB_SHA }}
+          gh pr create --title "${{ github.sha }} into gh-pages" --body "Updated documentation from commit ${{ github.sha }} in main." --base gh-pages --head documentation/${{ github.sha }}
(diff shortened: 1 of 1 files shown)


.github/workflows/doxygen.yml
```diff
@@ -11,36 +11,35 @@ jobs:
...
+        
+      - name: Prepare Secrets
+        run: echo "${{ secrets.GITHUB_TOKEN }}" >> $GITHUB_WORKSPACE/secret.txt
 
       #Mandatory step it appears otherwise pushing the commit will fail
       #Switched username and e-mail to something identifiable?
       - name: Configure Git
         run: |
-          gh auth login
+          gh auth login --with-token < $GITHUB_WORKSPACE/secret.txt
```
RubyNova, line 28: So how does this work?

capnkenny: We currently write the `GITHUB_TOKEN` secret that the workflow has to a file in the temp workspace created by the run - this way we aren't trying to pipe in a string (as `gh cli` _rees like no tomorrow when I do this_). 

Then, we pipe the token into the command following the CLI's instructions (as documented) - this is because `GITHUB_TOKEN` isn't available to our docker container, so it acts like it's a normal computer/VM instead of a specific GH-hosted runner. Because it's using the token from the Action, though, it's able to get the proper permissions.

Please note, we don't publish the `secret.txt` file, and in the runs it is censored as `***` so it isn't truly exposed afaik.

RubyNova (approved):
LGTM

PR #357, opened 2021-09-28
Added some ECS API sugar - new configurator type and a new place for us to define default systems when the time comes.

### LINKED ISSUE #351: Add ECS sugar for default systems.

RubyNova:
resolves #351


include/NovelRT/Maths/Utilities.h
```diff
@@ -5,12 +5,19 @@
...
 
 #include <NovelRT/Exceptions/InvalidOperationException.h>
+#include <array>
 
 namespace NovelRT::Maths
 {
     class Utilities
     {
     public:
+        [[nodiscard]] inline constexpr static std::array<uint8_t, 32> Log2DeBruijin() noexcept
```
tannergooding, line 15: Just calling out, before I forget, we'll want to add proper attribution here to both .NET Runtime and http://graphics.stanford.edu/~seander/bithacks.html

RubyNova:
Please note this PR now also fixes the race condition introduced with a previous fix.

include/NovelRT/Maths/Utilities.h
capnkenny: So uhh...can I get just a quick explanation on what this is?

include/NovelRT/Ecs/SystemScheduler.h
```diff
@@ -18,12 +18,6 @@ namespace NovelRT::Ecs
     class SystemScheduler
     {
     private:
-        struct QueueLockPair
-        {
-            std::optional<Atom> systemUpdateId;
-            std::mutex threadLock;
```
capnkenny, line 24: No more locking required?

capnkenny (commented):
Just need a few answers, tis all :)

src/NovelRT/Ecs/SystemScheduler.cpp
```diff
@@ -57,111 +90,112 @@ namespace NovelRT::Ecs
...
-                auto& pair = _threadWorkQueues[workerIndex];
-
-                if (systemIds.empty())
-                {
-                    break;
-                }
+            uint64_t workerIndex;
 
-                const std::lock_guard<std::mutex> lock(pair.threadLock);
+            while ((workerIndex = NovelRT::Maths::Utilities::LeadingZeroCount64(_threadAvailabilityMap)) == 64)
```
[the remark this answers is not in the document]
RubyNova: a `uint64_t` is 64 digits long - as such, if it returns 64, no workers are available.

include/NovelRT/Ecs/SystemScheduler.h
RubyNova: Yesn't.

We no longer lock in the traditional sense, but atomic operations might perform cheaper atomic locks on the XOR operations, or may just be volatile writes.

capnkenny: So are we hoping that we hoping that we never lock in this case (lock-free)? Or do we want to just defer to the CPU to determine this at runtime?

RubyNova: It's a case of the atomic lock would be so cheap on the XOR operations that it literally doesn't matter and you may as well consider this lockless due to the fact were using no mutexes at this point.

capnkenny:
Just need an explanation of what tanner pointed out _as well as updated attribution_ before I shall approve

include/NovelRT/Maths/Utilities.h
RubyNova: Seems I spelt Debruijn wrong here.

RubyNova:
In essence, the DeBruijn table provides a set of values for a table lookup which makes finding Log2 of the provided value faster as it takes less operations to perform.

In the future we'll be switching to the hardware intrinsic/asm instruction, but that is fiddly in C++, and so this software implementation was created.

PR #368, opened 2021-10-01
Use the force, Novel-chan...

### LINKED ISSUE #352: Doxygen workflow creates conflicts
Currently the doxygen workflow creates conflicts as it pulls from the `main` branch instead of doing a merge with main on the `gh-pages` branch

capnkenny:
Modifying the flow to allow for force-pushing - should only be done to gh-pages branch.
This is done because previously by committing a document commit to the gh-pages branch, I caused the branch to diverge which did not allow future merging and caused rebasing to fail.

Resolves #352

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Create documentation branch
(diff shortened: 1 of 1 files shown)


RubyNova (approved):
LGTM luke

Issue #343, opened 2021-09-25
Migrate to .NET 6 LTS from .NET 5 for the .NET, C# and Ink Support.


RubyNova:
We are currently tied to the .NET Core SDK for .NET 5. With the upcoming release of .NET 6, we need to update our CMakeLists to target .NET 6 instead. This means changing the minimum target version in our find script, at the very least. IT might mean additional CMake adjustments, but they are not known at this time.

PR #369, opened 2021-10-03
Migrate to .NET 6 - #343

Exadon:
Migrate to .NET 6 LTS from .NET 5 for the .NET, C# and Ink Support.


FiniteReality:
It looks like 6.0.100 doesn't exist yet; the latest version is 6.0.100-rc.1; so it will likely be some time until this will get merged properly.

RubyNova:
@FiniteReality looks like it all passed on windows here - based on the linux failure I'm guessing this is because we've yet to update our docker container on our end. Does the specified docker image we're inheriting from exist?

FiniteReality:
> @FiniteReality looks like it all passed on windows here - based on the linux failure I'm guessing this is because we've yet to update our docker container on our end. Does the specified docker image we're inheriting from exist?

:woman_shrugging: I'll have to check; but I think it *might* be better overall if we hold off until .NET 6 is released, rather than jumping over to it in the RC phase.

RubyNova:
So, I can see why you'd say that - however, if the CMake is already accepting the release version number (because CMake is *weird*) then I say we go with it, as ultimately we were on .NET 5 preview before we were on stable and it was fine, I think an RC is a safer bet than a preview vesion, and I'd rather not have this PR just lingering around assuming the dockerfile is valid.

README.md
```diff
@@ -32,7 +32,7 @@ Currently there are no binary distributions of the engine as of yet, and we are
 If you wish to attempt to build a basic visual novel with the existing C++ API, you must first install the following dependencies:
 
 ### Dependencies
-- .NET SDK 5.0.100
+- .NET SDK 6.0.100
```
RubyNova, line 35: I think we should change this to the RC 1 version as its the readme - I don't think 6.0.100 explicitly exists yet (even though CMake is like "yeah k I'll still work")

RubyNova (approved):
LGTM bar one minor thing that we can change or let through but its up for discussion. :)

PR #376, opened 2021-10-08
Graphics to ECS integration, part 2.

### LINKED ISSUE #56: OpenGL Instancing.

### LINKED ISSUE #96: Window size does not auto-adjust to 3840 x 1080 correctly
When the engine starts, window size auto-adjusts to main screen size whilst not adjusting correctly to the resolution.

Expected to correctly display at screen resolution or Window size to be limited to display resolution.

Window size adjusts to 32:9 aspect ratio without the display adjusting to 3840 x 1080 resolution.

To reproduce, run project on 32:9 monitor at 3840 x 1080 resolution. Testing on other monitor resolutions advised.

NovelRTBug

### LINKED ISSUE #198: OpenGL context invalidated on window close before resources
(Title is the fan theory at the moment)

Stuff like textures outlive the lifetime of the rendering/windowing service, so when the application closes we get a bunch of GLAD errors saying we tried to delete things on an invalid GL context

### LINKED ISSUE #361: Re-organise Maths headers.
The Maths headers (under include/NovelRT/Maths) are currently all self-contained. As per the new design philosophy, individual types should no longer be able to be included directly, but rather they should only be accessed via NovelRT.h or via their respective "namespace header", so Maths.h in this case. This also means moving all forward declarations and includes to this namespace header along with adding their respective preprocessor include guards and errors. You can see a good example of this implementation at include/NovelRT/Ecs/Ecs.h and the individual type headers.

RubyNova:
This PR introduces the default rendering system in the ECS for the common use case. Will also provide some basic switches and options, but these will be pretty barebone.

EDIT: Resolves #96 
EDIT: Resolves #361
EDIT: Resolves #56 as OpenGL is officially gone
EDIT: Resolves #198 as OpenGL is officially gone


RubyNova:
Resolves #96

include/NovelRT/Ecs/DefaultComponentTypes.h
```diff
@@ -0,0 +1,87 @@
...
+
+        inline EntityGraphComponent& operator+=(const EntityGraphComponent& other)
+        {
+            *this = other;
+            return *this;
+        }
+    };
+
+    /*
+    struct QuadEntityBlockComponent
```
FiniteReality, line 52: I'm assuming this was just missed, but we have git so we don't need commented out source code :stuck_out_tongue:

include/NovelRT/Ecs/Graphics/DefaultRenderingSystem.h
```diff
@@ -0,0 +1,133 @@
...
+        [[nodiscard]] Experimental::Threading::ConcurrentSharedPtr<VertexInfo> GetExistingVertexDataBasedOnName(
+            const std::string& vertexDataName);
+
+        [[nodiscard]] Experimental::Threading::ConcurrentSharedPtr<VertexInfo> GetExistingVertexDataBasedOnId(
+            Atom ecsId);
+
+        [[nodiscard]] Experimental::Threading::ConcurrentSharedPtr<GraphicsPipelineInfo> GetExistingPipelineInfoBasedOnId(Atom ecsId);
+
+        /*
+        [[nodiscard]] Experimental::Threading::FutureResult<ConstantBufferInfo> LoadConstantBufferDataIntoNewRegionRaw(
```
FiniteReality, line 101: Likewise w/ commented code

include/NovelRT/Utilities/Misc.h
```diff
@@ -11,7 +11,7 @@
 #if defined(NDEBUG)
 #define unused(x) (void)(x)
 #else
-#define unused(x) (void)(0)
+#define unused(x) (void)(x)
```
FiniteReality, line 14: So if we're using the same definition in both branches here, why do we need the branch in the first place?

samples/EcsPipeline/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+set(SOURCES
+        main.cpp
+        )
```
FiniteReality, line 3: Indentation, also on line 10 and 34.

samples/EcsPipeline/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    std::shared_ptr<NovelRT::Ecs::Graphics::DefaultRenderingSystem> renderingSystem =
+        scheduler.GetRegisteredIEcsSystemAs<NovelRT::Ecs::Graphics::DefaultRenderingSystem>();
+
+    NovelRT::Experimental::Threading::FutureResult<NovelRT::Ecs::Graphics::TextureInfo> textureFuture =
+        renderingSystem->GetOrLoadTexture("novel-chan");
+
+    renderingSystem->ForceVertexTextureFutureResolution();
+
+    // unused(renderingSystem->CreateSpriteEntityOutsideOfSystem(textureFuture.GetBackingConcurrentSharedPtr(),
```
FiniteReality, line 33: Probably best to remove this if it's in sample code and isn't used

samples/EcsPipeline/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+            {
+                TransformComponent newComponent{};
+                newComponent.rotationInEulerAngles = 20 * delta.getSecondsFloat();
+                newComponent.scale = NovelRT::Maths::GeoVector2F::zero();
+                transforms.PushComponentUpdateInstruction(entity, newComponent);
+            }
+        });
+
+    /*
+    std::random_device rd;                               // obtain a random number from hardware
```
FiniteReality, line 68: I'd also remove this code if it's in a sample and not being used.

samples/Simple/main.cpp
```diff
@@ -102,7 +104,7 @@ int main(int /*argc*/, char* /*argv*/[])
         NovelRT::Transform(NovelRT::Maths::GeoVector2F(1920 / 2, 1080 / 2), 2, NovelRT::Maths::GeoVector2F(762, 881));
 
     novelChanRect =
-        runner.getRenderer()->createImageRect(novelChanTransform, 3, (imagesDirPath / "novel-chan.png").string(),
+        runner.getRenderer()->createImageRect(novelChanTransform, 0, (imagesDirPath / "novel-chan.png").string(),
```
FiniteReality, line 107: spooky constant: what does 3 and 0 mean?

src/NovelRT/Atom.cpp
```diff
@@ -5,45 +5,86 @@
...
         return Atom(value);
     }
 
-    Atom Atom::getNextEntityId() noexcept
+    Atom Atom::GetNextEntityId() noexcept
     {
         static std::atomic_uintptr_t _nextEntityId(0);
         auto value = ++_nextEntityId;
+
+        if (value == std::numeric_limits<uintptr_t>::max())
```
FiniteReality, line 41: Surely if this branch is reached we have a problem anyway? uintptr_t is 2^64 on modern 64-bit builds. That's 18 quintillion. I'm certain any system we'd be running on would run out of memory far before we reached that number, as 18 quintillion equates to roughly 18 exabytes by itself.

src/NovelRT/Atom.cpp
```diff
@@ -5,45 +5,86 @@
...
+        return Atom(value);
+    }
+
+    Atom Atom::GetNextEcsPrimitiveInfoConfigurationId() noexcept
+    {
+        static std::atomic_uintptr_t _nextEcsPrimitiveInfoConfigurationId(0);
+        auto value = ++_nextEcsPrimitiveInfoConfigurationId;
+        return Atom(value);
+    }
+}
```
FiniteReality, line 90: It seems we lost the `// namespace NovelRT` comment here - is that intentional?

RubyNova: we shouldn't have those comments. If we do then its a merge issue.

src/NovelRT/Atom.cpp
RubyNova: Particle systems have entered the chat (in reality I have no idea and also didn't want to find out).

samples/Simple/main.cpp
RubyNova: It's the layer value. This sample isn't sticking around much longer anyways. I'm deleting the entire legacy pipeline once everything in ECS is working and in for the MVP.

samples/EcsPipeline/main.cpp
RubyNova: No idea what the final sample will look like yet for this. This is me just debugging different things at present.

include/NovelRT/Utilities/Misc.h
RubyNova: I have no idea, I don't even remember making this change.

src/NovelRT/Atom.cpp
FiniteReality: I imagine a *good* particle system would allocate a pool of entities to use for particle effects and then re-use them when available? Something like this really bad pseudocode:
```
std::queue<EntityId> pool;

void run_particle_system()
{
    if (pool.empty())
    {
        pool.push(new_particle());
    }
    else
    {
        auto entity = pool.front();
        do_stuff(entity);
        pool.pop();
    }
}

void do_stuff(EntityId entity)
{
    // magic
    pool.push(entity);
}
```

RubyNova: Yeah I know, and this branch will probably never ever get called, I just really didn't like leaving the possibility lurking around.

src/NovelRT/Atom.cpp
FiniteReality: I vaguely remember seeing these comments in other parts of this PR, hence my asking - this one seemed like the exception, rather than the rule.

RubyNova: I hate IDEs.

That is all.

include/NovelRT/Ecs/DefaultComponentTypes.h
```diff
@@ -0,0 +1,85 @@
...
+        Maths::GeoVector2F scale = Maths::GeoVector2F::one();
+        float rotationInEulerAngles = 0.0f;
+
+        inline TransformComponent& operator+=(const TransformComponent& other)
+        {
+            positionAndLayer += other.positionAndLayer;
+            rotationInEulerAngles += other.rotationInEulerAngles;
+            scale += other.scale;
+
+            if (rotationInEulerAngles > 359.999999f) // 360 == 0 pretty sure :]
```
tannergooding, line 25: I'd probably do this as `>= 360.0f`. `360.0` is exactly representable and will avoid a number of potential precision bugs
This also, notably, isn't accounting for a user passing in something that is say `759.0f`
And some users may expect this to be in radians instead; which also helps avoid precision problems and is a more natural to use representation for internal usage; even if users want to express things in degrees.

include/NovelRT/Ecs/DefaultComponentTypes.h
```diff
@@ -0,0 +1,85 @@
...
+            {
+                rotationInEulerAngles -= 359.999999f;
+            }
+            else if (rotationInEulerAngles < 0)
```
tannergooding, line 29: You might consider allowing this to be up to `-360.0f`. Even though the ultimate position is the "same", the "direction" is different and may impact interpolation or blur

include/NovelRT/Ecs/DefaultComponentTypes.h
RubyNova: If we do this in radians we will have to change the internal logic of NovelRT::Maths to use radians also. I did what a game developer would expect of a public API - which is euler angles. If we want to use radians everywhere that's a discussion for another GH issue I think.

I'm happy to change the value though. :)

include/NovelRT/Ecs/DefaultComponentTypes.h
RubyNova: To clarify - what would be the exact behaviour of  `-360`? Would it wrap around to `0` or go `+360`?

tannergooding: Just to elaborate a bit on this one; consider, for example, a clock face.

Winding forward by 30 degrees puts it at 1, while winding it backwards by 330 degrees puts it at the same position.

However, if the rotation is `30 degrees` and the time to transition is `10 s` then at 5s the hand will be halfway between 12 and 1. Where-as for `-330` degrees it should have been somewhere between the 6 and 7 markers.

include/NovelRT/Ecs/DefaultComponentTypes.h
tannergooding: I'm trying to say that there is a disconnect between what's good for internal usage vs what's good for user interface.

Users want to work with degrees; where-as math APIs and most optimizations here will work with `radians`.

RubyNova: There's certainly a disconnect in my brain after reading that - this component is public API. What are you wanting me to do here?

include/NovelRT/Ecs/Graphics/DefaultRenderingComponentTypes.h
```diff
@@ -0,0 +1,101 @@
...
+                   (markedForDeletion == other.markedForDeletion);
+        }
+
+        inline bool operator!=(const RenderComponent& other) const noexcept
+        {
+            return !(*this == other);
+        }
+    };
+
+    // TODO: I don't know how to use these given how GraphicsPrimitive currently works quite yet.
```
tannergooding, line 46: Just leaving a comment that I'm here in the review and should pick it back up from here; otherwise I'll forget.

Please ignore <3

include/NovelRT/Ecs/DefaultComponentTypes.h
tannergooding: Pardon the C#, but:
```csharp
private float _rotationInRadians;

public float RotationInDegrees => rad2deg(_rotationInRadians);

public float RotationInRadians => _rotationInRadians;
```

RubyNova: ....are we sure that won't break the trivially copyable requirement?

tannergooding: The term `Euler Angle` here is also confusing, because Euler angles can be specified in degrees or radians. You're specifically expressing `rotation in degrees`

RubyNova: I'm using Unity terminology, its what I know. You'll have to bear with me.

tannergooding: > ....are we sure that won't break the trivially copyable requirement?

Why would it? You're exposing a getter that does simple arithmetic on value and one that just returns the underlying value,

RubyNova: I don't know - its C++ :^)

tannergooding: That is, `is_trivially_copyable` doesn't have anything to do whether or not arbitrary members are exposed. It only impacts things that have move/copy constructors and move/copy assignment operators.

RubyNova: Gotcha.

tannergooding: https://en.cppreference.com/w/cpp/types/is_trivially_copyable

thaumanovic:
LGTM BOSS

include/NovelRT/Ecs/DefaultComponentTypes.h
RubyNova: @tannergooding I'll need you to double check I've implemented this one correctly before marking as resolved.

include/NovelRT/Ecs/DefaultComponentTypes.h
RubyNova: I've not implemented the helper members just yet, but wanted to flag that I've changed this functionally to use radians and so I need a second look at it @tannergooding

RubyNova:
This now also resolves #361 because literally nobody else was picking it up and it was bothering me. :(

README.md
```diff
@@ -36,15 +36,18 @@ If you wish to attempt to build a basic visual novel with the existing C++ API,
...
+- GLFW 3.3.2
 - glm
 - gtest/gmock 1.10.0
 - libpng 1.6.34
 - libsndfile 1.0.28
-- OpenAL 1.19.1
-- spdlog 1.4.2
 - Microsoft GSL 3.1.0
+- OneTBB 2021.3.0
+- OpenAL 1.21.1
```
tannergooding, line 46: Looks like the list is otherwise alphabetical
```suggestion
- OpenAL 1.21.1
- OneTBB 2021.3.0
```

tannergooding: Its `One`, not `Open`. I cannot read today :)

include/NovelRT/Experimental/Graphics/Graphics.h
```diff
@@ -64,14 +69,13 @@ namespace NovelRT::Experimental::Graphics
     enum class GraphicsBufferKind : uint32_t;
     enum class TexelFormat : uint32_t;
     class GraphicsSurfaceContext;
+    class GraphicsResourceManager;
```
tannergooding, line 72: nit: Might be nice to group enums vs classes vs structs together?

include/NovelRT/Experimental/Graphics/GraphicsContext.h
```diff
@@ -32,7 +32,13 @@ namespace NovelRT::Experimental::Graphics
...
         virtual void Copy(std::shared_ptr<GraphicsBuffer> destination, std::shared_ptr<GraphicsBuffer> source) = 0;
         virtual void Copy(std::shared_ptr<GraphicsTexture> destination, std::shared_ptr<GraphicsBuffer> source) = 0;
-        virtual void Draw(std::shared_ptr<GraphicsPrimitive> primitive) = 0;
+
+        inline void Draw(std::shared_ptr<GraphicsPrimitive> primitive)
+        {
+            Draw(std::move(primitive), 1);
+        }
+
+        virtual void Draw(std::shared_ptr<GraphicsPrimitive> primitive, int32_t instanceCount) = 0;
```
tannergooding, line 41: signed instance count?

include/NovelRT/Experimental/Graphics/GraphicsMemoryAllocator.h
```diff
@@ -27,9 +27,7 @@ namespace NovelRT::Experimental::Graphics
                                                          ? _settings.MaximumBlockCountPerCollection
                                                          : std::numeric_limits<int32_t>::max();
 
-            size_t maximumSharedBlockSize = _settings.MaximumSharedBlockSize.has_value()
-                                                ? _settings.MaximumSharedBlockSize.value()
-                                                : 256 * 1024 * 1024;
+            size_t maximumSharedBlockSize = _settings.MaximumSharedBlockSize.value_or(256 * 1024 * 1024);
```
tannergooding, line 30: nit: `GetValueOrDefault`? `value_or` is not "intuitive" IMO.

tannergooding: oh, this is part of std::optional; rip

include/NovelRT/Experimental/Graphics/GraphicsPipelineBlendFactor.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef NOVELRT_EXPERIMENTAL_GRAPHICS_GRAPHICSPIPELINEBLENDFACTOR_H
+#define NOVELRT_EXPERIMENTAL_GRAPHICS_GRAPHICSPIPELINEBLENDFACTOR_H
+
+#ifndef NOVELRT_EXPERIMENTAL_GRAPHICS_H
+#error NovelRT does not support including types explicitly by default. Please include Graphics.h instead for the Graphics namespace subset.
+#endif
+
+namespace NovelRT::Experimental::Graphics
+{
+    enum class GraphicsPipelineBlendFactor : uint32_t
```
tannergooding, line 13: nit: It looks like you just copied `VkBlendFactor` (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBlendFactor.html). This is going to be non-portable if we add `D3D12` later and it might risk introducing bugs if we just assume blittable casts between our enum and Vks enum

include/NovelRT/Experimental/Graphics/GraphicsResourceManager.h
```diff
@@ -0,0 +1,86 @@
...
+        GraphicsResourceManager(GraphicsResourceManager&& other) noexcept;
+
+        GraphicsResourceManager& operator=(const GraphicsResourceManager& other);
+        GraphicsResourceManager& operator=(GraphicsResourceManager&& other) noexcept;
+
+        ~GraphicsResourceManager() = default;
+
+        template<typename TData>
+        [[nodiscard]] GraphicsMemoryRegion<GraphicsResource> LoadVertexData(gsl::span<TData> data,
+                                                                            size_t alignment = 16)
```
tannergooding, line 45: nit: Use `0` and treat `0` as default so we can change alignment without breaking existing compiled code

tannergooding (approved):
Generally LGTM. Left a few nits

samples/Simple/main.cpp
RubyNova: File has been deleted. Resolving.

include/NovelRT/Experimental/Graphics/GraphicsPipelineBlendFactor.h
RubyNova: I have opened an issue, #391 , to address this. For now I am resolving as we do not have D3D12 currently.

PR #382, opened 2021-11-01
Add oneTBB and tbb::mutex to SystemScheduler.

RubyNova:
This should resolve the CPU maxing out we've been seeing.


CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ set(NOVELRT_VORBIS_VERSION "1.3.7" CACHE STRING "Opus version")
 set(NOVELRT_GSL_VERSION "3.1.0" CACHE STRING "ms-gsl version")
 set(NOVELRT_VULKAN_VERSION "1.2.172" CACHE STRING "Vulkan version")
 set(NOVELRT_VULKAN_MEMORY_ALLOCATOR_VERSION "2.3.0" CACHE STRING "VulkanMemoryAllocator version")
+set(NOVELRT_ONETBB_VERSION "2021.3.0" CACHE STRING "OneTBB version")
```
tannergooding, line 53: nit: The rest of these were alphabetical. Looks like `GSL` and `ONETBB` are the only out of order ones

RubyNova: Oops. Will fix. I've just kind of been tacking them on as I go.

conanfile.py
```diff
@@ -19,7 +19,8 @@ class NovelRTConan(ConanFile):
         ("spdlog/1.8.2"),
         ("ms-gsl/3.1.0"),
         ("vulkan-loader/1.2.182"),
-        ("vulkan-memory-allocator/2.3.0")
+        ("vulkan-memory-allocator/2.3.0"),
+        ("onetbb/2021.3.0")
```
tannergooding, line 23: same here, these look alphabetical except for `gsl` and `onetbb`

include/NovelRT/Ecs/SystemScheduler.h
```diff
@@ -32,6 +32,7 @@ namespace NovelRT::Ecs
 
         std::vector<Atom> _threadWorkItem;
         std::vector<std::thread> _threadCache;
+        std::vector<std::unique_ptr<tbb::mutex>> _mutexCache;
```
FiniteReality, line 35: IMO we should probably combine this with the threadAvailabilityMap TBH, something like this:

```cpp
class worker
{
    tbb::mutex sleep_blocker;
    std::atomic_bool _available;
};

std::vector<worker> _workers;
```

It may not be as efficient as the current solution, but it means we're not splitting the same information across multiple types, and then finding an available thread is effectively this:
```
for (auto& system : systems)
{
    auto available = [](const worker& worker){ return worker._available.load(/* some appropriate memory order */); }
    std::partition(_workers.begin(), _workers.end(), available);
    auto& worker = std::find_if(_workers.begin(), _workers.end(), worker);
    if (worker != _workers.end())
    {
        worker.workAvailable(system);
    }
}
```

FiniteReality (approved):
Other than the one suggestion I made earlier, I'm alright with this

PR #385, opened 2021-11-20
Improve cross-target dependency handling

FiniteReality:
(I hate CMake so much)

This improves cross-target dependency handling, much like #214 was supposed to. Unlike that effort, which died because CMake sucks, this one actually somewhat succeeded.

Ultimately, this should fix the incremental build issues we're seeing on Windows.


src/NovelRT/CMakeLists.txt
```diff
@@ -202,9 +199,11 @@ if(WIN32)
...
 
-  if(NOVELRT_INCLUDE_CONAN_PATHS)
-    add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_directory
+  # I hate CMake I hate it I hate it I hate it
+  # (work around for CMAKE REFUSING TO COPY NECESSARY FILES FOR DEBUGGING MAKING IT IMPOSSIBLE TO DEBUG)
+  add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_directory
       ${CMAKE_BINARY_DIR}/thirdparty
       $<TARGET_FILE_DIR:Engine>)
-  endif()
+  file(GLOB libs RELATIVE ${CMAKE_BINARY_DIR}/thirdparty ${CMAKE_BINARY_DIR}/thirdparty/*.dll)
```
FiniteReality, line 207: So normally this would be bad, but because CMake hates everybody who touches it, I gave up and just did the easy thing here. We're getting the contents of the `thirdparty` folder setup by Conan, and setting that as a list of resource files. That's what the whole copy_build_products system uses to copy additional "non-primary" outputs.

src/NovelRT.DotNet/CMakeLists.txt
```diff
@@ -115,22 +104,24 @@ elseif(UNIX)
...
 endif()
 
-add_library(CoreCLR::nethost SHARED IMPORTED GLOBAL)
-set_target_properties(CoreCLR::nethost
+add_library(Dotnet SHARED IMPORTED GLOBAL)
+add_dependencies(Dotnet Dotnet_Build)
+set_target_properties(Dotnet
   PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}"
   IMPORTED_LOCATION "${CMAKE_CURRENT_BINARY_DIR}/${nethost_lib}"
```
FiniteReality, line 112: Since we're now using the `RESOURCE` property to share additional build outputs, we can set the primary output file here, which is the .NET host library which is needed to link against, and also communicate the additional files necessary (in this case, the .NET wrapper libraries)

cmake/CopyBuildProducts.cmake
```diff
@@ -0,0 +1,97 @@
+function(copy_build_products target)
+  set(valid_options "DEPENDENCY;DEPENDENCY_FILES;TARGET_LOCATION;CONDITION")
+  foreach(arg IN LISTS ARGN)
```
FiniteReality, line 3: Had to write my own argument parsing here because CMake ~~sucks~~ typically doesn't like you specifying args in this manner when you use `parse_arguments` or whatever command it is again. Specifying args in this manner is more convenient for us, even if the error handling is incredibly primitive (read: nonexistant) and will likely cause strange errors if used incorrectly.

cmake/CopyBuildProducts.cmake
```diff
@@ -0,0 +1,97 @@
...
+          list(APPEND resx "${res}")
+        endif()
+      endforeach()
+    endif()
+
+    set(cond_succeeded NO)
+    if(NOT DEFINED "COPY_BUILD_PRODUCTS_${dependency_idx}_CONDITION")
+      set(cond_succeeded YES)
+    else()
+      cmake_language(EVAL CODE "if(COPY_BUILD_PRODUCTS_${dependency_idx}_CONDITION)\nset(cond_succeeded YES)\nendif()")
```
FiniteReality, line 48: I have no idea if this can be done in a better way. I tried multiple times and I kept getting the wrong result.

Chances are, it's probably still wrong but it Worked On My Machine:tm:.

cmake/CopyBuildProducts.cmake
```diff
@@ -0,0 +1,97 @@
...
+      endforeach()
+    endif()
+  endforeach()
+
+  list(JOIN copy_snippet "\n" copy_snippet)
+
+  file(GENERATE
+    OUTPUT copy_build_products.cmake
+    CONTENT "${copy_snippet}")
+  add_custom_target("copy_build_products_to_${target}" ALL
```
FiniteReality, line 85: Custom targets are always considered to be out-of-date, so this will always run. Fortunately, we're using `copy_if_different`, so this'll cause the least amount of updates and will (hopefully) be very fast.

cmake/CopyBuildProducts.cmake
```diff
@@ -0,0 +1,97 @@
...
+    COMMAND ${CMAKE_COMMAND} -P copy_build_products.cmake
+    COMMENT "Copying required build products to ${target}")
+  add_dependencies(${target} copy_build_products_to_${target})
+  add_dependencies(copy_build_products_to_${target} ${COPY_BUILD_PRODUCTS_DEPENDENCIES})
+  set_property(TARGET copy_build_products_to_${target} APPEND PROPERTY
+    ADDITIONAL_CLEAN_FILES "${COPY_BUILD_PRODUCTS_COPIED_RESOURCES}"
+  )
+  set_property(
+    TARGET ${target} APPEND PROPERTY
+    RESOURCE "${COPY_BUILD_PRODUCTS_COPIED_RESOURCES}"
```
FiniteReality, line 95: Honestly, kind of a pipe dream at this point: I was hoping this would Just Work:tm:, but because CMake, it obviously didn't.

resources/CMakeLists.txt
```diff
@@ -19,12 +19,12 @@ set(RESOURCES_FILES
...
-  configure_file(${resource} Resources/${resource} COPYONLY)
+  configure_file(${resource} ${resource} COPYONLY)
 endforeach()
 
-add_executable(Resources IMPORTED GLOBAL)
+add_library(Resources INTERFACE IMPORTED GLOBAL)
 set_target_properties(Resources
   PROPERTIES
-    IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/Resources/IF_YOU_SEE_THIS_YOU_TRIED_TO_RUN_THE_RESOURCES_TARGET
+    RESOURCE "${RESOURCES_FILES}"
```
FiniteReality, line 28: A neat thing about `RESOURCE` is that if you don't have any "linkable" outputs, you don't need to specify one. Will this break future consumers? Probably. Do I care right now? Nope.

src/NovelRT.DotNet/NovelRT.DotNet.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <None Update="NovelRT.DotNet.runtimeconfig.json" CopyToOutputDirectory="PreserveNewest" />
+    <None Include="NovelRT.DotNet.runtimeconfig.json" CopyToOutputDirectory="PreserveNewest" CopyToPublishDirectory="PreserveNewest" />
```
FiniteReality, line 10: So this is probably(?) why the runtimeconfig wasn't properly being copied from the MSBuild side. I thought it'd get picked up by DefaultNoneItems or whatever the property is in the .NET SDK but apparently not :shrug:

FiniteReality:
CMake really just had to have the last laugh, huh.

cmake/CopyBuildProducts.cmake
```diff
@@ -0,0 +1,98 @@
...
+)")
+      endforeach()
+    endif()
+  endforeach()
+
+  list(JOIN copy_snippet "\n" copy_snippet)
+
+  get_property(multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+  file(GENERATE
+    OUTPUT copy_build_products$<$<BOOL:${multi_config}>:_$<CONFIG>>.cmake
```
FiniteReality, line 84: This `$<$<BOOL:${multi_config}>:_$<CONFIG>>` nonsense evaluates to `_$<CONFIG>` if the build-system generator supports multiple configurations, avoiding a CMake error where we try to overwrite the same file multiple times with different content. This means that we end up generating multiple files, one of which is selected based on the configuration. For single-configuration generators, this isn't necessary, so the `_$<CONFIG>` suffix is dropped entirely.

cmake/CopyBuildProducts.cmake
```diff
@@ -0,0 +1,98 @@
...
+  ${tgt_dir}/${file}
+  ${COPY_BUILD_PRODUCTS_${dependency_idx}_TARGET_LOCATION}/${file}
+)")
+      endforeach()
+    endif()
+  endforeach()
+
+  list(JOIN copy_snippet "\n" copy_snippet)
+
+  get_property(multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
```
FiniteReality, line 82: Apparently some CMake things are properties at the global level, versus being variables at the global level. Thanks, CMake.

src/NovelRT/CMakeLists.txt
RubyNova: This will break Vulkan. Please include all config extensions as well.

capnkenny: So at the moment, this does not break on my machine with Vulkan SDK installed based on this branch. Do you mean for the future? We currently tell conan to copy `*.json` files, but it doesn't report that it copies any during configuration AFAIK (and there appears to be no dependency on them based on this branch or main).

capnkenny: Did some digging and spoke to @RubyNova - previously we used a `vulkan-validationlayers` conan package that produced JSON files at install, but we no longer use that package in mainline.

RubyNova: OK, then disregard what I said previously. This is resolved.

RubyNova (changes requested):
LGTM but we need to update CMake minimum version and the README to reflect the correct version. AFAIK we are still set to CMake 3.13, which is, uh, wrong.

FiniteReality:
> LGTM but we need to update CMake minimum version and the README to reflect the correct version. AFAIK we are still set to CMake 3.13, which is, uh, wrong.
> 
> EDIT: Could we also make the file handling function names contain "NovelRT" or "novelrt" or "NOVELRT" or whatever is more appropriate as a prefix? I want to avoid CMake clashes with other dependencies if possible and this seems like a good way to do it.

I bumped the minimum CMake version to 3.18, but I honestly don't think that we'll have any clashes because as far as I can tell, CMake's policy is to require projects to write this stuff themselves rather than do anything in any sort of sane manner.

RubyNova (approved):
LGTM

FiniteReality:
This just in: IT WORKED! We did it, GitHub!

:tada:

PR #386, opened 2021-12-04
Automagically build linux docker images

FiniteReality:
CC @RubyNova

This should automagically build Linux docker images when the Dockerfile in `main` changes. Which *should* keep it up to date. Hopefully.

I need a few secrets setup though, like the login info for Docker Hub so that we can log in. Alternatively, we could use the GitHub container registry instead of Docker Hub: https://github.com/marketplace/actions/docker-login#github-container-registry

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -1,0 +1,26 @@
+name: Build Docker Images
+
+on:
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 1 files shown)


RubyNova (commented):
Will this try to run on both debug and release?

FiniteReality:
> Will this try to run on both debug and release?

It'll only run when the dockerfile changes.

RubyNova (approved):
LGTM

PR #389, opened 2022-01-18
Improved C error handling

jeuxjeux20:
The current error handling system is a bit error prone, and error messages (`Nrt_setXErrMsg`) were sometimes forgotten.

As a quick fix before an entire error handling rework, there's now `Nrt_getXErrorInternal` methods that give an `NrtResult` while also setting the error message for it.

This makes maintaining errors easier and also shorter to write.


RubyNova (commented):
To clarify it looks like this PR only covers existing cases where we've set error messages right? Are we not missing any? I thought we were.

PR #399, opened 2022-02-02
Added Interop building flag

### LINKED ISSUE #398: Add flagging for building Interop libraries
At the moment, Interop is on pause until after the MVP is released. With that said, when building via Conan there is no way to stop the Interop sources from being build.

We should probably add a `NOVELRT_BUILD_INTEROP` CMake definition that defaults to OFF (ON after MVP when we finish it) - this will make it configurable and can also allow Conan to skip this step.

capnkenny:
Resolves #398 

Adds an interop building flag that's defaulted to `ON`. By setting this to off, effectively only the NovelRT engine core will build and link. (When Tests, etc are enabled they will be affected by this too).


src/CMakeLists.txt
```diff
@@ -1,2 +1,4 @@
 add_subdirectory(NovelRT)
-add_subdirectory(NovelRT.Interop)
+if (NOVELRT_BUILD_INTEROP)
+    add_subdirectory(NovelRT.Interop)
+endif()
```
RubyNova, line 4: Missing EOF.

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ find_package(GTest ${NOVELRT_GTEST_VERSION} REQUIRED
   COMPONENTS gtest_main
 )
 
+set(INTEROPBUILD 0)
```
RubyNova, line 5: Wat?

RubyNova (changes requested):
1 minor nit and 1 zubat-powered confuse ray.

tests/NovelRT.Tests/CMakeLists.txt
FiniteReality: Just use NOVELRT_BUILD_INTEROP? `$<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop>`

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -51,24 +36,54 @@ set(SOURCES
...
+endif()
 
 add_executable(Engine_Tests ${SOURCES})
+
+if(NOVELRT_BUILD_INTEROP)
+    copy_build_products(Engine_Tests
+      DEPENDENCY Engine
+      TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
+
+      DEPENDENCY Interop
```
FiniteReality, line 67: This should be a condition (`CONDITION TARGET Interop`) but right now it will fail. Remind me to fix it later.

tests/NovelRT.Tests/CMakeLists.txt
capnkenny: Yup - switched this since I didn't know about the BOOL treatment here

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -51,24 +34,53 @@ set(SOURCES
   Utilities/EventTest.cpp
 
   main.cpp
- )
+)
+
+if(NOVELRT_BUILD_INTEROP)
```
FiniteReality, line 39: Instead of using an if here, you should use generator expressions:
```cmake
set(sources
  ...
  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:the_file.cpp>
)
```

capnkenny: OOOOOOOoooo I did not know this was a thing.

tests/NovelRT.Tests/CMakeLists.txt
capnkenny: Created issue as a reminder

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -18,22 +18,22 @@ set(SOURCES
...
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Ecs/NrtSystemSchedulerTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/GeoBoundsTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/GeoMatrix4x4Test.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/GeoVector2FTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/GeoVector3FTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/GeoVector4FTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/QuadTreePointTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Maths/QuadTreeTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/SceneGraph/NovelRTSceneNodeTest.cpp>
+  $<$<BOOL:${NOVELRT_BUILD_INTEROP}>:Interop/Timing/NovelRTTimestampTest.cpp>
```
RubyNova, line 36: I'm wondering if there's a better way to do this. It's not a blocker, but maybe just like, "if interop, append these files" type deal? Not sure how doable or nice it would look, just musing.

RubyNova (approved):
LGTM

tests/NovelRT.Tests/CMakeLists.txt
capnkenny: That was essentially what my previous change had no?... It requires a separate `if` block, whereas this is inlined with the same block the rest of the test sources use (based on @FiniteReality's suggestion).

PR #404, opened 2022-02-05
Resolves window resizing and minimising crashes.

### LINKED ISSUE #401: Resizing or minimizing/maximizing window causes crash
Messing around with the InputEcsSample project; it seems resizing the window in any way (dragging, minimizing, maximizing) causes the engine to crash.

Debugging causes the app to break here:
https://github.com/novelrt/NovelRT/blob/267af6543c2590d7e1e81c5df83b167bdd6c934b/src/NovelRT/Graphics/Vulkan/VulkanGraphicsDevice.cpp#L500

RubyNova:
Resolves #401


src/NovelRT/Graphics/Vulkan/VulkanGraphicsDevice.cpp
```diff
@@ -345,10 +351,14 @@ namespace NovelRT::Graphics::Vulkan
             throw std::runtime_error("Failed to acquire next VkImage! Reason: " +
                                      std::to_string(acquireNextImageResult));
         }
-        _contextIndex = contextIndex;
 
-        presentCompletionGraphicsFence->Wait();
-        presentCompletionGraphicsFence->Reset();
+        if (!_isAttachedToResizeEvent)
```
tannergooding, line 355: Why is this needed?

RubyNova: It's so it can attach on first run. It's either here or on construction, which means making a ton of stuff non-lazy, which means a descent into C++ madness IIRC.

RubyNova: I don't remember the specifics but I remember it being problematic in some fashion with some lazies.

tannergooding (approved):

PR #411, opened 2022-02-14
Adding CMake Install target for Engine

capnkenny:


src/NovelRT/CMakeLists.txt
```diff
@@ -161,11 +161,38 @@ if(WIN32)
       Winmm
   )
 
+  set(CMAKE_INSTALL_PREFIX "C:/Program Files/NovelRT" CACHE PATH "..." FORCE)
```
FiniteReality, line 164: No.

RubyNova: Please provide an alternative @FiniteReality. "No." doesn't help anything other than you don't like it.

src/NovelRT/CMakeLists.txt
```diff
@@ -161,11 +161,38 @@ if(WIN32)
...
+  PUBLIC_HEADER DESTINATION include
+)
+install(
+  DIRECTORY "${CMAKE_SOURCE_DIR}/include/NovelRT"
+  DESTINATION "include"
+  FILES_MATCHING
+  PATTERN "*.h"
+)
+install(
+  DIRECTORY "${CMAKE_SOURCE_DIR}/Resources/Shaders"
```
FiniteReality, line 194: This should be in the resources target

src/NovelRT/CMakeLists.txt
```diff
@@ -161,11 +161,38 @@ if(WIN32)
...
+  TARGETS Engine 
+  EXPORT novelrt
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  RUNTIME DESTINATION bin
+  RESOURCE DESTINATION $<IF:$<BOOL:WIN32>,bin,lib>
+  PUBLIC_HEADER DESTINATION include
+)
+install(
+  DIRECTORY "${CMAKE_SOURCE_DIR}/include/NovelRT"
```
FiniteReality, line 188: This isn't necessary. Use PUBLIC_HEADER to specify the public headers.

src/NovelRT/CMakeLists.txt
```diff
@@ -161,11 +161,38 @@ if(WIN32)
...
+
+set_target_properties(Engine PROPERTIES PUBLIC_HEADER "${CMAKE_SOURCE_DIR}/include/NovelRT.h")
+
+install(
+  TARGETS Engine 
+  EXPORT novelrt
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  RUNTIME DESTINATION bin
+  RESOURCE DESTINATION $<IF:$<BOOL:WIN32>,bin,lib>
```
FiniteReality, line 184: This should probably go to "resources" or something. If you're looking to install the libraries, we should probably use RUNTIME_DEPENDENCIES, honestly

FiniteReality (changes requested):

src/NovelRT/CMakeLists.txt
FiniteReality: Setting `CMAKE_*` variables is a *massively* bad idea. It should be set by the CI infrastructure or the user, and *never ever* overridden.

RubyNova: I believe the problem is that its defaulting to Program Files (x86). We can't have that. This is a 64 bit software.

FiniteReality: Then tell users to specify `-DCMAKE_INSTALL_PREFIX=path/to/a/better/location`? Programs cannot install there anyway without admin permissions. Our build script should not assume you're running as an admin.

capnkenny: This is the feedback I was looking for as I wasn't sure we'd ever really want to install in the Program Files folders like CMake wants to do by default...

@RubyNova shall we just make sure to specify this in the readme?

src/NovelRT/CMakeLists.txt
capnkenny: WhenI attempted this, it only copied the files themselves without maintaining the folder structure. Surely that would break our headers, no?

src/NovelRT/CMakeLists.txt
capnkenny: Makes sense

src/NovelRT/CMakeLists.txt
```diff
@@ -161,11 +161,38 @@ if(WIN32)
...
   # I hate CMake I hate it I hate it I hate it
   # (work around for CMAKE REFUSING TO COPY NECESSARY FILES FOR DEBUGGING MAKING IT IMPOSSIBLE TO DEBUG)
-  add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_directory
-      ${CMAKE_BINARY_DIR}/thirdparty
-      $<TARGET_FILE_DIR:Engine>)
-  file(GLOB libs RELATIVE ${CMAKE_BINARY_DIR}/thirdparty ${CMAKE_BINARY_DIR}/thirdparty/*.dll)
+  #add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_directory
+  #    ${CMAKE_BINARY_DIR}/thirdparty
+  #    $<TARGET_FILE_DIR:Engine>)
+  #file(GLOB libs RELATIVE ${CMAKE_BINARY_DIR}/thirdparty ${CMAKE_BINARY_DIR}/thirdparty/*.dll)
```
capnkenny, line 171: Note to self - this breaks debugging on Windows...
@FiniteReality can we even remove this line? Or would this be subject to rework using RUNTIME_DEPENDENCIES like you mentioned? (Mainly because Linux does not appear to be an issue, only Windows)

src/NovelRT/CMakeLists.txt
RubyNova: I suppose so? I have no idea.

src/NovelRT/CMakeLists.txt
FiniteReality: I think using RUNTIME_DEPENDENCIES should fix it? Looking closer, it seems like that's a 3.21 feature though so we might not be able to support it.

src/NovelRT/CMakeLists.txt
RubyNova: Wait what? Debugging on windows is broken?

src/NovelRT/CMakeLists.txt
FiniteReality: It shouldn't do... Fabulist seems to have no problem with it

FiniteReality: Specifically, this is what's installed currently:
Screenshot_20220214_192131
And the headers are specified like:
```cmake
set(public_headers
    include/fabulist/compiler/compiler.hpp
)
set_target_properties(compiler
    PROPERTIES
        OUTPUT_NAME fabulist-compiler
        PUBLIC_HEADER "${public_headers}"
)
```

FiniteReality: Oh, excuse me I forgot to check the install() command:

```cmake
install(
    TARGETS compiler
    EXPORT fabulist
    LIBRARY DESTINATION lib
    PUBLIC_HEADER DESTINATION include/fabulist/compiler
)
```

:upside_down_face:

FiniteReality: https://gitlab.kitware.com/cmake/cmake/-/issues/7941
https://gitlab.kitware.com/cmake/cmake/-/issues/16739

Thanks, CMake. It's been a known issue since 2008. :unamused:

FiniteReality: Maybe it's time for that one insanely crazy idea I had to refactor the engine into separate modules just so that we're not fighting CMake at every possible moment.

src/NovelRT/CMakeLists.txt
capnkenny: I just saw an article about that - we'd have to bump the version up but it may work...not sure if that would be a bad thing or not 🤔

src/NovelRT/CMakeLists.txt
capnkenny: > Wait what? Debugging on windows is broken?

Was this a misplaced question?

src/NovelRT/CMakeLists.txt
capnkenny: @RubyNova - assuming you meant to comment here about debugging being broken, it currently uses the lines that are commented out (as it was called a hack lol). When commented, debugging breaks because the DLLs aren't copied to the engine path anymore

capnkenny:
Closing as I'm reprioritizing myself to finishing the Conan work for now - we can cross this bridge when it's time :)

PR #414, opened 2022-02-21
RBGAColour Refactor

### LINKED ISSUE #318: Refactor NrtRGBAConfigHandle
Currently, the C interop way of manipulating `RGBAConfig` values is by using the `Nrt_RGBAConfigHandle_*` methods. This is quite uncomfortable to use as it requires a heap-allocated value (`new`/`malloc`) for every new created colour; requires the C API consumer to use the `getR` ,`getB`, etc. methods for accessing each value; and is finally source of confusion, see #315 and #317.  

Because this is a relatively simple value, the `NrtRGBAConfigHandle` can probably be converted to a C struct.

SmylEFace:
this PR changes RGBAColour to a POD type. Resolves #318


.vs/tasks.vs.json
```diff
@@ -19,10 +19,15 @@
...
+      "type": "launch"
+    },
     {
       "taskLabel": "task-conanfile",
       "appliesTo": "conanfile.py",
       "type": "launch"
     }
   ]
-}
+}
```
RubyNova, line 33: We're missing an EOF here. Could you add that? :D

src/NovelRT.Interop/Graphics/NrtRGBAColour.cpp
```diff
@@ -14,132 +14,30 @@ extern "C"
...
-        RGBAColour* colourConfigPtr = reinterpret_cast<RGBAColour*>(colourConfig);
-        colourConfigPtr->setA(inputValue);
-
-        return NRT_SUCCESS;
-    }
-
-    float Nrt_RGBAColour_getRScalar(NrtRGBAColourHandle colourConfig)
-    {
-        RGBAColour* colourConfigPtr = reinterpret_cast<RGBAColour*>(colourConfig);
+        RGBAColour* colourConfigPtr = reinterpret_cast<RGBAColour*>(&colourConfig);
```
RubyNova, line 19: I think this might be more performant if we dereferenced the result of `reinterpret_cast` and used it as a value.

src/NovelRT.Interop/Graphics/NrtRGBAColour.cpp
```diff
@@ -14,132 +14,30 @@ extern "C"
...
         return colourConfigPtr->getBScalar();
     }
 
-    float Nrt_RGBAColour_getAScalar(NrtRGBAColourHandle colourConfig)
+    float Nrt_RGBAColour_getAScalar(NrtRGBAColour colourConfig)
     {
-        RGBAColour* colourConfigPtr = reinterpret_cast<RGBAColour*>(colourConfig);
+        RGBAColour* colourConfigPtr = reinterpret_cast<RGBAColour*>(&colourConfig);
         return colourConfigPtr->getAScalar();
     }
```
RubyNova, line 39: Same for these three, I think this might be more performant if we dereferenced the result of `reinterpret_cast` and used it as a value.

RubyNova (changes requested):
Looks good bar a couple nits on my end.

src/NovelRT.Interop/Graphics/NrtRGBAColour.cpp
jeuxjeux20: I agree, there's a lot of (de)referencing in there.

include/NovelRT.Interop/Graphics/NrtGraphicsTypedefs.h
```diff
@@ -0,0 +1,7 @@
+typedef struct
+{
+    int32_t r;
+    int32_t g;
+    int32_t b;
+    int32_t a;
+} NrtRGBAColour;
```
tannergooding, line 7: Why are we using a `128 bits per pixel/32 bits per component` format while the functions are normalizing assuming a `32 bits per pixel/8 bits per component` format?

I'd expect this is either using `uint8_t x;` or `uint32 x : 8;` for each component (I believe the latter is more efficient from an ABI perspective).

tannergooding: If this is intentionally 32-bits per component, then I'd largely expect the full range gets utilized as this would correspond to `VK_FORMAT_R32G32B32A32_SINT` (but even then, I'd probably expect `VK_FORMAT_R32G32B32A32_UINT` is better and these should be unsigned).

But, I'd also expect wanting 32-bits per component for integer formats is generally undesirable. For such a format, `float` is more typical.

RubyNova: We only currently use `RGBAColour` for the background clear, which uses the scalar methods which are 32 bit floats for each value in the `VkClearValue` object.

The size change should be fine. :)

PR #415, opened 2022-02-22
New custom render event.

RubyNova:
CC @tannergooding


include/NovelRT/Ecs/Graphics/DefaultRenderingSystem.h
```diff
@@ -53,6 +54,20 @@ namespace NovelRT::Ecs::Graphics
         void ResolveTextureFutureResults();
 
     public:
+        Utilities::Event<Timing::Timestamp,
```
tannergooding, line 57: This is a lot of args, it's going to be "very expensive". It should likely be changed to be a `struct`.

In general, you don't want to pass more than ~4-5 args directly due to how the ABI has to handle them.

RubyNova: Right, but I wasn't sure what you wanted, so I just "threw everything" at it.

If you can narrow down what it is you actually require then this could probably be reduced, otherwise I'll just make a struct with "everything" in it.

tannergooding: I think just having everything for now is convenient, and versionable

we can always trim it down later if appropriate, but a struct is still "goodness" because we can extend it with new parameters without a big breaking change

tannergooding (approved):
lgtm for now
can be tweaked as needed later

PR #417, opened 2022-02-22
Introduce macOS Support

capnkenny:
Did the bare minimum to get macOS support up and running w/ GLFW + Vulkan (MoltenVK).
Tested on a 2012 Mac Mini running macOS Monterey - this assumes Intel builds atm.

Was able to get the EcsPipeline demo up and running using this branch.
Does not include a workflow to test builds.


capnkenny:
I expect this to die horribly as I've removed the vulkan loader and VMA usage :D I'll fix asap

conanfile.py
```diff
@@ -80,29 +79,5 @@ def build(self):
...
-    #         self.copy("*freetype.dll", dst="bin", keep_path=False)
-    #         self.copy("*glfw3.dll", dst="bin", keep_path=False)
-    #         self.copy("*ogg.dll", dst="bin", keep_path=False)
-    #         self.copy("*OpenAL32.dll", dst="bin", keep_path=False)
-    #         self.copy("*opus.dll", dst="bin", keep_path=False)
-    #         self.copy("*sndfile.dll", dst="bin", keep_path=False)
-    #         self.copy("*vorbis.dll", dst="bin", keep_path=False)
-    #         self.copy("*vorbisenc.dll", dst="bin", keep_path=False)
-    #         self.copy("*vorbisfile.dll", dst="bin", keep_path=False)
-
```
capnkenny, line 105: Unneeded - this is only if we're distributing, which we're not with this particular conanfile.

src/NovelRT/CMakeLists.txt
```diff
@@ -13,7 +13,6 @@ find_package(spdlog ${NOVELRT_SPDLOG_VERSION} REQUIRED)
 find_package(TBB ${NOVELRT_ONETBB_VERSION} REQUIRED)
 find_package(Vorbis ${NOVELRT_VORBIS_VERSION} REQUIRED)
 find_package(Vulkan ${NOVELRT_VULKAN_VERSION} REQUIRED)
```
capnkenny, line 15: GLFW pulls in Vulkan based on how we have it set up, so this appears to link properly.

src/NovelRT/Graphics/Vulkan/VulkanGraphicsAdapterSelector.cpp
```diff
@@ -55,14 +55,21 @@ namespace NovelRT::Graphics::Vulkan
...
         }
-
+#else
+        if (!indices.IsComplete() ||
+            !CheckPhysicalDeviceRequiredExtensionSupport(physicalDevice) || supportDetails.formats.empty() ||
+            supportDetails.presentModes.empty())
+        {
+            score = -1;
+        }
+#endif
```
capnkenny, line 72: `deviceFeatures.geometryShader == VK_FALSE` causes the check to fail immediately as the particular device I have does not support a geometry shader. However, I cannot test on a newer Apple device to see if it's just a MoltenVK thing or not, therefore until we can properly test it I've just ifdef'ed that one check. Checks out atm and renders, but just something to keep in mind.

src/NovelRT/Graphics/Vulkan/VulkanGraphicsContext.cpp
```diff
@@ -358,8 +358,9 @@ namespace NovelRT::Graphics::Vulkan
             std::static_pointer_cast<VulkanGraphicsBuffer>(vertexBufferRegion.GetCollection());
         VkBuffer vulkanVertexBuffer = vertexBuffer->GetVulkanBuffer();
         size_t vulkanVertexBufferOffset = vertexBufferRegion.GetOffset();
+        VkDeviceSize castedVulkanVertexBufferOffset = static_cast<VkDeviceSize>(vulkanVertexBufferOffset);
 
-        vkCmdBindVertexBuffers(vulkanCommandBuffer, 0, 1, &vulkanVertexBuffer, &vulkanVertexBufferOffset);
+        vkCmdBindVertexBuffers(vulkanCommandBuffer, 0, 1, &vulkanVertexBuffer, &castedVulkanVertexBufferOffset);
```
capnkenny, line 363: `size_t` is different on macOS :^) If this isn't safe on other OSes, I can ifdef it.... I rather not though if this doesn't cause a failure on the others.

conanfile.py
```diff
@@ -70,6 +65,10 @@ def configure_cmake(self):
             cmake.definitions["NOVELRT_BUILD_SAMPLES"] = "On"
         else:
             cmake.definitions["NOVELRT_BUILD_SAMPLES"] = "Off"
+        if(self.options.buildinterop):
+            cmake.definitions["NOVELRT_BUILD_INTEROP"] = "On"
+        else:
+            cmake.definitions["NOVELRT_BUILD_INTEROP"] = "Off"
```
capnkenny, line 71: It was missing from a previous commit - I _could_ separate it but adding 4 lines in a separate PR makes no sense to me.

conanfile.py
```diff
@@ -50,8 +46,7 @@ class NovelRTConan(ConanFile):
     def imports(self):
         self.copy("*.dll", dst="thirdparty", src="bin")
         self.copy("*.dll", dst="thirdparty", src="lib")
-        self.copy("*.json", dst="thirdparty", src="bin")
-        self.copy("*.json", dst="thirdparty", src="lib")
```
capnkenny, line 54: Unused.

src/NovelRT/CMakeLists.txt
capnkenny: .... on Macs. Not on Windows, so this will properly refer to the Vulkan Loader conan package on those.

src/NovelRT/Graphics/Vulkan/VulkanGraphicsAdapterSelector.cpp
RubyNova: I'm not sure we need geometry shader support. This might be a leftover from other things. Please test this on Windows.

PR #418, opened 2022-02-22
Introduce macOS Support (Take 2)

capnkenny:
Same as #417


capnkenny:
To note - until we are ready, this _will not_ cover how to build as well as having CI for macOS. It will be covered in a new PR once I've fixed the Conan package issues

src/NovelRT/CMakeLists.txt
```diff
@@ -169,3 +171,10 @@ if(WIN32)
   file(GLOB libs RELATIVE ${CMAKE_BINARY_DIR}/thirdparty ${CMAKE_BINARY_DIR}/thirdparty/*.dll)
   set_target_properties(Engine PROPERTIES RESOURCE "${libs}")
 endif()
+if(APPLE)
+  add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_directory
+          ${CMAKE_BINARY_DIR}/thirdparty
+          $<TARGET_FILE_DIR:Engine>)
+  file(GLOB libs RELATIVE ${CMAKE_BINARY_DIR}/thirdparty ${CMAKE_BINARY_DIR}/thirdparty/*.dylib)
+  set_target_properties(Engine PROPERTIES RESOURCE "${libs}")
+endif()
```
RubyNova, line 180: Missing EOF

src/NovelRT/CMakeLists.txt
```diff
@@ -169,3 +171,10 @@ if(WIN32)
   file(GLOB libs RELATIVE ${CMAKE_BINARY_DIR}/thirdparty ${CMAKE_BINARY_DIR}/thirdparty/*.dll)
   set_target_properties(Engine PROPERTIES RESOURCE "${libs}")
 endif()
+if(APPLE)
+  add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_directory
+          ${CMAKE_BINARY_DIR}/thirdparty
+          $<TARGET_FILE_DIR:Engine>)
```
RubyNova, line 177: Can't this use copy_build_products instead?

RubyNova (changes requested):
One question and one change from me.

src/NovelRT/CMakeLists.txt
capnkenny: Not that I'm aware of - if so, I'd propose a separate PR for it considering it's an existing issue at that point, not something introduced by macOS support :)

src/NovelRT/CMakeLists.txt
capnkenny: Should be fixed now

src/NovelRT/CMakeLists.txt
capnkenny: Just adding another thought - we cannot do this as we don't have a proper target to attach the DLLs/dylibs to (they get attached to Engine after being initially copied so its accessible to anything dependent on Engine)

RubyNova (approved):
LGTM

PR #420, opened 2022-04-05
Remove Ink from readme.

RubyNova:


capnkenny:
Should we change/get rid of the .NET 5 SDK reference in the Windows Sectikn since its no longer needed?

RubyNova:
Probably. I just happened to notice ink still in the readme amongst multiple other issues.

I basically opened this just to get rid of Ink though. We should probably rewrite a lot of the readme properly.

capnkenny (approved):

Issue #400, opened 2022-02-02
`copy_build_products` resolves targets before evaluating conditions


capnkenny:
Self-titled, as described in Discord.

> This should be a condition (`CONDITION TARGET Interop`) but right now it will fail. Remind me to fix it later.

_Originally posted by @FiniteReality in https://github.com/novelrt/NovelRT/pull/399#discussion_r797519147_

Issue #387, opened 2022-01-08
CopyBuildProducts fails if path contains a space


FiniteReality:
CMake snippet from Discord demonstrating the issue:

```cmake
#  (Resources -> JavaBootstrapper)
execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different
  D:/Users/jeuxj/Documents/Source Projects Dev/NovelRT/build/resources/Fonts/Gayathri-Regular.ttf
  D:/Users/jeuxj/Documents/Source Projects Dev/NovelRT/build/src/NovelRT.JavaBootstrapper/Resources/Fonts/Gayathri-Regular.ttf
)
#  (Resources -> JavaBootstrapper)
```
Hopefully using `$<SHELL_PATH>` will work here.

RubyNova:
This is not high priority.

I would move this to good first issue or something. Your focus should be the narrative Lang for now @FiniteReality .

FiniteReality:
Fair enough. I just confirmed locally that it doesn't vomit w/ `$<SHELL_PATH>` wrappers here, hopefully that means it should work with spaces too.

jatinkr123:
Hello, i am new to git and want to solve this problem .Could you guide me for a bit, i have knowledge in c++ coding and i also know basic commands of git.

RubyNova:
Hi there @jatinkr123 ! This will require a basic understanding of CMake - if you have that then we'll be able to help you for sure. :)

PR #421, opened 2022-04-06
Add install target for NovelRT core

capnkenny:
Adds an install target for the NovelRT engine core. 
Also adds a flag to work around filepath resolution issues between the `copy_build_products` function and the `install` targets handling.

Installation was tested using `cmake --install . -v --config Release` on both Ubuntu 20.04 and Windows 11


capnkenny:
Also, some additional cleanup was done for NovelRT.DotNet and conan stuffs

include/NovelRT/NovelRT.h
```diff
@@ -74,55 +74,55 @@
   }
 
   //generic experimental files in root namespace
-  #include "NovelRT/EngineConfig.h"
```
FiniteReality, line 77: We should *probably* make these explicit `#include <NovelRT/Thing>` so that way they're using the same style users would inevitably use. Not particularly important for *this* PR though.

cmake/CopyBuildProducts.cmake
```diff
@@ -1,5 +1,5 @@
 function(copy_build_products target)
-  set(valid_options "DEPENDENCY;DEPENDENCY_FILES;TARGET_LOCATION;CONDITION")
+  set(valid_options "DEPENDENCY;DEPENDENCY_FILES;TARGET_LOCATION;CONDITION;DROP_PATH")
```
FiniteReality, line 2: Could you explain what DROP_PATH is supposed to do? I'm having a bit of a hard time visualising it right now.

src/NovelRT/CMakeLists.txt
```diff
@@ -167,14 +167,32 @@ if(WIN32)
...
+  RESOURCE DESTINATION bin
+)
+#Workaround due to hierarchy destruction when installing by TARGETS
+install(
+  DIRECTORY ${CMAKE_SOURCE_DIR}/include/NovelRT/
+  DESTINATION include/NovelRT
+  FILES_MATCHING PATTERN "*.h"
+)
+#Only done so that shaders are always copied
+install(
```
FiniteReality, line 194: This should be in the resources CMakeLists

include/NovelRT/NovelRT.h
capnkenny: This probably could get changed without much fuss in this PR. I'm indifferent either way we want to handle it.

cmake/CopyBuildProducts.cmake
capnkenny: `DROP_PATH` is probably a _really_ bad name but I could not come up with anything at the time.

Essentially it does two things:
- It drops the filepath of the file for the install target
(i.e. from `../../thirdparty/vorbis.dll` to `vorbis.dll`)
- It will adjust the written `copy_build_products` target to look up one additional directory
(from `../../thirdparty/vorbis.dll` to `../../../thirdparty/vorbis.dll`)

This was done for two reasons -
1. Using relative paths would cause `install` to fail:
>  CMake Error at src/NovelRT/cmake_install.cmake:60 (file):
>   file INSTALL cannot find "D:/repos/kennynovel/src/NovelRT/FLAC++.dll": No
>   error.


  2. Using absolute paths would cause `copy_build_products` to generate an invalid path when copying the Engine's build products

>   Copying required build products to EcsPipeline
>   Error copying file (if different) from "D:/repos/kennynovel/build/src/NovelRT/Debug/../../thirdparty/FLAC++.dll" to "
>   D:/repos/kennynovel/build/samples/EcsPipeline/Debug/../../thirdparty/FLAC++.dll".

Adding it as a configuration allows me to keep the resources separate (as they're working as is) yet also correct the issues that the `dll`s keep running into. I'm _200%_ open to suggestions or ideas on improving or correcting this as needed

FiniteReality: We could probably do another absolute path translation to fix point 2. This might be related to the fix for #387 - could you try that to see if it works?

capnkenny: Sure thing! I'll give that a shot

src/NovelRT/CMakeLists.txt
capnkenny: Moved in the latest version (i believe)

cmake/CopyBuildProducts.cmake
capnkenny: This wasn't working properly - I took a different approach and reverted the changes to the function.

include/NovelRT/NovelRT.h
capnkenny: Still planning on fixing this

capnkenny:
Will need a whole new review and verification from CI that this is good

generateDeps.py
```diff
@@ -0,0 +1,49 @@
...
+    set(ext "*.dylib")
+  endif()
+  install(FILES ${DEPENDENCY_FILES}
+    DESTINATION bin
+  )
+endif()"""
+
+    def CreateCMakeLists(self, depPath, cmakePath):
+        files = os.scandir(depPath)
+        safeExts = [".dll",".so",".a",".dylib"]
```
FiniteReality, line 30: .so/.a files likely don't need to be copied since they're handled by RUNPATH/RPATH (and .a files are basically just tarballs of .o files, which are linked directly into the target library/application anyway)

samples/EcsPipeline/CMakeLists.txt
```diff
@@ -9,11 +9,25 @@ target_link_libraries(EcsPipeline
   Engine
   )
 
-copy_build_products(EcsPipeline
-  DEPENDENCY Resources
-  TARGET_LOCATION $<TARGET_FILE_DIR:EcsPipeline>/Resources
+if(NOVELRT_INCLUDE_CONAN_PATHS AND (WIN32 OR APPLE))
```
FiniteReality, line 12: We can use `CONDITION ...` here rather than this weird if/else stuff if you fix #400

capnkenny:
Closing for now, potentially switching branches

PR #422, opened 2022-04-12
Added basic Installation targets

capnkenny:
CMake Install should now:
-- Copy the required DLL/Lib (Dylib not confirmed atm)
-- Copy all headers (respecting the hierarchy provided)
-- Copy the shaders _always_
-- Generate a `.cmake` to find the proper config


CMakeLists.txt
```diff
@@ -84,3 +84,9 @@ if(NOVELRT_BUILD_TESTS)
 
   add_subdirectory(tests)
 endif()
+
+install(
+  EXPORT NovelRT
+  NAMESPACE NovelRT::
```
RubyNova, line 90: Do we need the double colon here?

RubyNova (approved):
LGTM but gonna wait on @FiniteReality before merge

CMakeLists.txt
capnkenny: I believe not necessarily but it may help? I framed this off of Fabulist's export

RubyNova: I see, ok, well what do I know 😆

FiniteReality: The namespace is what it generates as the prefix in the CMake find_package scripts, for grouping all of the targets from a specific library together. For example:
```cmake
# Create imported target fabulist::lua
add_library(fabulist::lua STATIC IMPORTED)

set_target_properties(fabulist::lua PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "LUA_ANSI"
  INTERFACE_COMPILE_OPTIONS "\$<\$<CXX_COMPILER_ID:MSVC>:/experimental:external>;\$<\$<CXX_COMPILER_ID:MSVC>:/external:I>;\$<\$<CXX_COMPILER_ID:MSVC>:/external:W0>;\$<\$<CXX_COMPILER_ID:MSVC>:/external:templates->"
  INTERFACE_INCLUDE_DIRECTORIES "\$<\$<NOT:\$<CXX_COMPILER_ID:MSVC>>:>"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "\$<\$<NOT:\$<CXX_COMPILER_ID:MSVC>>:>"
)
```

src/NovelRT/CMakeLists.txt
```diff
@@ -156,6 +156,19 @@ target_link_libraries(Engine
     Vulkan::Vulkan
 )
 
+install(
+  TARGETS Engine
+  EXPORT NovelRT
+  DESTINATION lib
+  RUNTIME DESTINATION bin
+)
+#Workaround due to hierarchy destruction when installing by TARGETS
```
FiniteReality, line 165: Post MVP I want to restructure the engine such that each "area" is its own static/object library, and is then linked into the main "engine" - this should simplify the layout of our include files and flatten them so that we can use PUBLIC_HEADER or whatever. Hopefully this'll enforce a lot more separation between various components.

FiniteReality (approved):
Seems good enough to me

PR #424, opened 2022-04-13
Updated minimum CMake Version to 3.19

### LINKED ISSUE #416: Correct CMake Requirements
As found by our Conan PR, CMake version 3.18 is _not_ supported for builds as the `INTERFACE_LIBRARY` allowlist still exists, therefore our `Resources` target as well as our `copy_build_products` function both throw errors as we access properties that are not allowed in versions below 3.19.0.

We have two options here - 

- Bump the versioning up to 3.19
- Try to work around all the restrictions (namely being `RESOURCE` and `BINARY_DIR` properties being accessed by `INTERFACE` libraries in CMake)

I opt for the former as it's less work, and less catastrophic to our environment atm.

capnkenny:
Resolves #416

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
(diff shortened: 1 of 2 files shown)


RubyNova (approved):
LGTM

PR #425, opened 2022-04-16
Reworked Atom factory methods into an actual engine feature.

### LINKED ISSUE #249: Make atom factory functions internal
We can make our Atom factory functions de-facto internal if we make a single public factory function, which works using template parameters. These template parameters are mostly meaningless, but they allow us to instantiate a unique "flavour" of the function which follows the accessibility of the template parameter itself. If we then make the type name used to access this instantiation private, we can make the factory function de-facto internal, even if the overall definition is public in the Atom type.

The definition would look something like this:
```cpp
template <typename>
int get_id()
{
    static int id = 0;
    
    return id++;
}
```
And to use it, we'd define a type name (it doesn't even have to be complete; just something speakable and referable) in our implementation file(s) and use the function:
```cpp
namespace { // THIS BIT IS NECESSARY!
    class event_id;
}

Atom do_thing_and_get_its_id()
{
    return get_id<event_id>();
}
```
As for the `THIS BIT IS NECESSARY!` comment, that's an unnamed namespace, and it *basically* generates a unique name for the namespace, so that no other implementation file can get the same namespace, even if the symbol gets exported somehow. In C++17, it also gets "internal linkage" which means the symbol(s) don't get exported, meaning nobody else can use them.

The overall effect is that nobody else can have the same `get_id<event_id>();`, unless we explicitly make it so by giving the namespace a non-unique name.

RubyNova:
This PR reworks our existing static factory methods for generating atomic IDs for things into an actual engine feature.

The main motivation behind this change is that the existing factory methods are restrictive, not even technically supposed to be exposed to the public API, and are just messy to maintain. As such I have just reworked it into a supported API.

I am unsure if I need this for the upcoming engine persistence features to work but from some of my very rough plans it might be a possibility - so future changes I make for persistence will be based off this just in case.

Resolves #249 - albeit in a different way than I had originally intended.

### DIFF
--- a/include/NovelRT/Atom.h
+++ b/include/NovelRT/Atom.h
@@ -4,6 +4,7 @@
 #include <atomic>
 #include <cstddef>
 #include <limits>
+#include <string>
 
 #ifndef NOVELRT_ATOM_H
 #define NOVELRT_ATOM_H
@@ -71,29 +72,33 @@ namespace NovelRT
             return _value;
         }
 
-        // TODO: These should be internal to NovelRT
-
-        static Atom GetNextEventHandlerId() noexcept;
-
-        static Atom GetNextFontSetId() noexcept;
-
-        static Atom GetNextTextureId() noexcept;
-
-        static Atom GetNextComponentTypeId() noexcept;
-
-        static Atom GetNextEntityId() noexcept;
-
-        static Atom GetNextSystemId() noexcept;
-
-        static Atom GetNextEcsTextureId() noexcept;
-
-        static Atom GetNextEcsVertexDataId() noexcept;
-
-        static Atom GetNextEcsPrimitiveId() noexcept;
-
-        static Atom GetNextEcsGraphicsPipelineId() noexcept;
-
         static Atom GetNextEcsPrimitiveInfoConfigurationId() noexcept;
+    };
+
+    class AtomFactory
+    {
+    private:
+        std::atomic_uintptr_t _currentValue;
+        bool _moved;
+
+    public:
+        AtomFactory() noexcept;
+        explicit AtomFactory(Atom startingValue) noexcept;
+        AtomFactory(const AtomFactory& other) noexcept;
+        AtomFactory(AtomFactory&& other) noexcept;
+        AtomFactory& operator=(const AtomFactory& other) noexcept;
+        AtomFactory& operator=(AtomFactory&& other) noexcept;
+        ~AtomFactory() = default;
+
+        [[nodiscard]] Atom GetNext();
+
+        void SetToValue(Atom newValue);
+    };
+
+    class AtomFactoryDatabase
+    {
+    public:
+        [[nodiscard]] static AtomFactory& GetFactory(const std::string& factoryName) noexcept;
     };
 
     class AtomHashFunction
--- a/src/NovelRT/Atom.cpp
+++ b/src/NovelRT/Atom.cpp
@@ -5,10 +5,10 @@
 #ifndef __TBB_PREVIEW_MUTEXES
 #define __TBB_PREVIEW_MUTEXES 1
 #endif
+#include <NovelRT/Exceptions/Exceptions.h>
+#include <mutex>
 #include <oneapi/tbb/mutex.h>
 #include <unordered_map>
-#include <mutex>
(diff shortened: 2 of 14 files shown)


src/NovelRT/Atom.cpp
```diff
@@ -2,89 +2,90 @@
 // for more information.
 
 #include <NovelRT/Atom.h>
+#ifndef __TBB_PREVIEW_MUTEXES
+#define __TBB_PREVIEW_MUTEXES 1
+#endif
+#include <NovelRT/Exceptions/Exceptions.h>
+#include <mutex>
```
capnkenny, line 9: Do we need `std::mutex` if we're now using `tbb::mutex`?

include/NovelRT/Atom.h
```diff
@@ -71,29 +72,33 @@ namespace NovelRT
...
-        static Atom GetNextComponentTypeId() noexcept;
-
-        static Atom GetNextEntityId() noexcept;
-
-        static Atom GetNextSystemId() noexcept;
+        static Atom GetNextEcsPrimitiveInfoConfigurationId() noexcept;
+    };
 
-        static Atom GetNextEcsTextureId() noexcept;
+    class AtomFactory
```
capnkenny, line 78: Why is this not included in its own header file? Seems partially inconsistent with our standards but I could be confusing the old way vs new way

include/NovelRT/Atom.h
```diff
@@ -71,29 +72,33 @@ namespace NovelRT
...
+    };
 
-        static Atom GetNextEcsPrimitiveInfoConfigurationId() noexcept;
+    class AtomFactoryDatabase
+    {
+    public:
+        [[nodiscard]] static AtomFactory& GetFactory(const std::string& factoryName) noexcept;
     };
 
     class AtomHashFunction
```
capnkenny, line 104: 3 additional classes in one header file? o_O

src/NovelRT/Ecs/SystemScheduler.cpp
```diff
@@ -75,7 +75,9 @@ namespace NovelRT::Ecs
 
     void SystemScheduler::RegisterSystem(std::function<void(Timing::Timestamp, Catalogue)> systemUpdatePtr) noexcept
     {
-        Atom id = Atom::GetNextSystemId();
+        static AtomFactory& systemIdFactory = AtomFactoryDatabase::GetFactory("SystemId");
```
capnkenny, line 78: How are these strings defined to determine which factory to get? Is this something that will just be predefined by the Engine? (Meaning we provide users with a specific set of strings for "pre-defined" factories else they manage their own?)

capnkenny (commented):
Generally speaking, LGTM. Have a few questions about the strings and other things however preventing me from giving explicit approval just yet.

src/NovelRT/Atom.cpp
RubyNova: for `std::scoped_lock`

include/NovelRT/Atom.h
RubyNova: because I do sometimes group types that are used basically all the time together occasionally this isn't convention or anything

include/NovelRT/Atom.h
RubyNova: See previous comment

src/NovelRT/Ecs/SystemScheduler.cpp
RubyNova: If you check the implementation of GetFactory you'll see it just generates a new one for the string if it doesn't exist.

src/NovelRT/Atom.cpp
capnkenny: Coolio, ty for confirming

src/NovelRT/Ecs/SystemScheduler.cpp
capnkenny: Is there ever a case where an end user will want to use one of the internal factories?

capnkenny: After discussion in Discord - generally speaking, these strings should not be used externally nor shared. May be refactored into a private header at some point, but for MVP this is alright.

capnkenny (approved):
LGTM 💸

PR #427, opened 2022-04-18
Added release action

capnkenny:
Repurposed Fabulist's Build GH Action to allow for manually requesting a build and packaging of NovelRT's Engine.

Also fixed/tweaked a few areas as They were found during testing.


CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ set(NOVELRT_VULKAN_VERSION "1.2.172" CACHE STRING "Vulkan version")
 set(NOVELRT_VULKAN_MEMORY_ALLOCATOR_VERSION "2.3.0" CACHE STRING "VulkanMemoryAllocator version")
 
 if (NOVELRT_INCLUDE_CONAN_PATHS)
+  set(CONAN_CMAKE_SILENT_OUTPUT ON)
```
RubyNova, line 55: What's this do?

scripts/build.sh
```diff
@@ -94,11 +94,11 @@ function CreateDirectory {
 
 function Generate {
   conan config install https://github.com/novelrt/ConanConfig.git
-  conan install . -if "$BuildDir" --build=missing --build=bison --profile linux-clang10-amd64
+  conan install . -if "$BuildDir" --build=missing --profile linux-clang10-amd64
```
RubyNova, line 97: This works now?

src/NovelRT/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ set(SOURCES
 
 add_library(Engine SHARED ${SOURCES})
 
-set_property(TARGET Engine PROPERTY OUTPUT_NAME "NovelRT")
+set_property(TARGET Engine PROPERTY OUTPUT_NAME "NovelRT-$<CONFIG>")
```
RubyNova, line 92: What's this change do exactly?

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -56,18 +56,19 @@ set(SOURCES
 add_executable(Engine_Tests ${SOURCES})
 
 if(NOVELRT_BUILD_INTEROP)
-    copy_build_products(Engine_Tests
-      DEPENDENCY Engine
-      TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
+  copy_build_products(Engine_Tests
+    DEPENDENCY Engine
+    TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
```
RubyNova, line 61: What actually changed here? Just spacing?

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -56,18 +56,19 @@ set(SOURCES
...
 
-      DEPENDENCY Interop
-      TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
-    )
+    DEPENDENCY Interop
+    TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
+  )
```
RubyNova, line 65: Spacing changes again? Or am I missing something?

tests/NovelRT.Tests/CMakeLists.txt
```diff
@@ -56,18 +56,19 @@ set(SOURCES
...
 else()
-    copy_build_products(Engine_Tests
-            DEPENDENCY Engine
-            TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
-    )
+  include(${CMAKE_SOURCE_DIR}/cmake/CopyBuildProducts.cmake)
+  copy_build_products(Engine_Tests
+    DEPENDENCY Engine
+    TARGET_LOCATION $<TARGET_FILE_DIR:Engine_Tests>
+  )
```
RubyNova, line 71: I think this is also just spacing (and the include that was probably missing?) But just wanted to check.

RubyNova (commented):
Just questions from me

.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,146 @@
...
+    - uses: ncipollo/release-action@v1
+      id: create_release
+      with:
+        tag: ${{ steps.tag_version.outputs.new_tag }}
+        name: NovelRT ${{ steps.tag_version.outputs.new_tag }}
+        body: ${{ steps.tag_version.outputs.changelog }}
+        prerelease: true
+        draft: true
+
+  build-windows:
```
FiniteReality, line 31: You can merge this with build-ubuntu using conditions: <https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution>

I'd do something like `run: cibuild.cmd if: ${{ startsWith(matrix.os, 'windows') }}`

scripts/build.sh
```diff
@@ -75,7 +75,7 @@ function Build {
   if [ -z "$remaining" ]; then
     cmake --build "$BuildDir" --config "$configuration"
   else
-    cmake --build "$BuildDir" --config "$configuration" "${remaining[@]}"
+    cmake --build "$BuildDir" --config "$configuration" $remaining
```
FiniteReality, line 78: The removed quotes here are usually important - it'll stop CMake from treating `/path to/file with spaces.txt` as multiple arguments.

scripts/machine-setup.ps1
```diff
@@ -1,13 +1,12 @@
...
 # This is intended for setting up CI machines with the correct dependencies.
 
-$DotNetInstallDirectory = Join-Path -Path $HOME -ChildPath "dotnet"
 $ProgressPreference = 'SilentlyContinue'
 
 if (!(Get-Command python -ErrorAction SilentlyContinue ))
 {
   $PythonInstaller = New-TemporaryFile | Rename-Item -NewName { $_.Name + ".exe" } -PassThru
-  Invoke-WebRequest -Uri "https://www.python.org/ftp/python/3.8.3/python-3.8.3.exe" -OutFile $PythonInstaller
+  Invoke-WebRequest -Uri "https://www.python.org/ftp/python/3.10.4/python-3.10.4-amd64.exe" -OutFile $PythonInstaller
```
FiniteReality, line 9: We should extract this to a constant somewhere IMHO. Maybe we could even setup a build environment pipeline which rebuilds and pushes dockerfiles based on changes to them?

src/NovelRT/CMakeLists.txt
FiniteReality: It allows the artifacts to be merged so that downloading NovelRT allows CMake to choose debug/release builds based on the config.

FiniteReality: Specifically, it'll generate outputs looking something like this, after having its outputs merged: 
Screenshot_20220418_150149

scripts/build.sh
FiniteReality: Looks like it.

src/NovelRT/CMakeLists.txt
RubyNova: So this name will carry over to actual released games?

RubyNova: I should also clarify from a C API if the name of the DLL is constantly changing we can't do bindings.

CMakeLists.txt
capnkenny: This calms down the spewing of Conan information during CMake generation to a measily ~17 lines.

scripts/build.sh
RubyNova: We should update the README to reflect this then, probably.

src/NovelRT/CMakeLists.txt
FiniteReality: > So this name will carry over to actual released games?

We could likely drop the prefix for MinSizeRel builds - it was just easier to keep it to simplify the logic.

> I should also clarify from a C API perspective if the name of the DLL is constantly changing we can't do bindings.

For .NET there's [SetDllImportResolver](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver?view=net-6.0#system-runtime-interopservices-nativelibrary-setdllimportresolver(system-reflection-assembly-system-runtime-interopservices-dllimportresolver)) which should handle this just fine.

RubyNova: How do we use this via ClangSharp?

RubyNova: In addition, how does a game use this?

FiniteReality: > How do we use this via ClangSharp?

The same way as normal, plus a call to SetDllImportResolver if necessary.

> In addition, how does a game use this?

The same way as normal.

RubyNova: So we can call this from ClangSharp's command line? because that's the normal way

Could we set up this logic as part of the C# .NET template when creating a C# NovelRT game?

.github/workflows/publish_release.yml
capnkenny: The only issue I ran into with this was that I could not get the container to comply with the `if` statement to only be used with ubuntu (GH was stating I had a syntax error?)

src/NovelRT/CMakeLists.txt
FiniteReality: ClangSharp doesn't care about the library - SetDllImportResolver handles that. You don't need to set it up as part of the .NET template because the bindings library will handle calling SetDllImportResolver.

RubyNova: Oh OK I think I get you. Thanks!

RubyNova: cc @jeuxjeux20 how problematic will this be for Java/Kotlin?

.github/workflows/publish_release.yml
FiniteReality: Interesting. The syntax I suggested *should* work.

FiniteReality: Also, if I'm being honest, since we're going through all of this effort to simplify our CI pipeline, we should probably just remove the build scripts entirely. I seriously doubt that we're going to migrate away from GitHub Actions in the near future, and reproducing a CI environment is already possible using something like [act](https://github.com/nektos/act).

src/NovelRT/CMakeLists.txt
jeuxjeux20: > cc @jeuxjeux20 how problematic will this be for Java/Kotlin?

It's actually fine, we can easily choose which DLL to load in Kotlin with `System.load("<find_the_dll>")`. However I wonder how this will affect the "extension" library `NovelRT.JavaSupport.dll` which has a dependency on `NovelRT.dll`

RubyNova: Ah right because JNI is AOT C/C++ stuff right?

FiniteReality: As long as it's built against the right version, it should work just fine.

tests/NovelRT.Tests/CMakeLists.txt
capnkenny: Yep, just made spacing more consistent as for some reason certain IDEs seem to indent farther than they should with CMakeLists

tests/NovelRT.Tests/CMakeLists.txt
capnkenny: Same as previous comment with spacing

tests/NovelRT.Tests/CMakeLists.txt
capnkenny: Yeeee

capnkenny:
Closing PR temporarily so that I can test on my fork and re-open once complete

PR #453, opened 2022-06-10
Move Google Test to FetchContent

capnkenny:
Moves Google Test to Fetch Content call. Also cleans up a few loose ends in main.

EDIT: This PR also now includes updating CI to only invoke Conan to pull down dependencies that have not been migrated yet. All building should be done via CMake at this time.


RubyNova (approved):
I'm guessing the tests get turned off with FORCE due to the options stuff you mentioned to me in discord?

LGTM on a first glance

capnkenny:
So depending on the CMakeLists, if CMake policy CMP0077 isn't set CMake will ignore the set and force it back to default, which is ON in the case of jsonscons. I mainly did the change as it was similar to the one highlighted by Google Test (which came straight from their Quick Start Guide) to make sure that it's forced in the cache to be the value that we set.

It was also easier to keep it consistent lol.

.github/workflows/build-system.yml
```diff
@@ -22,6 +22,8 @@ jobs:
           container: novelrt/novelrt-build:latest
           profile: linux-clang10-amd64
           name: Ubuntu - Clang 10
+          c: CC=clang-10
+          cxx: CXX=clang++-10
```
capnkenny, line 26: Because we're no longer using Conan to define the compiler, we're explicitly defining it for Ubuntu as it is not the stock default one for GH's runners. We can optionally exclude it (as we do for macOS and Windows) to let GH compile with what is default.

.github/workflows/build-system.yml
```diff
@@ -54,7 +56,11 @@ jobs:
       - uses: novelrt/setup-cpp@v1
       - name: Build and Test NovelRT - ${{ matrix.configuration }}
         id: build
-        run: conan build . --build-folder build
+        run: |
+          ${{ matrix.c }} ${{ matrix.cxx }} cmake -S . -B build -DCMAKE_BUILD_TYPE=${{ matrix.configuration }}
+          cmake --build build --config ${{ matrix.configuration }}
+          cd build/tests/NovelRT.Tests
+          ctest
```
capnkenny, line 63: Just calling out: Conan was having issues running tests (it was invoking Run_Tests.vcxproj???), so I've switch to CMake builds only (as we're moving from Conan anyways).

Also, if our matrix doesn't have a definition for `c` or `cxx` (i.e. Windows or macOS currently), it just puts a space and shouldn't break the command.

internal/jsoncons/CMakeLists.txt
```diff
@@ -14,5 +14,5 @@ FetchContent_Declare(jsoncons
         INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/inst"
         LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/log"
         )
-
+set(JSONCONS_BUILD_TESTS OFF CACHE BOOL "" FORCE)
```
capnkenny, line 17: To add context to the conversation earlier in the request, if `CACHE BOOL "" FORCE` wasn't set here, CMake complains about policy CMP0077 not being set (it appears only set in GLFW) and ignores our setting of the option.

.github/workflows/build-system.yml
FiniteReality: There's likely a better way to do this, but for now I'm fine with leaving this just to get the gtest stuff in.

FiniteReality (approved):

PR #454, opened 2022-06-10
Move ms-gsl to FetchContent dependency.

RubyNova:


internal/GSL/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+        DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/dl"
+        SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/src"
+        SUBBUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}/build"
+        BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/bin"
+        INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/inst"
+        LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/log"
+        )
+
+FetchContent_MakeAvailable(GSL)
+install(
```
FiniteReality, line 19: Why are we installing it? If CMake is warning us that it's necessary, it needs to be in the built output directory or else people will have problems using NovelRT.

FiniteReality (approved):

internal/GSL/CMakeLists.txt
RubyNova: It's required because it is header only and therefore won't build.

It ends up in the correct place on install by doing this and allows people to use installed copies of the library.

PR #456, opened 2022-06-12
Move GLM to FetchContent dependency (and make header-only)

capnkenny:


RubyNova (approved):
LGTM yeet dab 420 get me off conan already

PR #459, opened 2022-06-14
Static link from-source build of libpng via FetchContent.

RubyNova:
This resolves libpng being a conan dependency. However I have had to temporarily add zlib to the conanfile.py because the zlib cmakelists is actually a spawn of satan and I couldn't get it to work in time for this PR - so the list of dependencies to knock off the conanfile hasn't actually changed.

### DIFF
--- a/internal/libpng/CMakeLists.txt
+++ b/internal/libpng/CMakeLists.txt
@@ -1,0 +1,21 @@
+include(FetchContent)
+
+FIND_PACKAGE(ZLIB)
+
+FetchContent_Declare(libpng
+        URL https://github.com/glennrp/libpng/archive/refs/tags/v1.6.35.zip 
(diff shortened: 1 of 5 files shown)


internal/libpng/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+include(FetchContent)
+
+FIND_PACKAGE(ZLIB)
```
capnkenny, line 3: This should be required so that if it fails CMake does not continue generation (in case something went wrong source/Conan-wise and zlib couldn't be found).

Should also require a version as minimum, but I'm not sure if it's 100% necessary to version lock like that atm.

capnkenny (changes requested):
Just one minor tweak I'd suggest, otherwise this looks fine for now

capnkenny (approved):

PR #461, opened 2022-06-16
Fixed MacOs Glfw SigAbort when creating a window

GooStuffedDoll:
Due to weirdness in the macOs system, when attempting to create a window it will return 0 extensions, which we had labeled as an error earlier. Now that we have fixed this issue we also ran into the problem of adding window attributes mac did not like. Added an ifdef and a nullptr check to solve the issues.
NOTE:: THIS DOES NOT SOLVE THE PROBLEM OF CONAN NOT PROVIDING A PROPER VULKAN LOADER FOR MACOS!!

### DIFF
--- a/src/NovelRT/Windowing/Glfw/GlfwWindowingDevice.cpp
+++ b/src/NovelRT/Windowing/Glfw/GlfwWindowingDevice.cpp
@@ -49,10 +49,10 @@ namespace NovelRT::Windowing::Glfw
             throw Exceptions::InitialisationFailureException("GLFW3 failed to initialise.", std::string(output));
         }
 
-        #ifndef __APPLE__
-            glfwSetWindowAttrib(window, GLFW_VISIBLE, GLFW_TRUE);
-        #endif
-        
+#ifndef __APPLE__
+        glfwSetWindowAttrib(window, GLFW_VISIBLE, GLFW_TRUE);
+#endif
+
         glfwSetWindowAttrib(window, GLFW_RESIZABLE, windowMode == NovelRT::Windowing::WindowMode::Windowed);
         glfwSetWindowUserPointer(window, this);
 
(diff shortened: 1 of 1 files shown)


src/NovelRT/Windowing/Glfw/GlfwWindowingDevice.cpp
```diff
@@ -71,9 +74,12 @@ namespace NovelRT::Windowing::Glfw
...
         {
-            const char* output = nullptr;
-            glfwGetError(&output);
-            throw Exceptions::InitialisationFailureException("GLFW3 failed to initialise.", std::string(output));
+           const char* output = nullptr;
+           glfwGetError(&output);
+           if(output != nullptr)
+           {
+                throw Exceptions::InitialisationFailureException("GLFW3 failed to initialise.", std::string(output));
+           }
```
RubyNova, line 82: This is missing the additional branch I asked for - I've appended a suggestion you can apply. This slightly differs from what I originally wanted in terms of the actual message, but it saves adding an additional ctor overload if it doesn't exist and semantically makes more sense given how Vulkan works.

```suggestion
           if(output != nullptr)
           {
                throw Exceptions::InitialisationFailureException("GLFW3 failed to initialise.", std::string(output));
           }
           else
           {
               throw Exceptions::InitialisationFailureException("GLFW3 failed to initialise.", "Attempting to fetch the required Vulkan extensions failed with a count of zero.");
           }
```

src/NovelRT/Windowing/Glfw/GlfwWindowingDevice.cpp
```diff
@@ -49,7 +49,10 @@ namespace NovelRT::Windowing::Glfw
             throw Exceptions::InitialisationFailureException("GLFW3 failed to initialise.", std::string(output));
         }
 
-        glfwSetWindowAttrib(window, GLFW_VISIBLE, GLFW_TRUE);
+        #ifndef __APPLE__
+            glfwSetWindowAttrib(window, GLFW_VISIBLE, GLFW_TRUE);
```
RubyNova, line 53: Indentation looks a bit scuffed here. Clang-format will probably fix this. I'd address the other change first though otherwise you'll be running clang-format twice.

RubyNova (changes requested):
Functionally looks fine, just a code style oddity and an exception UX issue that needs addressing.

PR #462, opened 2022-06-17
Fix libpng cmake not detecting neon on m1

GooStuffedDoll:


internal/libpng/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ FetchContent_Declare(libpng
         INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/inst"
         LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/log"
 )
-
+set(PNG_ARM_NEON "off" CACHE STRING "" FORCE)
```
FiniteReality, line 19: I feel like this is a setting we shouldn't be changing; turning off SIMD for ARM is likely going to result in far poorer performance.

RubyNova: We have to do so.

It literally breaks M1.

Unless you're willing to submit a PR to libpng this is our only option for now. That said, it should be in an `if()` to check for M1, and only disable it in that case.

GooStuffedDoll: With some more research we haven't seem to find a way for apple to automate this without manually adding the flag 
-DCMAKE_APPLE_SILICON_PROCESSOR="arm64"
When creating the build system with cmake .. <flag_here>
We've decided to add an addendum to the README guide on building NovelRT to solve this issue.

GooStuffedDoll: Adjusted readme and cmake file for libpng to allow people to easily build for apple-silicon

README.md
```diff
@@ -175,15 +175,16 @@ If you would like to change any of the settings prior to building, pass them in
...
-
-Then you should be able to configure and build the default setup, like so:
+If you have an intel mac system, then configure the cmake files like so:
 ```
-conan build .. --build-folder .
+cmake ..
 ```
-
-If you would like to build using CMake instead, you can configure and build, like so:
+If you have an apple-silicon (m1 or m2 or etc) system, you must add this flag after the `cmake ..` like so:
```
RubyNova, line 182: This reads a bit weird - I think it's because of how you've used the command as part of the sentence.

Instead I would probably say something like "you must add an additional flag to the configuration command, like so:"

That way the example is in one place without duplication and the sentence is more cohesive.

RubyNova (changes requested):
Minor readme change requested - I'll approve once that's fixed. Thanks again!

README.md
GooStuffedDoll: Fixed

RubyNova (approved):
LGTM

PR #464, opened 2022-06-20
Added freetype and zlib-ng via FetchContent

capnkenny:
- Moved to zlib-ng since regular ol' zlib was being a pain 
(static linking so we don't cause issues with system versions of zlib)

- Made adjustments so libpng uses zlib-ng

- Made freetype static link both of the above

### DIFF
--- a/internal/CMakeLists.txt
+++ b/internal/CMakeLists.txt
@@ -3,13 +3,12 @@ add_subdirectory(glfw)
 add_subdirectory(glm)
 add_subdirectory(GSL)
 add_subdirectory(jsoncons)
-add_subdirectory(oneTBB)
-add_subdirectory(spdlog)
-
 #Ordering this way required as libpng depends on zlib, and freetype depends on both
 add_subdirectory(zlib-ng)
 add_subdirectory(libpng)
 add_subdirectory(freetype)
+add_subdirectory(oneTBB)
+add_subdirectory(spdlog)
 
 if(NOVELRT_BUILD_TESTS)
   add_subdirectory(gtest)
--- a/internal/CMakeLists.txt
+++ b/internal/CMakeLists.txt
@@ -3,9 +3,13 @@ add_subdirectory(glfw)
 add_subdirectory(glm)
 add_subdirectory(GSL)
 add_subdirectory(jsoncons)
-add_subdirectory(libpng)
 add_subdirectory(oneTBB)
 add_subdirectory(spdlog)
+
+#Ordering this way required as libpng depends on zlib, and freetype depends on both
+add_subdirectory(zlib-ng)
+add_subdirectory(libpng)
+add_subdirectory(freetype)
 
 if(NOVELRT_BUILD_TESTS)
   add_subdirectory(gtest)
--- a/internal/freetype/CMakeLists.txt
+++ b/internal/freetype/CMakeLists.txt
@@ -1,0 +1,21 @@
+include(FetchContent)
+message(STATUS "Fetching Freetype")
+
+FetchContent_Declare(freetype
+        URL https://gitlab.freedesktop.org/freetype/freetype/-/archive/VER-2-12-1/freetype-VER-2-12-1.tar.gz
+        URL_HASH SHA512=d2c6e8af7f25fef12222405c0825fa9d96d724d4633435a736da53b74ef9ea9079ec64c08172f908e2a6af23305cb8743636fa5ca51b36d7a839c89c780e33e0
+
+        PREFIX "${CMAKE_CURRENT_BINARY_DIR}"
+        TMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/tmp"
+        STAMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/stamp"
+        DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/dl"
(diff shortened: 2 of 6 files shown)


internal/CMakeLists.txt
```diff
@@ -3,10 +3,14 @@ add_subdirectory(glfw)
...
 add_subdirectory(jsoncons)
-add_subdirectory(libpng)
 add_subdirectory(oneTBB)
 add_subdirectory(spdlog)
 
+#Ordering this way required as libpng depends on zlib, and freetype depends on both
+add_subdirectory(zlib-ng)
+add_subdirectory(libpng)
+add_subdirectory(freetype)
+
```
RubyNova, line 13: I'm wondering if this should be after jsoncons? That's how I would've done it if zlib had worked in the first place. Your call though!

internal/freetype/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+        DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/dl"
+        SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/src"
+        SUBBUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}/build"
+        BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/bin"
+        INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/inst"
+        LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/log"
+        )
+set(FT_DISABLE_BZIP2 TRUE CACHE BOOL "Disable BZip2 dependency" FORCE)
+set(FT_DISABLE_HARFBUZZ TRUE CACHE BOOL "Disable Harfbuzz dependency" FORCE)
+set(FT_DISABLE_BROTLI TRUE CACHE BOOL "Disable Brotli dependency" FORCE)
```
RubyNova, line 20: I forget - UE only uses zlib right?

src/NovelRT/CMakeLists.txt
```diff
@@ -184,6 +183,9 @@ if(WIN32)
   add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy
     $<TARGET_FILE:tbb>
     $<TARGET_FILE_DIR:Engine>/$<TARGET_FILE_NAME:tbb>)
+  add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy
+    $<TARGET_FILE:zlib>
+    $<TARGET_FILE_DIR:Engine>/$<TARGET_FILE_NAME:zlib>)
```
RubyNova, line 188: Do we do this for other targets? Or how are they handled?

src/NovelRT/CMakeLists.txt
```diff
@@ -193,6 +195,9 @@ elseif(APPLE)
   add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy
           $<TARGET_FILE:tbb>
           $<TARGET_FILE_DIR:Engine>/$<TARGET_FILE_NAME:tbb>)
+  add_custom_command(TARGET Engine POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy
+          $<TARGET_FILE:zlib>
+          $<TARGET_FILE_DIR:Engine>/$<TARGET_FILE_NAME:zlib>)
```
RubyNova, line 200: is this duplicated?

src/NovelRT/CMakeLists.txt
capnkenny: We do this for every target that builds a shared library - with that said, this can be removed since it's static now

src/NovelRT/CMakeLists.txt
capnkenny: It was purposefully but I think I can condense it

internal/CMakeLists.txt
capnkenny: I moved it - no preference here

internal/zlib-ng/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+        BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/bin"
+        INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/inst"
+        LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/log"
+)
+set(ZLIB_COMPAT ON CACHE BOOL "Compile with zlib compatible API" FORCE)
+set(ZLIB_ENABLE_TESTS OFF CACHE BOOL "Build test binaries" FORCE)
+set(INC_INSTALL_DIR "include" CACHE STRING "Relocate includes installation folder" FORCE)
+set(LIB_INSTALL_DIR "lib" CACHE STRING "Relocate libraries installation folder" FORCE)
+FetchContent_MakeAvailable(zlib)
+#set(SKIP_INSTALL_ALL ON CACHE BOOL "Skips default ZLIB installation" FORCE)
```
RubyNova, line 22: Maybe delete this?

RubyNova (changes requested):
Minor commented code removal otherwise LGTM just waiting on that

RubyNova (approved):

PR #465, opened 2022-06-21
Fix FutureResult race condition with the GPU.

RubyNova:


capnkenny (approved):
LGTM - tested on my machine (which exposed the bug originally) and appears to work on every load now.

PR #468, opened 2022-06-28
Added the stduuid library.

jeuxjeux20:
This includes the [stduuid](https://github.com/mariusbancila/stduuid) library to the `internal` folder, which enables the use of UUIDs in persistence with the MT19937 algorithm.  
We've used a specific version commit of stduuid containing important fixes for including GSL.


internal/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ add_subdirectory(Fabulist)
 add_subdirectory(glfw)
 add_subdirectory(glm)
 add_subdirectory(GSL)
+add_subdirectory(stduuid)
```
RubyNova, line 5: This looks to be out of order alphabetically. Can you correct this or put a comment as to why it is this way?

internal/stduuid/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+  TMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/tmp"
+  STAMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/stamp"
+  DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/dl"
+  SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/src"
+  SUBBUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}/build"
+  INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/inst"
+  LOG_DIR "${CMAKE_CURRENT_BINARY_DIR}/log"
+  )
+set(UUID_USING_CXX20_SPAN TRUE CACHE BOOL "Avoids using the bundled GSL library from stduuid" FORCE)
+FetchContent_MakeAvailable(stduuid)
```
RubyNova, line 18: Missing EOF.

RubyNova (changes requested):

internal/CMakeLists.txt
jeuxjeux20: I thought the alphabetical order was just entirely broken after the `zlib-ng` library: there's zlib, libpng, freetype, onetbb and spdlog. At this point i'm unsure where to put the line (considering stduuid sort of depends on GSL).

src/NovelRT/CMakeLists.txt
```diff
@@ -157,6 +157,7 @@ target_link_libraries(Engine
     png_static
     SndFile::SndFile
     spdlog
+    stduuid
```
capnkenny, line 160: If `stduuid` is a header only library, we probably don't want to be doing this as it's assuming we're linking to the `stduuid` target and will require an export target to be installed.

capnkenny (changes requested):
You may want to review GSL and see how it's being included - https://github.com/novelrt/NovelRT/blob/fe3f30a37c6dae26931778142d3c403fb5a09bf5/src/NovelRT/CMakeLists.txt#L107

As long as you can map out the source directory from the CMakeCache for `stduuid`, you can do the same thing here (after GSL of course) without having to link to it or create a CMake target dependency (to get around the install issue)

src/NovelRT/CMakeLists.txt
jeuxjeux20: Thanks, fixed!

capnkenny (approved):
LGTM

PR #474, opened 2022-07-03
Fix a Vulkan race condition between the swapchain and command buffer

tannergooding:
There was a race condition between `vkAcquireNextImageKHR` and `vkQueueSubmit` where the image may not have been "acquired" (and therefore may still be in use by the device) before new draws were made to it.

This is a minimal fix to ensure that the race doesn't exist. Ideally we eventually change these waits to happen in the "correct" places (e.g. the `presentCompletionGraphicsFence` should be waited just before `vkQueueSubmit` and `executeGraphicsFence` should be waited just before `vkQueuePresentKHR`). Optionally this could be done using the semaphores instead of full fences.


RubyNova (approved):
LGTM hopefully we stop seeing vsync tearing now

Issue #444, opened 2022-05-16
ResourceLoader has protected internal methods that should actually just be public overloads of their std::string counterparts.


RubyNova:
This shouldn't change much for the actual implementation of resource management plugins. It's just an accessor and method signature refactor so that the `Internal` methods are just an abstract overload.

You can find the specified methods defined in `include/NovelRT/ResourceManagement/ResourceLoader.h` under the `protected` access modifier.

This is a fairly straightforward issue to fix and is a good first issue if you are unfamiliar with the project and wish to contribute.

PR #475, opened 2022-07-06
Make internal methods public in the Resourceloader

RyadaProductions:
This PR does a few things:

- Make internal methods public and rename them to not have a suffix anymore
- Remove the old public methods that would implicitly use a preset folder
- The implementation of the abstract methods now prefixes the default folder if somebody uses a relative path

This makes the UX cleaner by only having 1 method instead of an overload and with added documentation is clear and easy to use.

Would solve Issue #444

### DIFF
--- a/include/NovelRT/ResourceManagement/ResourceLoader.h
+++ b/include/NovelRT/ResourceManagement/ResourceLoader.h
@@ -15,11 +15,6 @@ namespace NovelRT::ResourceManagement
     protected:
         std::filesystem::path _resourcesRootDirectory = Utilities::Misc::getExecutableDirPath() / "Resources";
 
-        [[nodiscard]] virtual TextureMetadata LoadTexture(
-            std::filesystem::path filePath) = 0; // TODO: I've realised these should probably be overloads, not internal
-                                                 // methods. Monke brain. :(
-        [[nodiscard]] virtual std::vector<uint8_t> LoadShaderSourceInternal(std::filesystem::path filePath) = 0;
-
     public:
         [[nodiscard]] inline std::filesystem::path& ResourcesRootDirectory() noexcept
         {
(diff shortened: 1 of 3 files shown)


include/NovelRT/ResourceManagement/ResourceLoader.h
```diff
@@ -31,16 +26,27 @@ namespace NovelRT::ResourceManagement
...
-            return LoadTexture(_resourcesRootDirectory / "Images" /
-                               fileName); // TODO: This should probably either be textures or sprites, not image? Thonk.
-        }
-
-        [[nodiscard]] inline std::vector<uint8_t> LoadShaderSource(const std::string& fileName)
-        {
-            return LoadShaderSourceInternal(_resourcesRootDirectory / "Shaders" / fileName);
-        }
+        /**
+         * @brief Loads texture from file on given path.
```
RubyNova, line 30:
```suggestion
         * @brief Loads a texture from a file on a given path.
```

include/NovelRT/ResourceManagement/ResourceLoader.h
```diff
@@ -31,16 +26,27 @@ namespace NovelRT::ResourceManagement
...
+         *
+         * Path can both be relative as well as absolute.
+         * When using relative path it will look in the Resources + Images directory
```
RubyNova, line 33:
```suggestion
         * The path can be either relative or absolute.
         * When using a relative path it will look in the Resources/Images directory.
```

include/NovelRT/ResourceManagement/ResourceLoader.h
```diff
@@ -31,16 +26,27 @@ namespace NovelRT::ResourceManagement
...
+         *
+         * @param filePath Relative or absolute path to the texture
+         * @return TextureMetadata of the file that was requested to load
```
RubyNova, line 36:
```suggestion
         * @param filePath Relative or absolute path to the texture.
         * @returns TextureMetadata The texture data contained in the file.
```

include/NovelRT/ResourceManagement/ResourceLoader.h
```diff
@@ -31,16 +26,27 @@ namespace NovelRT::ResourceManagement
...
+         */
+        [[nodiscard]] virtual TextureMetadata LoadTextureFromFile(std::filesystem::path filePath) = 0;
+
+        /**
+         * @brief Loads shader from file on given path.
```
RubyNova, line 41:
```suggestion
         * @brief Loads shader from a file on a given path.
```

include/NovelRT/ResourceManagement/ResourceLoader.h
```diff
@@ -31,16 +26,27 @@ namespace NovelRT::ResourceManagement
...
+         *
+         * Path can both be relative as well as absolute.
+         * When using relative path it will look in the Resources + Shaders directory
```
RubyNova, line 44:
```suggestion
         * The path can be either relative or absolute.
         * When using a relative path it will look in the Resources/Shaders directory.
```

include/NovelRT/ResourceManagement/ResourceLoader.h
```diff
@@ -31,16 +26,27 @@ namespace NovelRT::ResourceManagement
...
+         *
+         * @param filePath Relative or absolute path to the texture
+         * @return TextureMetadata of the file that was requested to load
```
RubyNova, line 47:
```suggestion
         * @param filePath Relative or absolute path to the shader.
         * @returns std::vector<uint8_t> Shader data as a memory block that was contained in the file.
```

RubyNova (changes requested):
We're almost there. Just need some of the wording fixed.

I also have a concern about exceptions. I know that the abstracted method itself doesn't throw anything (it is abstract - duh) - but what if it can't find a file? Should we spacify that file not found exception is a required exception of all implementors?

I am unsure how we want to handle that.

RubyNova (approved):
LGTM! Enabling auto merge.

PR #477, opened 2022-07-08
Vendor OpenAL, Ogg, and Vorbis (ditched libsndfile)

capnkenny:
Self-titled. Can disregard branch name, originally was going to use libnyquist but floating point sound files weren't playing nice with OpenAL at all.

- Vendors Ogg + Vorbis
- Vendors OpenAL
- Fixes a few bugs found with Input, Audio
- Adds background colour changing via DefaultRenderingSystem
- Manually loads WAV files and Vorbis files
- Also replaced a few resources

Seeing as I ran into a few heap corruption errors (that were fixed before I opened the PR), I wouldn't mind volunteers to try this branch out with `AudioEcsSample` :D


FiniteReality:
I'm not a huge fan of us implementing the code needed to read/write these files ourselves. It opens us up to a bunch of security concerns that we need to handle correctly, and we can't use an upstream source to assist with that.

capnkenny:
@FiniteReality Based on our discussion, the most recent commits revert back to libsndfile for file handling

internal/libsndfile/v1.1.0.patch
```diff
@@ -0,0 +1,30 @@
...
+-		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:Vorbis::vorbisenc>
+-		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:FLAC::FLAC>
++		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:$<TARGET_LINKER_FILE:vorbisenc>>
++		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:$<TARGET_LINKER_FILE:FLAC>>
+ 		$<$<AND:$<BOOL:${ENABLE_EXPERIMENTAL}>,$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>,$<BOOL:${HAVE_SPEEX}>>:Speex::Speex>
+-		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:Opus::opus>
++		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:$<TARGET_LINKER_FILE:opus>>
+ 		$<$<BOOL:${HAVE_MPEG}>:MPG123::libmpg123>
+ 		$<$<BOOL:${HAVE_MPEG}>:Lame::Lame>
+ 	)
```
RubyNova, line 29: *bad dalek impression*

EXPLAIN?

internal/vulkan/vulkan-loader/CMakeLists.txt
```diff
@@ -26,7 +26,12 @@ if(NOT vulkan-loader_POPULATED)
   set(BUILD_SHARED_LIBS_OLD ${BUILD_SHARED_LIBS})
   # Make subproject to use 'BUILD_SHARED_LIBS=ON' setting.
   set(BUILD_SHARED_LIBS ON CACHE INTERNAL "Build SHARED libraries")
+  set(BUILD_WSI_WAYLAND_SUPPORT OFF CACHE INTERNAL "")
```
RubyNova, line 29: Will this break Wayland?

samples/AudioEcsSample/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+        main.cpp
+        )
+
+add_executable(AudioEcsSample ${SOURCES})
+add_dependencies(AudioEcsSample Resources)
+target_link_libraries(AudioEcsSample
+        PRIVATE
+        Engine
+        )
+target_compile_definitions(AudioEcsSample PUBLIC -DNOVELRT_MOLTENVK_VENDORED=$<IF:$<BOOL:${NOVELRT_MOLTENVK_VENDORED}>,1,0>)
```
RubyNova, line 15: We really should open an issue about this and figure out a fix

samples/AudioEcsSample/main.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+int main()
+{
+    NovelRT::LoggingService logger = NovelRT::LoggingService();
+
+#if NOVELRT_MOLTENVK_VENDORED
+    auto icdPath = NovelRT::Utilities::Misc::getExecutablePath() / "MoltenVK_icd.json";
+    setenv("VK_ICD_FILENAMES", icdPath.c_str(), 0);
+    logger.logInfo("macOS detected - setting VK_ICD_FILENAMES to path: {}", icdPath.c_str());
+#endif
```
RubyNova, line 22: We should probably open an issue about this too tbh.

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -2,6 +2,7 @@
 // for more information.
 
 #include <NovelRT/Audio/Audio.h>
+#include <sndfile.h>
```
RubyNova, line 5: Is there a reason we moved it into here?

src/NovelRT/Audio/AudioService.cpp
```diff
@@ -69,10 +70,12 @@ namespace NovelRT::Audio
             return _noBuffer;
         }
 
-        std::vector<uint16_t> data;
+        std::vector<int16_t> data;
```
RubyNova, line 73: Why `int16_t` for a data buffer?

src/NovelRT/Ecs/Audio/AudioSystem.cpp
```diff
@@ -17,6 +17,18 @@ namespace NovelRT::Ecs::Audio
         _service->InitializeAudio();
     }
 
+    AudioSystem::AudioSystem(bool verbose)
```
RubyNova, line 20: Can this constructor throw?

src/NovelRT/Ecs/Graphics/DefaultRenderingSystem.cpp
```diff
@@ -255,7 +256,7 @@ namespace NovelRT::Ecs::Graphics
         ResolveGpuResourceCleanup();
         ResolveGpuFutures();
 
-        context->BeginDrawing(NovelRT::Graphics::RGBAColour(0, 0, 255, 255));
+        context->BeginDrawing(_backgroundColour);
```
RubyNova, line 259: Can we note in this PR that this is a supported feature now? For documentation purposes.

src/NovelRT/LoggingService.cpp
```diff
@@ -143,19 +143,19 @@ namespace NovelRT
...
+                spdlog::set_level(spdlog::level::level_enum::warn);
                 break;
             case LogLevel::Err:
-                _logger->set_level(spdlog::level::level_enum::err);
+                spdlog::set_level(spdlog::level::level_enum::err);
                 break;
             case LogLevel::Off:
-                _logger->set_level(spdlog::level::level_enum::off);
+                spdlog::set_level(spdlog::level::level_enum::off);
                 break;
```
RubyNova, line 159: What happened here?

RubyNova (changes requested):
I have questions but otherwise - if it works then thank GOD for that

src/NovelRT/LoggingService.cpp
capnkenny: Ahhhhh was supposed to revert - i opened an issue about this

src/NovelRT/Ecs/Graphics/DefaultRenderingSystem.cpp
capnkenny: Ye I can

src/NovelRT/Ecs/Audio/AudioSystem.cpp
capnkenny: Maybe?

src/NovelRT/Audio/AudioService.cpp
capnkenny: I had switched it because the datatype returned by libsndfile _should_ only be signed. This can be switched if desired however - I believe it'll work with uint16_t as well

src/NovelRT/Audio/AudioService.cpp
capnkenny: I had moved it here so anything referring to the Audio header wasn't also pulling in everything related to sndfile

src/NovelRT/Audio/AudioService.cpp
RubyNova: Generally I use `uint8_t` which is what prompted the question since its just a byte buffer right?

If that's gonna cause breakages then just ignore this is just me being curious

samples/AudioEcsSample/main.cpp
capnkenny: This is one and the same with the above issue, if I understand your concern

samples/AudioEcsSample/CMakeLists.txt
capnkenny: I'll open the issue but am concerned there won't be a true "fix" for this

.github/workflows/build-system.yml
```diff
@@ -52,14 +52,6 @@ jobs:
...
           cache: true
 
-      - name: Install Vulkan SDK (Windows)
-        uses: humbletim/setup-vulkan-sdk@v1.2.0
-        if: "contains(matrix.os, 'windows')"
-        with:
-          vulkan-query-version: 1.3.211.0
-          vulkan-components: Vulkan-Headers, Vulkan-Loader
-          vulkan-use-cache: true
-
```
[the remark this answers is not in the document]
capnkenny: Currently CI has two ways of getting Vulkan SDK prior to CMaking anything - the Linux/macOS method which installs a full copy of the SDK to a portable location on the CI VM (which provides those OSes with an interface for Vulkan as well), or the Windows way which creates a quasi-SDK with only the required elements listed.

Since vending the Vulkan Loader and Headers does the _same_ that the CI action does here, I removed the Windows version to stop the LunarG website 502s from blocking us. Linux and macOS remain unchanged

CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif()
 
 if(WIN32)
   # this is needed, because Windows can be weird, man.
-  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS OFF)
```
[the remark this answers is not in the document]
capnkenny: I'll double check but I moved this explicitly to the Engine as a property so that not _everything under the sun_ exports its symbols. OpenAL and libsndfile were being a real headache about it :/

CMakeLists.txt
```diff
@@ -29,20 +29,15 @@ option(NOVELRT_BUILD_SAMPLES "Build NovelRT samples" ON)
 option(NOVELRT_BUILD_DOCUMENTATION "Build NovelRT documentation" ON)
 option(NOVELRT_BUILD_TESTS "Build NovelRT tests" ON)
 option(NOVELRT_BUILD_INTEROP "Build NovelRT's Interop library" ON)
-option(NOVELRT_INCLUDE_CONAN_PATHS "Include conan_paths toolchain" ON)
+option(NOVELRT_INCLUDE_CONAN_PATHS "Include conan_paths toolchain" OFF)
```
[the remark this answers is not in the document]
capnkenny: Yeeeeeee :D

conanfile.py
```diff
@@ -19,12 +18,7 @@ class NovelRTConan(ConanFile):
...
     default_options = {
-        "libsndfile:shared":True,
         "openal:shared":True,
-        "flac:shared":True,
-        "Opus:shared":True,
-        "Ogg:shared":True,
-        "Vorbis:shared":True,
         "documentation": False,
         "buildtests":True,
         "buildsamples":True,
```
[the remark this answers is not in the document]
capnkenny: Doesn't really matter, does it? Conan removal == separate PR (that we anxiously await)

include/NovelRT/Ecs/Audio/AudioSystem.h
```diff
@@ -25,6 +25,7 @@ namespace NovelRT::Ecs::Audio
 
     public:
         AudioSystem();
+        AudioSystem(bool verbose);
```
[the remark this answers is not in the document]
capnkenny: Will remove - doesn't do what I wanted it to due to the Logging Service bug

internal/vulkan/vulkan-loader/CMakeLists.txt
capnkenny: Wayland's not broken but disabled...and I don't know why this is here :D

internal/libsndfile/v1.1.0.patch
capnkenny: 1) libsndfile thinks that the Opus lib being provided is in an _installed_ location :D so we patch this to be explicitly compatible with the built version...cuz we build it lol.

2) Originally libsndfile wanted `vorbisenc.obj` which wasn't being provided by the `vorbisenc` target, so I patched it to specifically used whatever library is provided by the target instead of assuming..it lead to similar breakages so I changed the ones we're concerned with.

internal/vulkan/vulkan-loader/CMakeLists.txt
RubyNova: I uhhh

Don't really follow entirely

This sounds bad/wrong but I'm unsure

CMakeLists.txt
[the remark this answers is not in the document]
capnkenny: This needs to be done to Interop as well otherwise Windows breaks

internal/vulkan/vulkan-loader/CMakeLists.txt
capnkenny: Ah! So this broke stuff because then this is on (which it is by default), Wayland is required otherwise the build fails. Which we can add our own option to toggle this, however without it off by default building with WSL2/WSLg failed.

conanfile.py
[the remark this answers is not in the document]
RubyNova: Fair enough I just think its like

Dead code

I guess

Idk it just kinda ticked the "nit" box when I first saw it :^) feel free to ignore

internal/vulkan/vulkan-loader/CMakeLists.txt
RubyNova: OK we should definitely add a bug that Wayland's gonna commit not alive at the moment with a setup for this

We can add a switch in another PR

CMakeLists.txt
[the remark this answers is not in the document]
RubyNova: I think the default value for this is `OFF` fwiw

internal/vulkan/vulkan-loader/CMakeLists.txt
capnkenny: Created #478 for this

samples/AudioEcsSample/CMakeLists.txt
capnkenny: Opened #479 for tracking this

samples/AudioEcsSample/main.cpp
capnkenny: #479

src/NovelRT/Audio/AudioService.cpp
capnkenny: Yeah, based on sndfile providing `short` values to `data`, this would need to be `uint16_t`  o_or `int16_t` therwise there's the potential to lose data based on casting (Windows does not like using `uint8_t` here).

src/NovelRT/Ecs/Audio/AudioSystem.cpp
capnkenny: This ctor has been removed ,but the original _might_ throw since we make a call to the service (which throws)

CMakeLists.txt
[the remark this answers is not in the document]
capnkenny: Should be resolved on Windows now, Interop and Engine explicitly export symbols but everything else should not

src/NovelRT/Ecs/Audio/AudioSystem.cpp
RubyNova: Gotcha.

src/NovelRT/Audio/AudioService.cpp
FiniteReality: It sounds like this is sample buffer data,so maybe it should be called something like `sampleData` rather than just `data`.

src/NovelRT/Audio/AudioService.cpp
FiniteReality: Encapsulating dependencies is usually a good thing. I approve.

.github/workflows/build-system.yml
```diff
@@ -60,26 +60,20 @@ jobs:
...
         run: |
           ${{ matrix.c }} ${{ matrix.cxx }} cmake -S . -B build -DCMAKE_BUILD_TYPE=${{ matrix.configuration }}
-          cmake --build build --config ${{ matrix.configuration }}
-          cd build/tests/NovelRT.Tests
-          ctest --output-on-failure
+          cmake --build build --config ${{ matrix.configuration }} --target compiler Engine Interop Samples
 
-      - name: Install NovelRT - ${{ matrix.configuration }}
+      - name: Test NovelRT - ${{ matrix.configuration }}
         if: steps.build.outcome == 'success'
```
RubyNova, line 71: What actually happened here? Can't make sense of it

samples/Experimental/VulkanRender/CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-include(CopyBuildProducts)
```
RubyNova, line 1: Is this no longer required?

RubyNova (approved):
LGTM just some questions

src/NovelRT/Ecs/Graphics/DefaultRenderingSystem.cpp
RubyNova: I don't think this has been done yet? 😅

.github/workflows/build-system.yml
capnkenny: Removed the install step in CI and bumped it to only testing NovelRT after the engine, interop, samples, and Fabulist compiler are built

samples/CMakeLists.txt
```diff
@@ -1,6 +1,19 @@
...
+add_custom_target(Samples
+  DEPENDS
+    VulkanRender
+    EcsPipeline
+    InputEcsSample
+    PersistenceSample
+    UIEventExample
+    FabulistTest
+    AudioEcsSample
+)
```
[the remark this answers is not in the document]
capnkenny: Adds a generic "Samples" target to force building all the samples instead of one at a time

samples/Experimental/VulkanRender/CMakeLists.txt
capnkenny: Not here, it was a dependency for all samples so I moved it into the higher level Samples CMakeLists for accessibility

PR #484, opened 2022-07-27
Removed nonexistent includes from NovelRT.h

jeuxjeux20:
Basically, if you had defined the `NOVELRT_C_API` macro, NovelRT wouldn't compile! Which is a tragic event for fans of C, and binding maintainers. Now it's fixed. Interop fans rejoice!


RubyNova (approved):
LGTM

Issue #405, opened 2022-02-07
Missing helpers for NovelRT::Ecs::Graphics


RubyNova:
This ticket is mostly just a bit of bulk work, and should be doable by anyone who can understand basic C and basic C++.

There is currently no well-defined C shim for simplifying access to the default graphics support within NovelRT::Ecs. This means that:
- it is currently impossible to attach the default rendering system to a `SystemScheduler` from C
- it is currently impossible to get the correct component type ID for the rendering components
- even if it were possible, the barebones C API for ECS means it would be cumbersome to use.

For something that is well-defined as this in the C++ API I believe that we should probably support the above behaviours in C also. This may or may not involve:
- making C-compliant versions of POD types for type punning
- opaque handles to complex types such as the default rendering system
- making C functions to register the default rendering components internally via C++ and caching the IDs so they can be returned via another C function (or by the same one)

ActualDio:
Ill work on it

RubyNova:
Issue is open again after user was banned.

NullSapphire:
I'll start working on it

jeuxjeux20:
I'll work on it (trust me, I will!!)

Issue #388, opened 2022-01-17
memcpy_s unavailable on Linux. Defines not working properly.


RubyNova:
As per the title I believe our defines for methods such as memcpy_s are incorrectly placed.

In addition I believ the issue also affects Win32. This is a side effect of us moving away from a centralised header.

What's the best way to handle this?

RubyNova:
After a discussion on discord it has been decided we are to remove calls to `memcpy_s` and other window-specific "secure" versions, and just run validation checks before calling these methods ourselves.

This is not a super complicated fix I don't think, but it is a little involved.

GooStuffedDoll:
I'll take care of it.

RubyNova:
Ticket is available again. The issue is also relatively simple so not sure why it wasn't marked as a good first issue previously but yeah. Fixed now.

JorchRL:
I'll look into this one 😅😅😅

RubyNova:
Not heard anything in over a month. Assuming it is abandoned.

PR #485, opened 2022-08-02
Interop support for the Configurator API and all plugins

jeuxjeux20:
It's currently hard to get anything — such as a window, at least — popping on your screen with the C API. This PR implements the entire Configurator API and all of its plugins. Making your dreams of a game engine usable in C a reality.

### What's implemented

- **`NrtConfigurator`**: methods to add plugin providers, as well as default systems
- **All variants of `IPluginProvider`**: (Graphics, Input, ResourceManagement, Windowing)
- **`GraphicsProvider`**: barebones with only a `Destroy` method
- **`IWindowingDevice`**: with all methods (except `TearDown`), such as `ProcessMessages`, `GetShouldClose` and `WindowTitle` manipulation
- **`IInputDevice`**: with all methods (except `Initialise` and `Update`)
- **`ResourceLoader`**: only the `ResourceRootDirectory` property has been implemented

#### LifetimeExtender
The `AllObjects<T>` function uses a templated static variable for each type, which can maybe cause issues with multiple TUs. However, linkers should automatically choose only *one* static variable across multiple TUs: that behavior is called [vague linkage](https://gcc.gnu.org/onlinedocs/gcc/Vague-Linkage.html).

#### NrtConfigurator
The `AddXPluginProvider` methods use `Lifetime::Find` to get a `shared_ptr` of a plugin provider. This method uses linear search on all `shared_ptr`s created by the C API. It shouldn't be an issue considering we don't spawn thousands of PluginProviders, but it's something to note.

#### NrtInputActionList
I choose to implement a span of `NovelRT::Input::InputAction` instances with a struct that mimics an actual `span`:  
```cpp
typedef struct {
    NrtInputActionHandle begin;
    NrtInputActionHandle end;
    intptr_t increment; // sizeof(InputAction)
} NrtInputActionList;
```
Contrary to iterators, they demand more "discipline" to be used, but they are much more efficient for performance (no allocation, no need to call an externally-linked method every time), easier to implement (it's just one struct!) and it actually makes sense for a span.
The `intptr_t Nrt_InputActionList_IncrementSize()` method is there if you want to make it a constant. (it's also pretty cool for a JIT!)


src/NovelRT.Interop/Graphics/NrtGraphicsProvider.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    if (Lifetime::Release(reinterpret_cast<GraphicsProvider*>(provider)))
+    {
+        return NRT_SUCCESS;
+    }
+    else
+    {
+        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
+        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
+    }
```
RubyNova, line 28:
```suggestion
    if (!Lifetime::Release(reinterpret_cast<GraphicsProvider*>(provider)))
    {
        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
    }
    
    return NRT_SUCCESS;
```

src/NovelRT.Interop/LifetimeExtender.h
```diff
@@ -0,0 +1,57 @@
...
+            if ((*it).get() == object)
+            {
+                aliveObjects.erase(it);
+                return true;
+            }
+        }
+        return false;
+    }
+}
+
```
RubyNova, line 56: Since this is C++ please add the correct qualifiers. I can't rationalise if this throws or mutates or what. :D

src/NovelRT.Interop/LifetimeExtender.h
```diff
@@ -0,0 +1,57 @@
...
+    template<typename T> std::shared_ptr<T> Find(T* object)
+    {
+        for (auto& aliveObject : AliveObjects<T>())
+        {
+            if (aliveObject.get() == object)
+            {
+                return aliveObject;
+            }
+        }
+        return {nullptr};
```
RubyNova, line 39: does this not work?

```suggestion
        return nullptr;
```

src/NovelRT.Interop/LifetimeExtender.h
```diff
@@ -0,0 +1,57 @@
...
+    template<typename T> std::vector<std::shared_ptr<T>>& AliveObjects()
+    {
+        static std::vector<std::shared_ptr<T>> aliveObjects{};
+
+        return aliveObjects;
+    }
+
+    template<typename T> void KeepAlive(std::shared_ptr<T>&& object)
+    {
+        AliveObjects<T>().emplace_back(std::move(object));
```
RubyNova, line 22: I think you might need `std::forward` not `std::move` here.

src/NovelRT.Interop/PluginManagement/NrtDefaultPluginSelector.cpp
```diff
@@ -0,0 +1,128 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputGraphicsPlugin == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& graphicsPlugin = reinterpret_cast<DefaultPluginSelector*>(pluginSelector)
```
RubyNova, line 59: Out of curiosity, what does the `auto&&` achieve here as opposed to just `auto`?

src/NovelRT.Interop/PluginManagement/NrtDefaultPluginSelector.cpp
```diff
@@ -0,0 +1,128 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputWindowingPlugin == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& windowingPlugin = reinterpret_cast<DefaultPluginSelector*>(pluginSelector)
```
RubyNova, line 109: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/PluginManagement/NrtDefaultPluginSelector.cpp
```diff
@@ -0,0 +1,128 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputInputPlugin == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& inputPlugin = reinterpret_cast<DefaultPluginSelector*>(pluginSelector)
```
RubyNova, line 84: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/PluginManagement/NrtDefaultPluginSelector.cpp
```diff
@@ -0,0 +1,128 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputResourceManagementPlugin == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& resourcePlugin = reinterpret_cast<DefaultPluginSelector*>(pluginSelector)
```
RubyNova, line 34: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/PluginManagement/NrtIGraphicsPluginProvider.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    if (Lifetime::Release(reinterpret_cast<IGraphicsPluginProvider*>(plugin)))
+    {
+        return NRT_SUCCESS;
+    }
+    else
+    {
+        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
+        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
+    }
```
RubyNova, line 28:
```suggestion
    if (!Lifetime::Release(reinterpret_cast<IGraphicsPluginProvider*>(plugin)))
    {
        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
    }
    
    return NRT_SUCCESS;
```

src/NovelRT.Interop/PluginManagement/NrtIGraphicsPluginProvider.cpp
```diff
@@ -0,0 +1,52 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputProvider == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& graphicsProvider = reinterpret_cast<IGraphicsPluginProvider*>(plugin)->GetGraphicsProvider();
```
RubyNova, line 46: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/PluginManagement/NrtIInputPluginProvider.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    if (Lifetime::Release(reinterpret_cast<IInputPluginProvider*>(plugin)))
+    {
+        return NRT_SUCCESS;
+    }
+    else
+    {
+        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
+        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
+    }
```
RubyNova, line 28:
```suggestion
    if (!Lifetime::Release(reinterpret_cast<IInputPluginProvider*>(plugin)))
    {
        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
    }
    
    return NRT_SUCCESS;
```

src/NovelRT.Interop/PluginManagement/NrtIInputPluginProvider.cpp
```diff
@@ -0,0 +1,50 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputDevice == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& inputDevice = reinterpret_cast<IInputPluginProvider*>(plugin)->GetInputService();
```
RubyNova, line 45: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/PluginManagement/NrtIResourceManagementPluginProvider.cpp
```diff
@@ -0,0 +1,51 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputLoader == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& resourceLoader = reinterpret_cast<IResourceManagementPluginProvider*>(plugin)->GetResourceLoader();
```
RubyNova, line 46: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/PluginManagement/NrtIResourceManagementPluginProvider.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    if (Lifetime::Release(reinterpret_cast<IResourceManagementPluginProvider*>(plugin)))
+    {
+        return NRT_SUCCESS;
+    }
+    else
+    {
+        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
+        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
+    }
```
RubyNova, line 28:
```suggestion
    if (!Lifetime::Release(reinterpret_cast<IResourceManagementPluginProvider*>(plugin)))
    {
        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
    }
    
    return NRT_SUCCESS;
```

src/NovelRT.Interop/PluginManagement/NrtIWindowingPluginProvider.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    if (Lifetime::Release(reinterpret_cast<IWindowingPluginProvider*>(plugin)))
+    {
+        return NRT_SUCCESS;
+    }
+    else
+    {
+        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
+        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
+    }
```
RubyNova, line 28:
```suggestion
    if (!Lifetime::Release(reinterpret_cast<IWindowingPluginProvider*>(plugin)))
    {
        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
    }

    return NRT_SUCCESS;
```

src/NovelRT.Interop/PluginManagement/NrtIWindowingPluginProvider.cpp
```diff
@@ -0,0 +1,50 @@
...
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
+    }
+
+    if (outputDevice == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto&& windowingDevice = reinterpret_cast<IWindowingPluginProvider*>(plugin)->GetWindowingDevice();
```
RubyNova, line 45: Out of curiosity, what does the auto&& achieve here as opposed to just auto?

src/NovelRT.Interop/ResourceManagement/NrtResourceLoader.cpp
```diff
@@ -0,0 +1,76 @@
...
+    auto& path = reinterpret_cast<ResourceLoader*>(loader)->ResourcesRootDirectory();
+    auto pathString = path.string();
+
+#ifdef WIN32
+    size_t length = pathString.length() + 1;
+    *outputPath = static_cast<char*>(malloc(length));
+    strcpy_s(*outputPath, length, pathString.c_str());
+#else
+    *outputPath = strdup(pathString.c_str());
+#endif
```
RubyNova, line 35: We are no longer using the _s variants. Please put in the explicit checks required to use the function that exists in all places for what you require.

src/NovelRT.Interop/ResourceManagement/NrtResourceLoader.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    if (Lifetime::Release(reinterpret_cast<ResourceLoader*>(loader)))
+    {
+        return NRT_SUCCESS;
+    }
+    else
+    {
+        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
+        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
+    }
```
RubyNova, line 75:
```suggestion
    if (!Lifetime::Release(reinterpret_cast<ResourceLoader*>(loader)))
    {
        Nrt_setErrMsgIsAlreadyDeletedOrRemovedInternal();
        return NRT_FAILURE_ALREADY_DELETED_OR_REMOVED;
    }

    return NRT_SUCCESS;
```

src/NovelRT.Interop/Windowing/NrtIWindowingDevice.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    auto title = reinterpret_cast<IWindowingDevice*>(device)->GetWindowTitle();
+
+#ifdef WIN32
+    size_t size = strlen(title.c_str()) + 1;
+    char* result = reinterpret_cast<char*>(malloc(size));
+    strcpy_s(result, size, title.c_str());
+#else
+    char* result = strdup(title.c_str());
+#endif
```
RubyNova, line 51: We are no longer using the _s variants. Please put in the explicit checks required to use the function that exists in all places for what you require.

RubyNova (changes requested):
Overall a good first pass, definitely is going to require some adjustments but nothing catastrophic, also a few questions from me on some of the bits and pieces.

include/NovelRT.Interop/Graphics/NrtGraphicsTypedefs.h
```diff
@@ -1,7 +1,25 @@
...
+#define NOVELRT_INTEROP_GRAPHICS_NRTGRAPHICSTYPEDEFS_H
+
+#ifdef __cplusplus
+extern "C"
 {
-    uint8_t r;
-    uint8_t g;
-    uint8_t b;
-    uint8_t a;
-} NrtRGBAColour;
```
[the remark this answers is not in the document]
jeuxjeux20: Just added an include guard and another type, to make it ressemble other `NrtXTypedefs.h` files.

include/NovelRT.Interop/Windowing/NrtIWindowingDevice.h
```diff
@@ -0,0 +1,25 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    NrtResult Nrt_IWindowingDevice_Destroy(NrtIWindowingDeviceHandle device);
+    NrtBool Nrt_IWindowingDevice_GetIsVisible(NrtIWindowingDeviceHandle device);
+    NrtBool Nrt_IWindowingDevice_GetShouldClose(NrtIWindowingDeviceHandle device);
+    // The returned string is allocated, remember to free it!
+    char* Nrt_IWindowingDevice_GetWindowTitle(NrtIWindowingDeviceHandle device);
```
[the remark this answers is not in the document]
jeuxjeux20: It's quite hard currently to have consistent `char*` string allocation conventions, since `IWindowingDevice::GetWindowTitle` returns an `std::string` which inevitably copies the string, but other APIs don't return a string by copy and instead return it by reference (such as `Nrt_getLastError`).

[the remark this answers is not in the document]
RubyNova: Possible bug in C++ then?

src/NovelRT.Interop/Ecs/NrtConfigurator.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    if (provider == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto* pluginProvider = reinterpret_cast<IGraphicsPluginProvider*>(provider);
+    auto providerSharedPtr = Lifetime::Find(pluginProvider); // We assume they didn't destroy it.
```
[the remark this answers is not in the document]
jeuxjeux20: Well, I should probably add checks in there...

src/NovelRT.Interop/Ecs/NrtConfigurator.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    if (outputSystemScheduler == nullptr)
+    {
+        Nrt_setErrIsNullArgProvidedInternal();
+        return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
+    }
+
+    auto* configuratorPtr = reinterpret_cast<Configurator*>(configurator);
+    // Move that scheduler to the heap.
+    auto* systemScheduler = new SystemScheduler{configuratorPtr->InitialiseAndRegisterComponents()};
```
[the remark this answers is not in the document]
jeuxjeux20: This should "move" the created scheduler into the heap, using the default move constructor. I don't think there's any issue there (Also, works on my machine)

src/NovelRT.Interop/LifetimeExtender.h
jeuxjeux20: Isn't `std::forward` reserved for perfect forwarding? For cases where you have `T&&` in a template (and not `std::shared_ptr<T>&&`).

Here, `std::move` just conducts the moving operation, because `emplace_pack(object)` would still consider `object` as an l-value.

src/NovelRT.Interop/PluginManagement/NrtDefaultPluginSelector.cpp
jeuxjeux20: Nothing much, using `auto` would lead to the same behavior. I liked `auto&&` because it sort of says that "this value is temporary and will be moved".

src/NovelRT.Interop/LifetimeExtender.h
RubyNova: Can't you perfect forward the pointer here? its a `T&&`

src/NovelRT.Interop/ResourceManagement/NrtResourceLoader.cpp
jeuxjeux20: Wouldn't it be better to create a function in `NrtMisc.h` containing this piece of code (with strcpy) for copying strings?

src/NovelRT.Interop/LifetimeExtender.h
RubyNova: I don't have time to check that - can emplace throw? can erase throw? I'm mostly trusting you to get it right here

jeuxjeux20: Nothing throws!

src/NovelRT.Interop/LifetimeExtender.h
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    template<typename T> void KeepAlive(const std::shared_ptr<T>& object) noexcept
+    {
+        AliveObjects<T>().emplace_back(object);
+    }
+
+    template<typename T> [[nodiscard]] std::shared_ptr<T> Find(T* object) noexcept
```
RubyNova, line 30: Can't this be `const`?

RubyNova (changes requested):

src/NovelRT.Interop/LifetimeExtender.h
jeuxjeux20: `WithPluginProvider` requires a non-const `std::shared_ptr<T>` so I don't think so.

RubyNova: uh, incorrect. The function does not mutate `this`. This method can be `const`. I'm pretty sure.
```suggestion
    template<typename T> [[nodiscard]] std::shared_ptr<T> Find(T* object) const noexcept
```

jeuxjeux20: But it's a method contained outside of a class, so it's not possible?

From C++ reference:
> cv - const/volatile qualification, only allowed in non-static member function declarations

RubyNova: I see. I thought these members were in a class.

How come they aren't?

jeuxjeux20: Thought that making a class would be overkill, what's the point of making a fully static class when C++ has functions with namespaces?

RubyNova: It's probably my C# brain talking. I'm not sure if there's a benefit.

src/NovelRT.Interop/Input/NrtInputAction.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+    auto* inputAction = reinterpret_cast<InputAction*>(action);
+    return NrtInputActionInfo{inputAction->actionName.c_str(),
+                              Nrt_NovelKey_ReadAsInfo(reinterpret_cast<NrtNovelKeyHandle>(&inputAction->pairedKey)),
+                              static_cast<NrtKeyState>(inputAction->state)};
+}
+intptr_t Nrt_InputActionList_IncrementSize()
+{
+    return sizeof(InputAction);
+}
```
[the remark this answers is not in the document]
capnkenny: If this were to be a `constexpr` function in C++, I believe it would become compile-time and therefore work as a constant (since `sizeof` is considered compile-time as well), no? Without that I agree - it'd be a runtime function and wouldn't provide the same benefit as expected.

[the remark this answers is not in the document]
RubyNova: This is C

constexpr does not exist

:^)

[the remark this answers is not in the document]
RubyNova: Also P/Invoke will not care for your constexpr :D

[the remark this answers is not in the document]
capnkenny: Well its getting exposed as C, but its a c++ function 😛 (otherwise this file should end in `.c`)
Anyways if C# isn't going to care for it (how would Java then???), then either way it does appear to be moot

include/NovelRT.Interop/Windowing/NrtIWindowingDevice.h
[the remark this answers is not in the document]
capnkenny: Maybe? In places that we have a `std::string`, i would thing we should be returning using `.c_str()` which gives us a `const char*`...

It's also odd because we don't use it much, but the entirety of the error handling API calls upon pre-defined `const char*`s

src/NovelRT.Interop/Input/NrtInputAction.cpp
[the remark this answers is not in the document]
RubyNova: You can't make a C compliant function constexpr even if it is internally C++

src/NovelRT.Interop/ResourceManagement/NrtResourceLoader.cpp
capnkenny: Not sure when we decided this, @RubyNova ? It's still used in two other places and I couldn't find history of it - please open an issue to cover the spots where this is still used if that's the case?

src/NovelRT.Interop/LifetimeExtender.h
capnkenny: Considering that this appears to be used across a few classes, I think this makes sense, in which case `const` would fall over at compile-time since there's no class for it.

@RubyNova if you're cool with it being outside of a class, then I think this is fine.

capnkenny:
Note: may need to squash commits upon approval

src/NovelRT.Interop/ResourceManagement/NrtResourceLoader.cpp
RubyNova: We already have an open issue, see #388

src/NovelRT.Interop/LifetimeExtender.h
RubyNova: C API is not my code owner area primarily - this is why I pinged you in. Your say is the final one here I think.

include/NovelRT.Interop/Input/NrtNovelKey.h
```diff
@@ -0,0 +1,28 @@
...
+#include "../NrtTypedefs.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    NrtNovelKeyInfo Nrt_NovelKey_ReadAsInfo(NrtNovelKeyHandle key);
+
+    const char* Nrt_NovelKey_GetKeyName(NrtNovelKeyHandle key);
```
capnkenny, line 15: @RubyNova just to confirm, should we be directly outputting a `const char*` here? Or should we be returning a `NrtResult` and using a `const char**` parameter? We seem to use output parameters in other places but its not consistent.

include/NovelRT.Interop/Windowing/NrtIWindowingDevice.h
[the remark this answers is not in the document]
capnkenny: Also another place to point out - direct output here? Or should we shift to a `NrtResult` return with an output parameter?

src/NovelRT.Interop/ResourceManagement/NrtResourceLoader.cpp
capnkenny: Ah looks like someone's been assigned it already - cool beans

include/NovelRT.Interop/Ecs/Graphics/NrtDefaultRenderingSystem.h
```diff
@@ -0,0 +1,25 @@
...
+extern "C"
+{
+#endif
+
+    NrtResult Nrt_DefaultRenderingSystem_FindInScheduler(NrtSystemSchedulerHandle scheduler,
+                                                         NrtDefaultRenderingSystemHandle* outputResult);
+
+    NrtResult Nrt_DefaultRenderingSystem_ForceVertexTextureFutureResolution(NrtDefaultRenderingSystemHandle system);
+
+    // TODO: Implement GetOrLoadTexture, CreateSpriteEntity and CreateSpriteEntityOutsideOfSystem
```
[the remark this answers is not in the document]
RubyNova: This should be part of @NullSapphire 's work, see #405

src/NovelRT.Interop/Input/NrtInputAction.cpp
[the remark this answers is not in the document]
jeuxjeux20: > I think C# would need to define its own constant. This function is at runtime. I don't entirely follow how this would work.

That's the problem. C# *can't* guess the size of `InputAction`. This is why it can make it a sort-of constant:
```cs
public static readonly nuint Size = Nrt_InputActionList_IncrementSize();
```

include/NovelRT.Interop/Input/NrtNovelKey.h
RubyNova: We originally agreed for getter only noexcept methods (so, pure property getters) we'd just do this - but that was a long time ago. You, as code owner, might feel this is no longer a good idea; but I think this is how it works.

include/NovelRT.Interop/Ecs/Graphics/NrtDefaultRenderingSystem.h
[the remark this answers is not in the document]
capnkenny: Just confirming - ty!

src/NovelRT.Interop/Windowing/NrtIWindowingDevice.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+NrtBool Nrt_IWindowingDevice_GetShouldClose(NrtIWindowingDeviceHandle device)
+{
+    return reinterpret_cast<IWindowingDevice*>(device)->GetShouldClose();
+}
+
+char* Nrt_IWindowingDevice_GetWindowTitle(NrtIWindowingDeviceHandle device)
+{
+    auto title = reinterpret_cast<IWindowingDevice*>(device)->GetWindowTitle();
```
capnkenny, line 43: If this doesn't throw in C++ land, can we not use `.c_str()` to handle the allocation for us and return a `const char*` instead?

jeuxjeux20: That won't work. `GetWindowTitle` returns a `std::string` by value which copies the string onto the stack (this is what happens in this line). So using `c_str()` here will return a pointer to a string that's going to deleted as soon as the function ends.

src/NovelRT.Interop/Input/NrtInputAction.cpp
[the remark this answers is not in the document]
RubyNova: Gotcha. Thanks. :D

src/NovelRT.Interop/Windowing/NrtIWindowingDevice.cpp
RubyNova: Ah, now I remember.

I think this is going to require more thought, might be out of scope for this PR otherwise I think this change will get held up too long.

RubyNova:
Update for this - I've gone ahead and implemented the resource loader APIs. I don't know how it affects your plugin support, but please review it.

jeuxjeux20:
Update for this - I did the changes and adapted the existing code to the new APIs, the only thing missing is... the tests.

RubyNova:
If you can't test some of it that's fine, just test what you can, we need to get this moving.

RubyNova (approved):

PR #486, opened 2022-08-02
Improved interop for StepTimer with a new Event C API

jeuxjeux20:
Currently, the interop implementation for `NovelRT::Timing::StepTimer` is broken. Here are some problematic snippets: 
```cpp
NrtResult Nrt_StepTimer_create(uint32_t targetFrameRate, double maxSecondDelta, NrtStepTimerHandle* output)
{
    // null-check omitted
    NovelRT::Timing::StepTimer timer = NovelRT::Timing::StepTimer(targetFrameRate, maxSecondDelta); // It's in the stack...
    *output = reinterpret_cast<NrtStepTimerHandle>(&timer); // Returning a pointer to the stack?!?
    return NRT_SUCCESS;
}

NrtTimestamp Nrt_StepTimer_getTargetElapsedTime(NrtStepTimerHandle timer)
{
    NovelRT::Timing::Timestamp* time = new NovelRT::Timing::Timestamp(0); // Allocating an int64 to the heap
    *time = reinterpret_cast<NovelRT::Timing::StepTimer&>(timer).getTargetElapsedTime(); // That reference-cast seems wrong
    return reinterpret_cast<NrtTimestamp&>(*time); // That NrtTimestamp is now pointing to undestroyable heap memory!
}

NrtResult Nrt_StepTimer_setTargetElapsedTime(NrtStepTimerHandle timer, NrtTimestamp target)
{
    // null-check omitted
    NovelRT::Timing::StepTimer time = reinterpret_cast<NovelRT::Timing::StepTimer&>(timer); // Casting a handle as a reference
    time.setTargetElapsedTime(NovelRT::Timing::Timestamp(target)); // Segfault!  
    return NRT_SUCCESS;
}
```
Probably that `StepTimer` could previously fit in int64_t, which would explain the reference-casts. Anyway, that code does not work. That PR is going to fix this, and implements APIs for `NrtUtilitiesEvent` and `NrtUtilitiesEventWithTimestamp` to make fully usage of the `StepTimer`. Hopefully, C maniacs will now experience the niche pleasure of staring at a timer ticking excessively fast.

### Additions and renovations

- The `StepTimer` interop implementation has been mostly rewritten
- Methods have been added to create `Event<>` and `Event<NovelRT::Timing::Timestamp>`, add event listeners, and invoke them in C.

### Implementation remarks
In the `NrtEvent.cpp` file, there are templated "generic" versions of methods and a `EventImplementationDetails` struct to easily add more event types in the future in necessary. (It's macro-free!)


include/NovelRT/Utilities/Event.h
```diff
@@ -3,6 +3,7 @@
 
 #include "../Atom.h"
 #include <functional>
+#include <vector>
```
jeuxjeux20, line 6: Added this as an attempt to fix a probable issue with macOS. But CI randomly decided to *not* run the macOS pipeline...

include/NovelRT/Timing/StepTimer.h
```diff
@@ -5,6 +5,7 @@
 // Original code is available under the MIT Licence
 
 #include "Timestamp.h"
+#include "../Utilities/Event.h"
```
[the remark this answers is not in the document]
RubyNova: What does this do?

include/NovelRT.Interop/Utilities/NrtEvent.h
```diff
@@ -0,0 +1,46 @@
...
+
+    NrtResult Nrt_EventWithTimestamp_Invoke(NrtUtilitiesEventWithTimestampHandle event, NrtTimestamp timestamp);
+
+    NrtResult Nrt_EventWithTimestamp_Destroy(NrtUtilitiesEventWithTimestampHandle event);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // NOVELRT_INTEROP_UTILITIES_NRTEVENT_H
```
RubyNova, line 46: I'm not sure this is how we agreed to wrap events long term. CC @capnkenny do you remember what we said about this way back when?

src/NovelRT.Interop/Timing/NrtStepTimer.cpp
```diff
@@ -3,150 +3,138 @@
 
 #include <NovelRT.Interop/NrtErrorHandling.h>
 #include <NovelRT.Interop/Timing/NrtStepTimer.h>
-#include <NovelRT/NovelRT.h>
+#include <NovelRT/Timing/StepTimer.h>
+#include <NovelRT/Utilities/Event.h> // The order is important for some reason.
```
RubyNova, line 7: Can you clarify the comment here?

src/NovelRT.Interop/Timing/NrtStepTimer.cpp
```diff
@@ -3,150 +3,138 @@
...
+        Nrt_setErrIsNullInstanceProvidedInternal();
+        return NRT_FAILURE_NULL_INSTANCE_PROVIDED;
     }
 
-#ifdef __cplusplus
+    auto* time = reinterpret_cast<NovelRT::Timing::StepTimer*>(timer);
+    auto* eventWithTimestamp = reinterpret_cast<NovelRT::Utilities::Event<NovelRT::Timing::Timestamp>*>(event);
+    time->tick(*eventWithTimestamp);
+    return NRT_SUCCESS;
 }
```
RubyNova, line 140: I do wonder if we should be fixing the casing on the function names here since we're updating the C API but maybe that should be a separate issue.

include/NovelRT/Utilities/Event.h
jeuxjeux20: Works now!

include/NovelRT/Timing/StepTimer.h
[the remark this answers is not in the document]
jeuxjeux20: `StepTimer.h` depends on `Event.h`. Without this, including `StepTimer.h` alone doesn't work.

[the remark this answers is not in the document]
RubyNova: ...how on earth was this working previously? Can you figure that out?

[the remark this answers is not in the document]
jeuxjeux20: It worked only with of the main `NovelRT.h` header, since it included headers in the right order (Event first, Timer second): https://github.com/novelrt/NovelRT/blob/main/include/NovelRT/NovelRT.h#L103

src/NovelRT.Interop/Timing/NrtStepTimer.cpp
jeuxjeux20: I left that out by accident, so it's gone now.

include/NovelRT/Timing/StepTimer.h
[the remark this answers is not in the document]
RubyNova: I see. Thanks.

include/NovelRT.Interop/Utilities/NrtEvent.h
capnkenny: This looks a bit cleaner than what we initially anticipated since we were going to explicitly wrap each defined type as an `Nrt` opaque handle...

With that said, @jeuxjeux20  we'll need unit tests added to ensure that `Nrt_Event_*` and friends work properly as our original concern was that since it could be of type `T`, C obviously does not have a very good way of handling this (although your solution looks to do just that).

src/NovelRT.Interop/Timing/NrtStepTimer.cpp
capnkenny: Separate issue, but good callout :)

RubyNova (commented):
How is the templated implementation working for the C-compliant function signatures here?

src/NovelRT.Interop/Timing/NrtStepTimer.cpp
capnkenny: Created #547 for this

include/NovelRT.Interop/Utilities/NrtEvent.h
capnkenny: Created #548

jeuxjeux20:
Almost ready to merge! (let's wait for CI, and please squash though)

capnkenny (approved):

capnkenny:
No squash. We'd like to preserve history as-is (even if there's large gaps in history).

Good stuff!

PR #490, opened 2022-08-06
WIP: Refactor CMake usage w/ installation targets

capnkenny:
STILL WORK IN PROGRESS - mainly for review/approval of general changes before continuing work

This is a refactor of how we are handling dependencies and most Platform-specific changes, as well as a way for us to allow installation of NovelRT via `cmake --install .`

Failures are expected at this time seeing as I have not tested Linux or macOS


CMakeLists.txt
```diff
@@ -1,6 +1,35 @@
-cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
 
 cmake_policy(SET CMP0091 NEW)
```
FiniteReality, line 3: We should add a comment explaining what these are (something like `SET CMP00091 NEW) # Select MSVC runtime library flags using an abstraction`) just for future reference.

CMakeLists.txt
```diff
@@ -40,17 +58,40 @@ set(NOVELRT_VULKAN_VERSION "1.3.211" CACHE STRING "Vulkan version")
 set(NOVELRT_VULKAN_MEMORY_ALLOCATOR_VERSION "2.3.0" CACHE STRING "VulkanMemoryAllocator version")
 set(NOVELRT_VULKAN_SDK_PATH "" CACHE STRING "Directory for NovelRT to find the Vulkan SDK (mainly for macOS support)")
 
-find_package(Doxygen ${NOVELRT_DOXYGEN_VERSION}
-  COMPONENTS dot
-)
+if(NOVELRT_PIE)
+  set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
```
FiniteReality, line 62: We should use the POSITION_INDEPENDENT_CODE property instead of a global CMake variable: https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html

CMakeLists.txt
```diff
@@ -40,17 +58,40 @@ set(NOVELRT_VULKAN_VERSION "1.3.211" CACHE STRING "Vulkan version")
...
+endif()
 
-if(NOT DOXYGEN_FOUND)
-  message("Doxygen ${NOVELRT_DOXYGEN_VERSION} was not found, building documentation is disabled.")
-  set(NOVELRT_BUILD_DOCUMENTATION OFF)
+# Remove unused symbols in non-debug mode.
+# This is useful even with LTO (-84 KiB with MinSizeRel).
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+  # For some reason, adding to CMAKE_CXX_FLAGS results in a slightly smaller
+  # binary than using `add_compile/link_options`
```
FiniteReality, line 69: I believe CMake actually makes an `install/strip` target if you have install targets in your build script, which should automatically do this: https://cmake.org/pipermail/cmake/2013-March/054021.html

CMakeLists.txt
```diff
@@ -40,17 +58,40 @@ set(NOVELRT_VULKAN_VERSION "1.3.211" CACHE STRING "Vulkan version")
...
+
+  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${_extra_flags}")
+  set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} ${_extra_flags}")
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} ${_extra_flags}")
+  set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} ${_extra_flags}")
+  set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} ${_extra_flags}")
+  set(CMAKE_C_FLAGS_MINSIZEREL "${CMAKE_C_FLAGS_MINSIZEREL} ${_extra_flags}")
 endif()
 
+set(CMAKE_CXX_STANDARD 17)
```
FiniteReality, line 86: These should use per-target properties:
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html
https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html
https://cmake.org/cmake/help/latest/prop_tgt/EXPORT_COMPILE_COMMANDS.html

FiniteReality (commented):
I can't really say I'm a fan, but if it gets things building then I can't really argue.

capnkenny:
Curious, is there a way you would rather see it structured? I also wanted to revisit `find_package` as it seems that it would make things easier for installation as well

FiniteReality:
> Curious, is there a way you would rather see it structured? I also wanted to revisit `find_package` as it seems that it would make things easier for installation as well

Ideally, I'd have a structure something like this:
- Build dependencies externally and install them somewhere
- Pass this location to CMake when building
- Use find_package to locate them
- Use target properties where possible
- Use `install/strip` to strip symbols and install

This'll still work with the vendoring approach, but we'll have to find a way to drive the work of building them externally somehow.

RubyNova:
@FiniteReality would this allow us to just use `find_package` in all cases and put stuff on a switch or two? Like, I'm thinking in the case of where we will be using WIC on windows but libpng on Linux, but still providing the option to use libpng on windows if people want etc.

FiniteReality:
> @FiniteReality would this allow us to just use `find_package` in all cases and put stuff on a switch or two? Like, I'm thinking in the case of where we will be using WIC on windows but libpng on Linux, but still providing the option to use libpng on windows if people want etc.

More or less, yeah. You'd have a bunch of feature flags (e.g. `NOVELRT_USE_WIC` which is dependent on `WIN32` and defaults to `ON`) and then you'd select WIC or libpng.

PR #494, opened 2022-09-03
Add support for entity mapping for component instruction state updating.

RubyNova:


capnkenny (approved):
Don't see anything glaring at me - LGTM

PR #495, opened 2022-09-06
Add support for GUID binding to relative UNIX-style file paths for asset tracking.

RubyNova:


include/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.h
```diff
@@ -15,8 +15,12 @@ namespace NovelRT::ResourceManagement::Desktop
     private:
         LoggingService _logger;
 
+    protected:
+        void WriteAssetDatabaseFile() final;
+        void LoadAssetDatabaseFile() override;
+
     public:
-        [[nodiscard]] TextureMetadata LoadTextureFromFile(std::filesystem::path filePath) final;
+        [[nodiscard]] TextureMetadata LoadTexture(std::filesystem::path filePath) final;
```
capnkenny, line 23: Is this to be an example for future file loading for resources? More of a curiosity question

src/NovelRT/ResourceManagement/ResourceLoader.cpp
```diff
@@ -0,0 +1,44 @@
...
+                guid = gen();
+            }
+
+            guidPathMap.emplace(guid, filePath);
+            pathGuidMap.emplace(filePath, guid);
+        }
+
+        WriteAssetDatabaseFile();
+    }
+}
```
capnkenny, line 44: Nit: newline

include/NovelRT/Utilities/Misc.h
```diff
@@ -60,6 +60,25 @@ namespace NovelRT::Utilities
...
+            while (end != std::string::npos)
+            {
+                returnVec.emplace_back(targetString.substr(start, end - start));
+                start = end + delimiter.size();
+                end = targetString.find(delimiter, start);
+            }
+
+            returnVec.emplace_back(targetString.substr(start)); // THIS LINE ANGERS ME
+
+            return returnVec;
```
capnkenny, line 80: is this safe just because it's from a static method? (Meaning that we don't lose it on the stack)

capnkenny (changes requested):
Generally looks okay - what will be generating the Asset file, though? Is it to be generated by the Engine? Or is it to be hand-made first, and then eventually loaded?

include/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.h
RubyNova: This was more to bring it in line with the other resource loading methods, it looked weird. :D

capnkenny: No worries - Audio is currently considered a Resource but the loading method is not here, so this was more of a "do we have a future TODO" q

include/NovelRT/Utilities/Misc.h
capnkenny: yes

RubyNova: why wouldn't it be safe? a vector is not a span that's having its underlying memory being yanked. Aren't we already doing this in other utility methods?

capnkenny: I just forgot that this is copying value and not returning a ref 🙃 sorry about that!

capnkenny (approved):

PR #496, opened 2022-09-06
Propagate database handles.

RubyNova:


FiniteReality (approved):
LGTM from skimming through

PR #498, opened 2022-09-07
Add rendering persistence rule. Add location for default persistence rules.

RubyNova:
This PR solves two problems - the first one is a permanent solution, the second is just a duct tape-style fix until I can figure out something better for the current structure of the namespaces.

The main issue this solves is adding a custom serialisation rule for `NovelRT::Ecs::Graphics::RenderingComponent` so that it can be packaged as a `BinaryPackage` correctly.

The secondary duct tape hotfix is adding a location for default persistence rules. This will be replaced in the future when I come up with a better design, but for now, I don't care enough.


include/NovelRT/Persistence/Graphics/RenderingComponentPersistenceRule.h
```diff
@@ -0,0 +1,35 @@
...
+    private:
+        std::shared_ptr<Ecs::Graphics::DefaultRenderingSystem> _renderingSystem;
+
+    public:
+        explicit RenderingComponentPersistenceRule(
+            std::shared_ptr<Ecs::Graphics::DefaultRenderingSystem> renderingSystem) noexcept;
+
+        [[nodiscard]] inline size_t GetSerialisedSize() const noexcept final
+        {
+            return sizeof(uuids::uuid) * 3;
```
capnkenny, line 24: Magic number? Whats 3 for?

capnkenny (commented):
Just one question

include/NovelRT/Persistence/Graphics/RenderingComponentPersistenceRule.h
RubyNova: Its the size of the serialised data. It's 3 GUIDs.

capnkenny (approved):

PR #499, opened 2022-09-09
Add resource loading support for audio files.

RubyNova:


src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
```diff
@@ -327,4 +330,43 @@ namespace NovelRT::ResourceManagement::Desktop
...
+
+        const size_t _bufferSize = 2048;
+        SF_INFO info;
+        info.format = 0;
+        auto filePathString = filePath.string();
+        SNDFILE* file = sf_open(filePathString.c_str(), SFM_READ, &info);
+
+        if (file == nullptr)
+        {
+            throw NovelRT::Exceptions::FileNotFoundException(filePath.string());
```
capnkenny, line 349: Is this entirely true here? - the original logging would log the error message returned by sndfile (whether it was malformed file, wrong/unsupported format, etc) so we should probably do the same before throwing (or we include sndfile's error in the exception we throw)

capnkenny (changes requested):
LGTM bar one issue with an exception message

RubyNova:
This PR is blocked until I can add interop support for resource loading APIs.

capnkenny (approved):
LGTM

PR #503, opened 2022-09-29
Rework build system to generate installable NovelRT Core / Interop

### LINKED ISSUE #455: Test install targets for NovelRT
Given our move to FetchContent-driven dependencies we need to make sure installed copies of the engine still work.

### LINKED ISSUE #479: Vendored MoltenVK fails to find ICD without workaround
At this time, if a NovelRT user vendors MoltenVK for macOS builds, they will need to apply the following workaround for their applications to run:

`#if NOVELRT_MOLTENVK_VENDORED`

`auto icdPath = NovelRT::Utilities::Misc::getExecutablePath() / "MoltenVK_icd.json";`
`setenv("VK_ICD_FILENAMES", icdPath.c_str(), 0);`
`logger.logInfo("macOS detected - setting VK_ICD_FILENAMES to path: {}", icdPath.c_str());`

`#endif`

This is pretty unsightly, but it works to allow MoltenVK to search the proper path and find the ICD file. However, it needs to be done for every sample or other application using NovelRT as a library lest we make macOS throw exceptions due to not initialising Vulkan. 

We should find an alternative solution to this that would be more portable and cleaner if we can.

capnkenny:
This PR accomplishes the following:

- Forces the scaffolding of version-specific dependencies prior to building (as part of the configuration step)
- Installs the dependencies to a build directory specific folder
- Builds NovelRT, Interop, Samples, and Tests using the scaffolded dependencies
- Allows installing NovelRT, Interop, and dependencies required

Also does the following:
- Switches the build system in CI to produce Release / Debug artifacts (not uploaded yet since we're not publishing)
- Creates debuggable OSX "App Bundles" for the samples (to mimic what an end product should resemble)
- Bumps a few dependency versions up - notably Vulkan SDK 1.3.231.1, as the MoltenVK support includes Vulkan 1.2 and better supported ICD searching (as part of an App Bundle)


.gitattributes
```diff
@@ -13,3 +13,4 @@
 *.ttf       binary
 *.ogg       binary
 *.wav       binary
+*.icns	    binary
```
RubyNova, line 16: Is this an alternative extension to .ico?

.github/workflows/build-system.yml
```diff
@@ -29,44 +30,49 @@ jobs:
...
-        exclude:
-        - os: macos-11
-          configuration: RelWithDebInfo
 
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
 
-      - name: Install Vulkan SDK (Linux / macOS)
-        uses: humbletim/install-vulkan-sdk@v1.1.1
```
RubyNova, line 44: Isn't this version too old?

CMakeLists.txt
```diff
@@ -1,11 +1,16 @@
 cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
 cmake_policy(SET CMP0091 NEW)
+if(POLICY CMP0135)
+  cmake_policy(SET CMP0135 NEW)
```
RubyNova, line 5: What's this do?

CMakeLists.txt
```diff
@@ -1,11 +1,16 @@
...
+endif()
+
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum macOS / OS X deployment version")
 
 project(NovelRT
-  VERSION 0.1.0
+  VERSION 1.0.0 #MVP Version
   DESCRIPTION "NovelRT game engine"
-  HOMEPAGE_URL "https://github.com/novelrt/NovelRT"
+  HOMEPAGE_URL "https://novelrt.dev"
```
RubyNova, line 13: Does this direct to an actual website?

CMakeLists.txt
```diff
@@ -17,60 +22,252 @@ if(MSVC)
...
-set(NOVELRT_GSL_VERSION "4.0.0" CACHE STRING "ms-gsl version")
+set(NOVELRT_FLAC_VERSION "1.3.4" CACHE STRING "FLAC version")
+set(NOVELRT_GLFW_VERSION "3.3.7" CACHE STRING "GLFW3 version")
+set(NOVELRT_GSL_VERSION "4.0.0" CACHE STRING "Microsoft.GSL version")
 set(NOVELRT_ONETBB_VERSION "2021.5.0" CACHE STRING "OneTBB version")
 set(NOVELRT_OPENAL_VERSION "1.21.1" CACHE STRING "OpenAL version")
+set(NOVELRT_OGG_VERSION "1.3.5" CACHE STRING "Ogg version")
+set(NOVELRT_OPUS_VERSION "1.3.1" CACHE STRING "Opus version")
 set(NOVELRT_PNG_VERSION "1.6.35" CACHE STRING "PNG version")
-set(NOVELRT_VULKAN_VERSION "1.3.211" CACHE STRING "Vulkan version")
```
RubyNova, line 39: What happened here?

cmake/CopyRuntimeDependencies.cmake
```diff
@@ -0,0 +1,186 @@
...
+                break()
+              endif()
+              if("${lib_idx}" MATCHES ".*::.*")
+                string(REGEX REPLACE ".*::" "" lib_safe_name "${lib_idx}")
+              else()
+                set(lib_safe_name "${lib_idx}")
+              endif()
+              list(APPEND dep_lib_snippet
+                "
+                if(EXISTS otool.txt)
```
RubyNova, line 42: What is all this doing?

src/NovelRT/Graphics/Vulkan/VulkanGraphicsProvider.cpp
```diff
@@ -242,6 +242,10 @@ namespace NovelRT::Graphics::Vulkan
         createInfo.ppEnabledLayerNames = allValidationLayerPtrs.data();
         createInfo.enabledLayerCount = static_cast<uint32_t>(validationLayerLength);
 
+#ifdef __APPLE__
+        createInfo.flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
```
RubyNova, line 246: Is this for making MoltenVK functional?

RubyNova, line 246: Is this for making MoltenVK functional?

RubyNova (commented):
Just questions I need answering, assuming the samples mostly work it LGTM

.gitattributes
capnkenny: Nope, this is a specific "Icon set" format for OSX/iOS (providing different copies of the icon at different res)

.github/workflows/build-system.yml
capnkenny: macOS vendors it with its own script, and Linux comes preinstalled from the docker image so this is no longer needed (although the version there matches the Action, not the SDK version 😄 )

CMakeLists.txt
capnkenny: This redirect's to the GitHub Organization page we have currently

CMakeLists.txt
capnkenny: This changes the behaviour when downloading via FetchContent to use the newest downloaded and extracted content when fetching for CMake 3.24+ (which visibly complains when it's not set)

CMakeLists.txt
capnkenny: It just got moved to line 58, and version bumped to 1.3.231.1

cmake/CopyRuntimeDependencies.cmake
capnkenny: _Heh._ :sans: 
Currently, this will:
- Run `otool` on the provided target library/app with `grep` to find the RPATH entry that was linked for the dependency
- Then it will use `install_name_tool` (IIRC, on mobile atm) and update said library/app with a new RPATH for the dependency

I did this so that on load, if _for some reason the samples get moved_, it knows where to look inside the App Bundles for the dependent libs. This is specifically Apple-only, Linux doesn't need this and Windows just uses CMake's support for identifying runtime DLLs.

src/NovelRT/Graphics/Vulkan/VulkanGraphicsProvider.cpp
capnkenny: Yes, as of 1.3.216 this is required otherwise MoltenVK will fail to opt-in to the portability standard and cause a segfault

capnkenny: Yup, same as last explanation about 1.3.216

RubyNova (approved):
Let's go

PR #505, opened 2022-10-02
Fix broken test that validated duplicate values upon binpackage load.

RubyNova:
This test was broken after the new featureset that actually managed duplicate entities was introduced to allow additive loading to work correctly.


capnkenny (approved):
LGTM

PR #512, opened 2022-10-16
Added IEcsSystem management in the Configurator class

### LINKED ISSUE #502: `NovelRT::Ecs::Configurator` appears to be missing `WithSystems` overload for `IEcsSystem` implementors.
The `Configurator` helper class appears to only support the classic `std::function<void(TimeStamp, Catalogue)>` signature at the moment, but the `SystemScheduler` has support for also registering `IEcsSystem` objects. We should extend the `Configurator` to support this new overload.

bjauny:
- [x] Tests for the changes have been added (for bug fixes / features)

This adds support for the `IEcsSystem` use in the `Configurator` class.

Fixes #502 .

`Configurator` only manages `std::function<void(TimeStamp, Catalogue)>` systems.

`Configurator` manages `IEcsSystem` systems.

No.

**Other information**:

### DIFF
--- a/include/NovelRT/Ecs/Configurator.h
+++ b/include/NovelRT/Ecs/Configurator.h
@@ -1,5 +1,3 @@
-#include <utility>
-
 // Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
 // for more information.
 
--- a/include/NovelRT/Ecs/Configurator.h
+++ b/include/NovelRT/Ecs/Configurator.h
@@ -1,5 +1,5 @@
-#include <utility>
 #include "IEcsSystem.h"
+#include <utility>
 
 // Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
 // for more information.
@@ -93,7 +93,7 @@ namespace NovelRT::Ecs
          * @param functions collection of IEcsSystem instances to attach.
          * @returns A reference to this to allow method chaining.
          */
-        [[nodiscard]] inline Configurator& WithIEcsSystems(
+        [[nodiscard]] inline Configurator& WithSystems(
             std::initializer_list<std::shared_ptr<IEcsSystem>>&& systems) noexcept
         {
             for (auto&& system : systems)
--- a/include/NovelRT/Ecs/Configurator.h
+++ b/include/NovelRT/Ecs/Configurator.h
@@ -1,4 +1,5 @@
 #include <utility>
+#include "IEcsSystem.h"
 
 // Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
 // for more information.
@@ -21,6 +22,7 @@ namespace NovelRT::Ecs
         bool _shouldAddDefaultSystems = false;
         std::optional<uint32_t> _threadCount;
         std::vector<std::function<void(Timing::Timestamp, Catalogue)>> _systems;
+        std::vector<std::shared_ptr<IEcsSystem>> _iecsSystems;
         std::shared_ptr<PluginManagement::IGraphicsPluginProvider> _graphicsPluginProvider;
         std::shared_ptr<PluginManagement::IWindowingPluginProvider> _windowingPluginProvider;
         std::shared_ptr<PluginManagement::IResourceManagementPluginProvider> _resourceManagementPluginProvider;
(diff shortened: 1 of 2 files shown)


bjauny:
This is a first and naive implementation.

Now that I reread the original issue, I think I should just rename the `WithIEcsSystems` to an overload of `WithSystems`. If so, I will add another test to manage the initalization with the overload instead of managing both in the current test.

RubyNova:
I will wait for your changes to do a review then to avoid duplication. :D

bjauny:
> I will wait for your changes to do a review then to avoid duplication. :D

All done!

include/NovelRT/Ecs/Configurator.h
```diff
@@ -1,3 +1,4 @@
+#include "IEcsSystem.h"
```
RubyNova, line 1: Please remove this, it has the potential to break due to how our headers are structured. If it's causing problems to remove I can probably walk you through how to fix it correctly.

Looks like there's an additional include here that should not exist and a missing namespace error preprocessor error we have in our headers, maybe we should fix that while we're here.

include/NovelRT/Ecs/Configurator.h
```diff
@@ -85,6 +87,23 @@ namespace NovelRT::Ecs
             return *this;
         }
 
+        /**
+         * @brief Configures systems for registration for this ECS instance (IEcs variant).
+         *
+         * @param functions collection of IEcsSystem instances to attach.
```
RubyNova, line 93: The parameter name is wrong here.

include/NovelRT/Ecs/Configurator.h
```diff
@@ -85,6 +87,23 @@ namespace NovelRT::Ecs
             return *this;
         }
 
+        /**
+         * @brief Configures systems for registration for this ECS instance (IEcs variant).
```
RubyNova, line 91:
```suggestion
         * @brief Configures systems for registration for this ECS instance (IEcsSystem variant).
```
We do this to allow better doxygen linkage.

tests/NovelRT.Tests/Ecs/ConfiguratorTest.cpp
```diff
@@ -40,6 +40,31 @@ TEST(ConfiguratorTest, ConfiguratorCanProduceSystemSchedulerWithCustomComponentA
...
+        {
+            auto intComponents = catalogue.GetComponentView<int32_t>();
+            for (auto [entity, component] : intComponents)
+            {
+                intComponents.PushComponentUpdateInstruction(entity, 10);
+            }
+        }
+    };
+
+    std::shared_ptr<TestSystem> tester = std::make_shared<TestSystem>();
```
RubyNova, line 58: This could just be `auto` given the fact you're calling make shared. No reason to have type information twice unless you're explicitly changing said type I don't think.

include/NovelRT/Ecs/Configurator.h
```diff
@@ -21,6 +22,7 @@ namespace NovelRT::Ecs
         bool _shouldAddDefaultSystems = false;
         std::optional<uint32_t> _threadCount;
         std::vector<std::function<void(Timing::Timestamp, Catalogue)>> _systems;
+        std::vector<std::shared_ptr<IEcsSystem>> _iecsSystems;
```
RubyNova, line 25: I do wonder of there's a better name for this collection but it does what it says on the tin and I can't think of anything better right now so best just leave it alone. Mostly just writing this for transparency in case other maintainers see this and have an idea.

RubyNova (changes requested):
Basically almost there, just some typos and admin stuff I think. :)

include/NovelRT/Ecs/Configurator.h
bjauny: Is it just a matter of including `Ecs.h` explicitely and remove the `#error`?

RubyNova: Remove the include, add the error define unless it's already in Configurator.h and I missed it?

tests/NovelRT.Tests/Ecs/ConfiguratorTest.cpp
bjauny: Done.

include/NovelRT/Ecs/Configurator.h
bjauny: Done.

include/NovelRT/Ecs/Configurator.h
bjauny: Done.

include/NovelRT/Ecs/Configurator.h
bjauny: Done. There is indeed a `#error` if the ECS define is not set, is that what you're expecting? The same can be found in 28 other files.

RubyNova: Yeah I see it now.

There's still an include we should remove here but I'll flag it in the second pass. :)

RubyNova: Ok nvm GitHub mobile won't let me - can you remove the `#include utility` at the top? The very first thing should be the file comment and there should be no additional includes in here.

Basically just fixing something that should not have been a problem in the first place.

RubyNova (changes requested):
The only change I ask for is the one I mentioned in a previous comment.

I'll approve once that's in. :)

bjauny:
And we're in!

RubyNova (approved):
Pizza party time! I'm happy with it, LGTM!

PR #525, opened 2022-10-28
Add fix for Wayland support in GLFW.

RubyNova:

This change enables support for the Wayland windowing system inside our build of GLFW.

No issue to track for this bug.

GLFW will compile for X11 only, causing wayland-only systems to fail.

Wayland-only systems should be able to use the GLFW windowing backend correctly.

No.

**Other information**:
Documentation that directed this change can be found here: https://www.glfw.org/docs/latest/compile.html#compile_options_wayland

### DIFF
--- a/internal/glfw/CMakeLists.txt
+++ b/internal/glfw/CMakeLists.txt
@@ -22,7 +22,7 @@ set(GLFW_BUILD_DOCS OFF CACHE BOOL "" FORCE)
 
 if (UNIX AND NOT APPLE)
     message(STATUS "Linux detected, enabling Wayland support for GLFW.")
-    set(GLFW_BUILD_WAYLAND ON CACHE BOOL "" FORCE)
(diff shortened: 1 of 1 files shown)


RubyNova:
Nevermind. Ignore. Just read that you need the X11 interface for better NVIDIA support anyway.

FiniteReality:
We should probably keep this option on anyway, as most applications will gracefully fall back to X11 if Wayland is not present. It's unlikely that a NVIDIA user would be using Wayland at the current time anyway.

capnkenny:
That would leave this blocked, as GLFW has a bug (fixed in main but not released yet) that actually prevents X11 & Wayland support from compiling together (as we accidentally discovered today 😅)

FiniteReality:
Oh, that's unfortunate. In that case, yeah, we should open a blocked issue for Wayland support.

Issue #500, opened 2022-09-10
Prepare NovelRT for C++ 20


Pheubel:
Although NovelRT compiles for C++ 17 on both Clang and MSVC, there are some obstacles that prevent the engine from moving forwards to C++ 20.

## Fabulist 
Fabulist currently fails to compile in C++ 20 due to a change in how `std:accumulate()` works. Relevant issue: https://github.com/novelrt/Fabulist/issues/14

## NovelRT/LoggingService
Due to how C++ 20 behaves, the interaction with spdlog and fmt becomes a bit more rough. The following lines are affected:
https://github.com/novelrt/NovelRT/blob/2559c00a896c77759f062d01143c3fe8805c2db8/include/NovelRT/LoggingService.h#L61-L76

One way to get around this issue is by wrapping the argument we want to pass in `fmt::runtime()`. Here is an example with `logInfo()`:
```cpp
template<typename I, typename... IRest> void logInfo(I current, IRest... next) const
{
    _logger->info(fmt::runtime(current), std::forward<IRest>(next)...);
}
```
Another option that can be explored is to have spdlog use C++'s [`std::format()`](https://en.cppreference.com/w/cpp/utility/format/format) by setting `SPDLOG_USE_STD_FORMAT`, it is to be noted that in the current release of spdlog, it is currently marked as experimental. 

## NovelRT/Graphics/IGraphicsMemoryRegionCollection
Right now something strange is going on within `IGraphicsMemoryRegionCollection::DefaultMetadata::Clear()`. with both clang and MSVC it currently compiles just fine on C++ 17, however when switching to C++ 20 MSVC will have issues with `GetDevice()`, it is no longer able to access it as expected. You will find that during building it will say `a call of a non-static member function requires an object`.

https://github.com/novelrt/NovelRT/blob/2559c00a896c77759f062d01143c3fe8805c2db8/include/NovelRT/Graphics/IGraphicsMemoryRegionCollection.h#L221-L237

replacing it with `GetDeviceInternal()->shared_from_this()` seems to work around this issue.

It seems like a compiler bug and we will most likely have to wait until it is fixed unless we want to use the work around mentioned above.

## JsonCons
JsonCons seems to be causing trouble when compiling with C++ 20. After a bit of talk in Discord, it seems that a good option would be to replace it with nlohmann json, as it is already being used for Fabulist and seems to compile just fine for C++ 20.

___
[...]

PR #535, opened 2022-11-19
Feature/replace jsoncons

Pheubel:

Replaces jsoncons with nlohmann json as json provider for NovelRT

This PR partially resolves https://github.com/novelrt/NovelRT/issues/500, more specifically; jsoncons was causing issues when compiling for C++ 20. 

The `DesktopResourceLoader` uses jsoncons for bson deserialization/serialization

The `DesktopResourceLoader` uses nlohmann json for bson deserialization/serialization

If users relied on jsoncons as json provider, they will have to either set up the inclusion themself, or convert their json code to work with nlohmann json.
If users relied on nlohmann json as json provider (indirectly provided through Fabulist) then they should not have to change their code.

**Other information**:
So far I only touched the CMake and the C++ code. I do not really know what exact alterations I would have to make to the [LICENCE-DIST.md](https://github.com/novelrt/NovelRT/blob/main/LICENCE-DIST.md) and the [build-system.yml](https://github.com/novelrt/NovelRT/blob/main/.github/workflows/build-system.yml).


RubyNova:
The build system YML should not require changes but I'll leave that up to @capnkenny to confirm.

The licence distributable file requires you to add the licence for the new dependency, and remove the jsoncons licence assuming I added it originally, and that's it.

internal/CMakeLists.txt
```diff
@@ -1,9 +1,12 @@
+add_subdirectory(nlohmannJson)
+
+set(FABULIST_THIRDPARTY_JSON OFF)
```
capnkenny, line 3: This should ideally be set inside of Fabulist's dependency CMakeLists, before the `MakeAvailable` call

capnkenny (commented):
Overall seems good - just needs the licensing change, and I would move the one flag being set.

capnkenny:
Just confirming - build system does not need to change and passes as-is

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
```diff
@@ -294,15 +296,15 @@ namespace NovelRT::ResourceManagement::Desktop
     {
         filePath = _resourcesRootDirectory / filePath;
 
-        jsoncons::json j(jsoncons::json_object_arg);
-        std::vector<jsoncons::json> memberMetadataJson{};
+        nlohmann::json j{};
+        nlohmann::json memberMetadataJson{};
```
capnkenny, line 300: Is there a reason this isn't a std::vector of the json object anymore?

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
```diff
@@ -294,15 +296,15 @@ namespace NovelRT::ResourceManagement::Desktop
...
 
         for (auto&& member : package.memberMetadata)
         {
-            jsoncons::json newMemberJson(jsoncons::json_object_arg);
+            nlohmann::json newMemberJson{};
 
             newMemberJson["name"] = member.name;
-            newMemberJson["type"] = static_cast<uint32_t>(member.type);
+            newMemberJson["type"] = member.type;
```
capnkenny, line 307: Is `member.type` already a uint32_t for nlohmann::json?

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
```diff
@@ -269,20 +269,22 @@ namespace NovelRT::ResourceManagement::Desktop
...
 
-        for (auto&& obj : metadata.array_value())
+        for (auto&& obj : metadata.array())
         {
-            BinaryMemberMetadata newMemberMetadata{
-                obj["name"].as<std::string>(), static_cast<BinaryDataType>(obj["type"].as<uint32_t>()),
-                obj["location"].as<size_t>(),  obj["sizeOfTypeInBytes"].as<size_t>(),
-                obj["length"].as<size_t>(),    obj["sizeOfSerialisedDataInBytes"].as<size_t>()};
+            BinaryMemberMetadata newMemberMetadata{obj.value<std::string>("name", std::string()),
+                                                   obj.value<BinaryDataType>("type", BinaryDataType::NullOrUnknown),
```
capnkenny, line 278: What are the possibly `BinaryDataType`s here?

internal/nlohmannJson/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+target_include_directories(nlohmann_json::nlohmann_json
+    INTERFACE
+        $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:$<BUILD_INTERFACE:${json_SOURCE_DIR}/include>>
+)
+
+target_compile_options(nlohmann_json::nlohmann_json
+    INTERFACE
+        $<BUILD_INTERFACE:$<$<CXX_COMPILER_ID:MSVC>:/external:I${json_SOURCE_DIR}/include>>
+        $<$<CXX_COMPILER_ID:MSVC>:/external:W0>
```
capnkenny, line 34: Was this taken from Fabulist's version of the nlohmann::json dependency CMakeLists?

capnkenny (commented):
Just a few questions that need answering before approval or changes requested

internal/nlohmannJson/CMakeLists.txt
Pheubel: yeah it was, i did bump the version, as fabulist is a little behind.

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
Pheubel: If you mean how it's serialized: it's serialized as an integer by default,they can be serialized as strings as seen [here](https://json.nlohmann.me/features/enum_conversion/).

if you mean the values `BinaryDataType` can represent, then a full list can be seen [here](https://github.com/novelrt/NovelRT/blob/main/include/NovelRT/ResourceManagement/BinaryDataType.h). examples are:
* Boolean
* Int32
* Int64

internal/nlohmannJson/CMakeLists.txt
capnkenny: Not a problem - my only ask is that we double check how this aligns with the Engine's compiler flags? I believe there's a difference between the two and I'm not sure which we prefer...

Cc @RubyNova for compiler flag input

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
capnkenny: CC @RubyNova - should we be typing as UInt32 here? I see the static cast has been dropped, but I'm not sure if its necessary anymore (CI appears to pass but idk if we specifically test this)

internal/nlohmannJson/CMakeLists.txt
RubyNova: Since it is an external dependency I am unsure if it matters what flags this dependency requires.

You're best off asking @FiniteReality .

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
Pheubel: The json object behaves like std::vector, i feel like having it be a json object instead would make more sense in showing what it actually is.

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
Pheubel: peeking at the [json header file](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/conversions/to_json.hpp#L320), it does seem to already look at the underlying value in the enum, but i could be wrong. it is not the easiest library to look through

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
RubyNova: To confirm, do we have a test covering this code path in the case the vector would have multiple entries?

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
RubyNova: LGTM

internal/nlohmannJson/CMakeLists.txt
FiniteReality: Ideally we'd be using this format going forward, if I understand correctly. The `/external:W0` here is just used to turn off warnings for external code.

RubyNova (approved):
All questions have been resolved, LGTM

src/NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.cpp
FiniteReality: It's converted to an int by default, but it can be customised: https://json.nlohmann.me/features/enum_conversion/

I would suggest doing so, as this avoids potential reordering errors.

FiniteReality:
Merging on @RubyNova's instructions. Fingers crossed this doesn't break everything :stuck_out_tongue_winking_eye:

PR #536, opened 2022-11-20
Update README.md

YukiiVR:
This is an update to fix the dependency list since CMake handles most of them internally. This will hopefully clear some confusion for new contributors in the future. 

**Other information**: As of right now, the only things that must be installed is the Vulkan SDK.


RubyNova (approved):
LGTM

Issue #402, opened 2022-02-05
Add full GCC support.


RubyNova:
At present, NovelRT will fail to compile of GCC is used. Due to some concerns over platform support, it appears we need to provide support for GCC. It will also make our lives easier for Conan packaging.

capnkenny:
Just noting - Nice to have, but not required for our MVP

RubyNova:
Update: we no longer use or support Conan, but this is still something that'd be cool to fix regardless.

Pheubel:
In regards to the type punning issues presented by GCC [`std::bit_cast`](https://en.cppreference.com/w/cpp/numeric/bit_cast) seems like a possible solution to get the desired behavior. However since it is introduced in C++ 20, support for that must be figured out first.

Pheubel:
I feel like i can get the type punning issue out of the way with some black magic

Pheubel:
With the type punning out of the way, there are still some obstacles in the way from having GCC compile without having ti tinker with flags. A major one being templates, with the current structure it does not properly resolve the templates. However, a majority of these templates are in files that are currently being worked on in the [ImGui support PR](https://github.com/novelrt/NovelRT/pull/522), so doing that now is unwise.

PR #539, opened 2022-11-23
Added mechanism to type pun semi-safe

Pheubel:
- [x] Tests for the changes have been added (for bug fixes / features)

This PR aims to solve the type punning issues of https://github.com/novelrt/NovelRT/issues/402. I am not certain if that will be enough to get GCC support up and running.

This PR 

NovelRT fails to build with GCC.

NovelRT no longer fails to build with GCC due to type punning issues.

This should not introduce a breaking change to the API or behavior.

**Other information**:

this handles most cases, as long as C++ 17 is used and the macros don't cry.

in C++ 20 std::bit_set is used.
some compilers support it in earlier C++ versions as a sort of compiler intrinsic, if one is available it might use it. if the compiler has not implemented std::bit_set yet or a built in implementation cannot be found, it will fallback on memcpy for GNU based compilers except clang and icc, these use reinterpret_cast as fallback.

the list of gnu exceptions is a tricky one to pin down and this is my current best attempt.

### DIFF
--- a/include/NovelRT/Maths/GeoMatrix4x4F.h
+++ b/include/NovelRT/Maths/GeoMatrix4x4F.h
@@ -72,8 +72,8 @@ namespace NovelRT::Maths
          */
         inline void Translate(Maths::GeoVector3F vector)
         {
-            *reinterpret_cast<glm::mat4*>(this) =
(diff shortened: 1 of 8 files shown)


RubyNova (changes requested):
LGTM, just waiting on a unit test or two for the new BitCast utility method and then I'll merge.

RubyNova (approved):
LGTM

PR #540, opened 2022-11-23
Rewrite `GraphicsResourceManager` to remove arbitrary frame creation from the internals.

RubyNova:

This PR introduces a fix to resolve a possible bug with the `GraphicsResourceManager` arbitrarily ending Vulkan cmd lists and then starting new ones in the middle of a frame, by removing that behaviour.

N/A

The `GraphicsResourceManager` will end and start new cmd lists (the engine's concept of a "frame") at-will, which is probably causing unintentional side-effects during frames that are being prepped for actual rendering as opposed to pure resource loading.

The `GraphicsResourceManager` will now just append to the current cmd list, so frames may consist of both resource loading and actual drawing + presenting.

Any previous code relying on the old behaviour may no longer work - the `DefaultRenderingSystem` has already been checked and verified to still work as intended via the InputEcsSample and UIEventExample targets.

**Other information**:
This potential bug was discovered in the imgui branch, but I deemed it out of scope to be a part of that change.


include/NovelRT/Graphics/GraphicsResourceManager.h
```diff
@@ -75,11 +93,11 @@ namespace NovelRT::Graphics
...
-
+        void FreeConstantBufferData(GraphicsMemoryRegion<GraphicsResource>& regionToFree);
         void FreeVertexData(GraphicsMemoryRegion<GraphicsResource>& vertexResource);
+        void FreeIndexData(GraphicsMemoryRegion<GraphicsResource>& indexResource);
         void FreeTextureData(GraphicsMemoryRegion<GraphicsResource>& textureResource);
     };
 }
 
-#endif // !NOVELRT_GRAPHICS_GRAPHICSRESOURCEMANAGER_H
+#endif // !NOVELRT_GRAPHICS_GRAPHICSRESOURCEMANAGER_H
```
capnkenny, line 103: Nit: EOL

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -366,7 +454,7 @@ namespace NovelRT::Graphics
     GraphicsMemoryRegion<GraphicsResource> GraphicsResourceManager::AllocateConstantBufferRegion(size_t size,
                                                                                                  size_t alignment)
     {
-        auto bufferPtr = GetOrCreateConstantBufferForAllocationSize(size);
+        auto bufferPtr = GetOrCreateGraphicsBufferForAllocationSize(size, GraphicsBufferKind::Constant);
         return bufferPtr->Allocate(size, alignment);
     }
-}
+}
```
capnkenny, line 460: Nit: EOL

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +111,138 @@ namespace NovelRT::Graphics
         stagingBuffer->UnmapAndWrite(vertexRegion);
         currentContext->Copy(vertexBuffer, stagingBuffer);
 
-        currentContext->EndFrame();
-        _graphicsDevice->Signal(currentContext->GetFence());
-        _graphicsDevice->WaitForIdle();
-        currentContext->BeginFrame();
```
capnkenny, line 92: Are these just called somewhere else internally now? Or do we not need these calls anymore?

capnkenny (dismissed):
Overall LGTM, just needs two EOL corrections, and I have a question about some of the BeginFrame/EndFrame usage that previously existed. Also - will we need to make these changes in the Interop API eventually? (If so, I don't remember if we have a ticket open for a contributor to pick it up.)

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: > What is the new behavior (if this is a feature change)?
The GraphicsResourceManager will now just append to the current cmd list, so frames may consist of both resource loading and actual drawing + presenting.

It is assumed you are in a frame when you call these methods as per the behavioural change outlined in the ticket

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -48,22 +71,51 @@ namespace NovelRT::Graphics
...
         other._stagingBufferSize = 0;
+        _contextIndex = other._contextIndex;
+        other._contextIndex = 0;
 
         return *this;
     }
 
+    void GraphicsResourceManager::PrepForFrameWithContextIndex(size_t newContextIndex)
+    {
+        size_t finalContextCount = _contextIndex < newContextIndex ? newContextIndex : _contextIndex;
```
FiniteReality, line 90: What happens in the case of overflow here? I'm guessing overflow is an unlikely enough event that we have bigger issues in this case?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -48,22 +71,51 @@ namespace NovelRT::Graphics
...
+
+        size_t currentStagingBufferContextSize = _stagingBuffers.size();
+        while (currentStagingBufferContextSize <= finalContextCount)
+        {
+            _stagingBuffers.emplace_back(std::vector<std::shared_ptr<GraphicsBuffer>>{});
```
FiniteReality, line 95: Does this move or copy here?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -48,22 +71,51 @@ namespace NovelRT::Graphics
...
+            ++currentStagingBufferContextSize;
+        }
+
+        if (_stagingBuffers[_contextIndex].size() == 0)
```
FiniteReality, line 99: Is this supposed to be `finalContextIndex`?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -48,22 +71,51 @@ namespace NovelRT::Graphics
...
+            for (auto&& buffer : _stagingBuffers[_contextIndex])
+            {
+                totalSize += buffer->GetSize();
+            }
+
+            totalSize /= _stagingBuffers.size();
+            _stagingBufferSize = totalSize;
+        }
+
+        _stagingBuffers[newContextIndex].clear();
```
FiniteReality, line 116: Is this supposed to be `finalContextIndex`?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -48,22 +71,51 @@ namespace NovelRT::Graphics
...
+
+        if (_stagingBuffers[_contextIndex].size() == 0)
+        {
+            _stagingBufferSize = _graphicsDevice->GetMemoryAllocator()->GetSettings().MinimumBlockSize;
+        }
+        else
+        {
+            size_t totalSize = 0;
+
+            for (auto&& buffer : _stagingBuffers[_contextIndex])
```
FiniteReality, line 107: Is this supposed to be `finalContextIndex`?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
-        std::shared_ptr<GraphicsBuffer> stagingBuffer = _stagingBuffer.getActual();
+        _stagingBufferSize = _stagingBufferSize < sizeToStage ? sizeToStage : _stagingBufferSize;
+
+        std::shared_ptr<GraphicsBuffer> chosenBuffer = nullptr;
+
+        for (auto&& buffer : _stagingBuffers[_contextIndex])
+        {
+            if (buffer->GetLargestFreeRegionSize() >= sizeToStage)
+            {
+                chosenBuffer = buffer;
```
FiniteReality, line 183: Can't we just return the buffer directly here, avoiding the null check later on?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
         }
-        return stagingBuffer;
+
+        return chosenBuffer;
     }
 
     std::shared_ptr<GraphicsBuffer> GraphicsResourceManager::CreateStagingBuffer()
     {
-        return _graphicsDevice->GetMemoryAllocator()->CreateBufferWithDefaultArguments(
+        auto bufferToReturn = _graphicsDevice->GetMemoryAllocator()->CreateBufferWithDefaultArguments(
```
FiniteReality, line 198: Might be better to just place this entire thing in the `emplace_back` so that it can move/whatever as appropriate.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
             Graphics::GraphicsBufferKind::Default, Graphics::GraphicsResourceAccess::Write,
             Graphics::GraphicsResourceAccess::Read, _stagingBufferSize);
+
+        _stagingBuffers[_contextIndex].emplace_back(bufferToReturn);
```
FiniteReality, line 202: You can likely return emplace_back here too.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
+        return bufferToReturn;
     }
 
     std::shared_ptr<GraphicsBuffer> GraphicsResourceManager::GetOrCreateGraphicsBufferForAllocationSize(
-        size_t allocationSize)
+        size_t allocationSize,
+        GraphicsBufferKind bufferKind)
     {
-        for (auto&& buffer : _vertexBuffers)
+        std::vector<std::shared_ptr<GraphicsBuffer>>* buffers = nullptr;
```
FiniteReality, line 211: I'd prefer using a lambda here, something like `const auto buffers = [&]{ /* the switch statement, using return instead */ }();`.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
+            newBuffer = _graphicsDevice->GetMemoryAllocator()->CreateBufferWithDefaultArguments(
+                bufferKind, GraphicsResourceAccess::None, GraphicsResourceAccess::Write,
+                ((allocationSize > _tenMegabytesAsBytes) ? allocationSize : _tenMegabytesAsBytes));
+        }
 
-        _constantBuffers.emplace_back(newBuffer);
+        switch (bufferKind)
+        {
+            case GraphicsBufferKind::Default:
+                _defaultBuffers.emplace_back(newBuffer);
```
FiniteReality, line 262: I'd switch these to `return ...` instead of breaking.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
+                break;
+            case GraphicsBufferKind::Index:
+                _indexBuffers.emplace_back(newBuffer);
+                break;
+            case GraphicsBufferKind::Constant:
+                _constantBuffers.emplace_back(newBuffer);
+                break;
+        }
 
         return newBuffer;
```
FiniteReality, line 275: If the switch was changed to return the value of `emplace_back`, this should instead be changed to fail-fast/throw and return nullptr.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
```diff
@@ -86,86 +139,138 @@ namespace NovelRT::Graphics
...
                 return buffer;
             }
         }
 
-        size_t minimumBlockSize = _graphicsDevice->GetMemoryAllocator()->GetSettings().MinimumBlockSize;
-        size_t sizeToAllocate = _graphicsDevice->GetMemoryAllocator()->GetSettings().MinimumBlockSize;
+        std::shared_ptr<GraphicsBuffer> newBuffer = nullptr;
 
-        while (sizeToAllocate < allocationSize)
+        if (bufferKind == GraphicsBufferKind::Constant)
```
FiniteReality, line 239: I wonder if you could do something like `newBuffer = bufferKind == constant ? createConstant() : [&]{ /* create correct buffer and return it */ }();`.

FiniteReality (changes requested):

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: This is out of date.

RubyNova: Wait, actually, nevermind ignore previous comment - what overflow? It's always set by Vulkan.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: on MSVC internally this just calls `push_back` I believe, so whatever that's doing is what's happening here. I was mostly just trying to get it to compile and work as I want.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: No because we are checking if the previous frame did any uploading with this.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: No because we are accounting for the previous frame's upload sizes iwth this.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
FiniteReality: Mostly just curious on what happens if `newContextIndex` overflows from <some big value> to 0 - does Vulkan say that'll never happen?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: No because this is for the new context that is getting registered.

What is finalContextIndex anyway? I don't see it in the code.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: Vulkan will do whatever the GPU specifies in terms of minimum context count. Usually that's never more than 3 or 4.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
FiniteReality: I misread and got confused between finalContextCount and newContextIndex. My bad :sweat_smile:

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: No because an adequate buffer might not be found.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
FiniteReality: I see. I misread what the intent behind this variable was, and I was thinking it was an index.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
FiniteReality: I'm not sure what you mean; if this code is executing then the buffer is not null, and it's also adequate enough, right?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: It is an index - but it is an index semantically tied to a graphics context, which there is never more than 3 or 4 of.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: ah nevermind, I see what you mean - generally I don't like multiple return points but I can change it if you want, I just think it makes the code flow harder to understand.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: I don't know how this would look at all

FiniteReality: Something like this:
```cpp
const auto buffers = [&]{
    switch (bufferKind) {
        case GraphicsBufferKind::Default:
            return _defaultBuffers;
        // etc.
    }
    // fail-fast for invalid bufferKind
}();
````

FiniteReality: If it helps, the C++ core guidelines has a rule for this to help explain it: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-lambda-init

RubyNova: Ah, that's interesting, so its a const correctness thing then?

FiniteReality: Yeah, a little bit. It also just makes it clearer what values are allowed, and keeps the selection colocated with the variable so that it can't get easily shuffled around in a refactor by accident.

src/NovelRT/Graphics/GraphicsResourceManager.cpp
RubyNova: You want a whole new private method for creating a constant buffer? Or?

FiniteReality: No, I simplified the code for creating a constant buffer to `createConstant()` since it's a single method call. My bad :sweat_smile:

FiniteReality: To clarify, `createConstant` was a placeholder for lines 241-243.

RubyNova: OK, so I've got it as a ternary, but its now an unreadable mess. You sure you want this?

src/NovelRT/Graphics/GraphicsResourceManager.cpp
FiniteReality: It might be worth looking into this later, then, as a potential optimisation that could be made. Not super important for now though.

FiniteReality (approved):

PR #541, opened 2022-11-25
Fix C++20 template resolution for MSVC in C++20 mode.

RubyNova:

This PR introduces an MSVC-specific C++20 fix - it appears MSVC has a bug in template resolution when in C++20 mode, this seems to resolve it.

N/A

C++20 fails to build becuase it thinks we are calling instance methods in static contexts when we aren't.

C++20 resolves the template correctly and so can build on MSVC.

C++17 should be fine, so no.

**Other information**:
N/A

### DIFF
--- a/include/NovelRT/Graphics/IGraphicsMemoryRegionCollection.h
+++ b/include/NovelRT/Graphics/IGraphicsMemoryRegionCollection.h
@@ -227,7 +227,7 @@ namespace NovelRT::Graphics
 
                 _regions->clear();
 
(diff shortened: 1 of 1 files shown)


Pheubel:
I think this would close https://github.com/novelrt/NovelRT/issues/500

FiniteReality (approved):
It's a bit weird that you've put the parameters on the same line on line 606 and not on 614, but if that's what the clang-format says then who am I to argue?

Issue #391, opened 2022-01-24
GraphicsPipelineBlendFactor needs reworking.


RubyNova:
This issue is an offshoot of imperfections from #376 .

I have introduced a new enum for handling blend modes, however it is just a direct copy of Vulkan currently. We need to make a superset of the different blend modes between D3D12 and Vulkan, and probably metal as well.

RyadaProductions:
I have created the PR for this, there was no need to create a superset as both D3D12 and Vulkan support the same blend operations according to their docs. 
[D3D12_BLEND docs](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_blend)
[VkBlendFactor docs](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkBlendFactor.html)

PR #571, opened 2023-03-19
#391 - Add preliminary support for D3D12_BLEND and MTLBlendFactor enums

RyadaProductions:

Introduce preliminary support for the D3D12_BLEND and MTLBlendFactor enums.
The changes add a translation layer between our render agnostic enum and the actual render enums.
It will not impact any working system as of yet as this only provides the utility mapping and not an actual wiring of the utility.

**Is there an open issue that this resolves? If so, please [link it here](https://github.com/novelrt/NovelRT/issues/391).**

N/A

N/A

No current systems are being impacted by this PR so it should not cause any issues.

**Other information**:
Currently the utility returns magic numbers, but this is due to the fact that we don't have the D3D12 and Metal dependency included yet.
Once that dependency has been added we can remove the magic number returns and uncomment the commented returns.
Those are already set to the actual values of the D3D12 and Metal enum names so it will be an easy edit once it's in place.


include/NovelRT/Graphics/D3D12/Utilities/PipelineBlendFactor.h
```diff
@@ -0,0 +1,102 @@
...
+#error NovelRT does not support including types explicitly by default. Please include Graphics.D3D12.Utilities.h instead for the Graphics::D3D12::Utilities namespace subset.
+#endif
+
+// TODO: Once D3D12 has been added as a dependency and we have access to the blend enum D3D12_BLEND remove magic numbers
+// and uncomment the returns with the enums.
+namespace NovelRT::Graphics::D3D12::Utilities
+{
+    [[nodiscard]] inline VkBlendFactor GetD3D12BlendFactor(GraphicsPipelineBlendFactor blendFactor)
+    {
+        switch (blendFactor)
```
tannergooding, line 17: A lookup table is likely going to be better/more efficient

include/NovelRT/Graphics/D3D12/Utilities/PipelineBlendFactor.h
```diff
@@ -0,0 +1,102 @@
...
+        {
+            case GraphicsPipelineBlendFactor::Zero:
+                // return D3D12_BLEND_ZERO;
+                return 1;
+                break;
```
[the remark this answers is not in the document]
tannergooding: It'd be better to not introduce "dead code" in more places. We should instead remove it here and log an issue to remove it in the other place as well.

include/NovelRT/Graphics/D3D12/Utilities/PipelineBlendFactor.h
RyadaProductions: It depends on how you look at it, the switch will as far as I know not allocate anything and it's performance issue comes from the fact that branch prediction can get it completely wrong. 
So it's a question of do we want to possibly give up a tiny bit of cpu cycles for memory access. 
And lookup tables for integer based lookups (which is essentially what we would create as the enum is just a fancy integer) is generally slower as far as I'm aware than a switch.
I could of course be wrong on this.

tannergooding: > the switch will as far as I know not allocate anything

neither will a lookup table, such tables can be embedded in the executable file as `rodata`

> We could potentially apply constexpr as well I think? Might be a good improvement for the vulkan version too

Not sure that will help. I expect this will often be flowing from code and not from a constant.

tannergooding: > With a lookup table I assume you just mean an std::array?

No, you do something like the following which is a fairly standard setup:
```cpp
#include <stdint.h>
#include <d3d12.h>

enum class GraphicsPipelineBlendFactor : uint32_t {
    Zero = 0,
    One = 1,
    SrcColour = 2,
    OneMinusSrcColour = 3,
    DstColour = 4,
    OneMinusDstColour = 5,
    SrcAlpha = 6,
    OneMinusSrcAlpha = 7,
    DstAlpha = 8,
    OneMinusDstAlpha = 9,
    ConstantColour = 10,
    OneMinusConstantColour = 11,
    ConstantAlpha = 12,
    OneMinusConstantAlpha = 13,
    SrcAlphaSaturate = 14,
    Src1Colour = 15,
    OneMinusSrc1Colour = 16,
    Src1Alpha = 17,
    OneMinusSrc1Alpha = 18,
    __COUNT,
};

static const D3D12_BLEND NovelRTToD3D12Blend[] = {
    D3D12_BLEND_ZERO,                   // Zero = 0,
    D3D12_BLEND_ONE,                    // One = 1,
    D3D12_BLEND_SRC_COLOR,              // SrcColour = 2,
    D3D12_BLEND_INV_SRC_COLOR,          // OneMinusSrcColour = 3,
    D3D12_BLEND_DEST_COLOR,             // DstColour = 4,
    D3D12_BLEND_INV_DEST_COLOR,         // OneMinusDstColour = 5,
    D3D12_BLEND_SRC_ALPHA,              // SrcAlpha = 6,
    D3D12_BLEND_INV_SRC_ALPHA,          // OneMinusSrcAlpha = 7,
    D3D12_BLEND_DEST_ALPHA,             // DstAlpha = 8,
    D3D12_BLEND_INV_DEST_ALPHA,         // OneMinusDstAlpha = 9,
    D3D12_BLEND_BLEND_FACTOR,           // ConstantColour = 10,
    D3D12_BLEND_INV_BLEND_FACTOR,       // OneMinusConstantColour = 11,
    D3D12_BLEND_ALPHA_FACTOR,           // ConstantAlpha = 12,
    D3D12_BLEND_INV_ALPHA_FACTOR,       // OneMinusConstantAlpha = 13,
    D3D12_BLEND_SRC_ALPHA_SAT,          // SrcAlphaSaturate = 14,
    D3D12_BLEND_SRC1_COLOR,             // Src1Colour = 15,
    D3D12_BLEND_INV_SRC1_COLOR,         // OneMinusSrc1Colour = 16,
    D3D12_BLEND_SRC1_ALPHA,             // Src1Alpha = 17,
    D3D12_BLEND_INV_SRC1_ALPHA,         // OneMinusSrc1Alpha = 18,
};

D3D12_BLEND GetD3D12BlendFactorLookup(GraphicsPipelineBlendFactor blendFactor)
{
    static_assert((sizeof(NovelRTToD3D12Blend) / sizeof(D3D12_BLEND)) == static_cast<uint32_t>(GraphicsPipelineBlendFactor::__COUNT));
    return NovelRTToD3D12Blend[static_cast<uint32_t>(blendFactor)];
}
```

> This sounds like a pain to maintain for an extremely small gain if we even get a gain from this.

It is a significant gain, the actual codegen is:
```asm
        mov     eax, ecx
        lea     rcx, OFFSET FLAT:D3D12_BLEND const * const NovelRTToD3D12Blend
        mov     eax, DWORD PTR [rcx+rax*4]
        ret     0
```

with an embedded data table:
```asm
D3D12_BLEND const * const NovelRTToD3D12Blend DD 01H        ; NovelRTToD3D12Blend
        DD      02H
        DD      03H
        DD      04H
        DD      09H
        DD      0aH
        DD      05H
        DD      06H
        DD      07H
        DD      08H
        DD      0eH
        DD      0fH
        DD      014H
        DD      015H
        DD      0bH
        DD      010H
        DD      011H
        DD      012H
        DD      013H
```

Compare that to the `switch` based approach:
```asm
        cmp     ecx, 18
        ja      $LN2@GetD3D12Bl
        mov     eax, ecx
        lea     rdx, OFFSET FLAT:__ImageBase
        mov     ecx, DWORD PTR $LN24@GetD3D12Bl[rdx+rax*4]
        add     rcx, rdx
        jmp     rcx
$LN4@GetD3D12Bl:
        mov     eax, 1
        ret     0
$LN5@GetD3D12Bl:
        mov     eax, 2
        ret     0
$LN6@GetD3D12Bl:
        mov     eax, 3
        ret     0
$LN7@GetD3D12Bl:
        mov     eax, 4
        ret     0
$LN8@GetD3D12Bl:
        mov     eax, 9
        ret     0
$LN9@GetD3D12Bl:
        mov     eax, 10
        ret     0
$LN10@GetD3D12Bl:
        mov     eax, 5
        ret     0
$LN11@GetD3D12Bl:
        mov     eax, 6
        ret     0
$LN12@GetD3D12Bl:
        mov     eax, 7
        ret     0
$LN13@GetD3D12Bl:
        mov     eax, 8
        ret     0
$LN14@GetD3D12Bl:
        mov     eax, 14
        ret     0
$LN15@GetD3D12Bl:
        mov     eax, 15
        ret     0
$LN16@GetD3D12Bl:
        mov     eax, 20
        ret     0
$LN17@GetD3D12Bl:
        mov     eax, 21
        ret     0
$LN18@GetD3D12Bl:
        mov     eax, 11
        ret     0
$LN19@GetD3D12Bl:
        mov     eax, 16
        ret     0
$LN20@GetD3D12Bl:
        mov     eax, 17
        ret     0
$LN21@GetD3D12Bl:
        mov     eax, 18
        ret     0
$LN22@GetD3D12Bl:
        mov     eax, 19
$LN2@GetD3D12Bl:
        ret     0
```

Which likewise needs its own embedded data table:
```asm
$LN24@GetD3D12Bl:
        DD      $LN4@GetD3D12Bl
        DD      $LN5@GetD3D12Bl
        DD      $LN6@GetD3D12Bl
        DD      $LN7@GetD3D12Bl
        DD      $LN8@GetD3D12Bl
        DD      $LN9@GetD3D12Bl
        DD      $LN10@GetD3D12Bl
        DD      $LN11@GetD3D12Bl
        DD      $LN12@GetD3D12Bl
        DD      $LN13@GetD3D12Bl
        DD      $LN14@GetD3D12Bl
        DD      $LN15@GetD3D12Bl
        DD      $LN16@GetD3D12Bl
        DD      $LN17@GetD3D12Bl
        DD      $LN18@GetD3D12Bl
        DD      $LN19@GetD3D12Bl
        DD      $LN20@GetD3D12Bl
        DD      $LN21@GetD3D12Bl
        DD      $LN22@GetD3D12Bl
```

So not only does the switch have to do the same general "data table lookup", it uses that to feed a dynamic jump meaning that it cannot be accurately predicted and it can pessimize perf quite a bit.

The lookup table approach also gets implicit constant folding done. There isn't a downside here ;)

> For example if we had to make changes or there was a bug with the translation layer we would have to open 2 files at the same time.
One with the lookup map and one with the NovelRT enum and you would need to cross reference constantly which one translates to which enum.
Granted we could technically comment the novelrt enum value on the indices inside of the lookup table as well as the actual enum value it translates to (due to not having Metal or D3D12 added) but this does not really sound maintainable. as if we change our enum people have to remember to update the translation layers to the correct indices.

Write tests and/or asserts to validate correctness. This is effectively "write once, never touch again".

Removing entries from our enum is both source and binary breaking. Changing existing entries is binary breaking. We shouldn't ever be doing it.

We should only be appending new options to the end of our enum; and that will ultimately only be when Vulkan, Metal, or D3D12 do it themselves.

RyadaProductions: I didn't know this version of lookup tables this makes a lot more sense compared to what I knew.

This is also way cleaner than how I imagined it looking so I am going to implement it this way for all 3 versions of the translation layers.

Thanks for the detailed response as well that makes it really clear as to the benefits and the (non existent) drawbacks of the lookup table.

RubyNova: While the asm in the pure runtime copy does show that it can still optimise to the same thing, `constexpr` could potentially cut quite a bit of chaff.

From a readability perspective I think its good to have as well.

RyadaProductions:
Build is failing due to the required windows SDK being larger than what is provided by default.
The SDK that the PR relies on is: v10.0.22621
This is due to the last 2 enum values in d3d12 only being provided in the latest windows sdk and not in the ones prior.

include/NovelRT/Graphics/D3D12/Graphics.D3D12.h
```diff
@@ -0,0 +1,33 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_GRAPHICS_D3D12_H
+#define NOVELRT_GRAPHICS_D3D12_H
+
+#ifdef WIN32
```
RubyNova, line 7: Can we move this to NovelRT.h instead?

include/NovelRT/Graphics/D3D12/Utilities/Graphics.D3D12.Utilities.h
```diff
@@ -0,0 +1,17 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_GRAPHICS_D3D12_UTILITIES_H
+#define NOVELRT_GRAPHICS_D3D12_UTILITIES_H
+
+#ifdef WIN32
```
RubyNova, line 7: We can move this to NovelRT.h to reduce duplication and to keep stuff easier to rationalise about

include/NovelRT/Graphics/D3D12/Utilities/PipelineBlendFactor.h
```diff
@@ -0,0 +1,52 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_GRAPHICS_D3D12_UTILITIES_PIPELINEBLENDFACTOR_H
+#define NOVELRT_GRAPHICS_D3D12_UTILITIES_PIPELINEBLENDFACTOR_H
+
+#ifdef WIN32
```
RubyNova, line 7: We can move this to NovelRT.h to reduce duplication and to keep stuff easier to rationalise about

include/NovelRT/Graphics/Metal/Graphics.Metal.h
```diff
@@ -0,0 +1,33 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_GRAPHICS_METAL_H
+#define NOVELRT_GRAPHICS_METAL_H
+
+#ifdef TARGET_OS_MAC
```
RubyNova, line 7: Let's move this to NoevlRT.h - same reasoning as the D3D12 preprocessor being moved

include/NovelRT/Graphics/Metal/Utilities/Graphics.Metal.Utilities.h
```diff
@@ -0,0 +1,17 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_GRAPHICS_METAL_UTILITIES_H
+#define NOVELRT_GRAPHICS_METAL_UTILITIES_H
+
+#ifdef TARGET_OS_MAC
```
RubyNova, line 7: Let's move this to NoevlRT.h - same reasoning as the D3D12 preprocessor being moved

include/NovelRT/Graphics/Metal/Utilities/PipelineBlendFactor.h
```diff
@@ -0,0 +1,52 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_GRAPHICS_METAL_UTILITIES_PIPELINEBLENDFACTOR_H
+#define NOVELRT_GRAPHICS_METAL_UTILITIES_PIPELINEBLENDFACTOR_H
+
+#ifdef TARGET_OS_MAC
```
RubyNova, line 7: Let's move this to NoevlRT.h - same reasoning as the D3D12 preprocessor being moved

include/NovelRT/Graphics/Vulkan/Utilities/PipelineBlendFactor.h
```diff
@@ -10,71 +10,38 @@
...
+        VK_BLEND_FACTOR_CONSTANT_COLOR,           // ConstantColour = 10,
+        VK_BLEND_FACTOR_ONE_MINUS_CONSTANT_COLOR, // OneMinusConstantColour = 11,
+        VK_BLEND_FACTOR_CONSTANT_ALPHA,           // ConstantAlpha = 12,
+        VK_BLEND_FACTOR_ONE_MINUS_CONSTANT_ALPHA, // OneMinusConstantAlpha = 13,
+        VK_BLEND_FACTOR_SRC_ALPHA_SATURATE,       // SrcAlphaSaturate = 14,
+        VK_BLEND_FACTOR_SRC1_COLOR,               // Src1Colour = 15,
+        VK_BLEND_FACTOR_ONE_MINUS_SRC1_COLOR,     // OneMinusSrc1Colour = 16,
+        VK_BLEND_FACTOR_SRC1_ALPHA,               // Src1Alpha = 17,
+        VK_BLEND_FACTOR_ONE_MINUS_SRC1_ALPHA,     // OneMinusSrc1Alpha = 18,
+    };
```
RubyNova, line 33: Is a C style array the best option for this?

I don't really mind too much it's just kind of odd

include/NovelRT/NovelRT.h
```diff
@@ -110,6 +110,8 @@
   // Experimental types
   #include <NovelRT/Graphics/Graphics.h>
   #include <NovelRT/Graphics/Vulkan/Graphics.Vulkan.h>
+  #include <NovelRT/Graphics/D3D12/Graphics.D3D12.h>
+  #include <NovelRT/Graphics/Metal/Graphics.Metal.h>
```
RubyNova, line 114: The preprocessor defines should move to here

RubyNova (changes requested):

include/NovelRT/Graphics/Vulkan/Utilities/PipelineBlendFactor.h
RyadaProductions: Seems that c style arrays are what people use for lookup tables. Haven't been able to find an example that uses something else.

RubyNova (approved):
LGTM

CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ if(POLICY CMP0135)
 endif()
 
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum macOS / OS X deployment version")
+set(CMAKE_SYSTEM_VERSION 10.0.22621.0)
```
RubyNova, line 9: Can I ask what this does?

RyadaProductions: It sets the minimal windows sdk to target for builds. The minimal version I added is the first one that includes all the Blend enum values.

RubyNova: Can we add a comment here? THhis will probably need to be moved when we plugin-ify things.

PR #572, opened 2023-03-28
#388 memcpy_s removal

RyadaProductions:

Changes the memcpy_s instances to use memcpy directly but adds asserts for debug time issues and if applicable nullptr checks on the destination/source

**Is there an open issue that this resolves? If so, please [link it here](https://github.com/novelrt/NovelRT/issues/388).**

N/A

N/A

No breaking change for the end user

**Other information**:
N/A

### DIFF
--- a/include/NovelRT/Utilities/Memory.h
+++ b/include/NovelRT/Utilities/Memory.h
@@ -1,0 +1,48 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#ifndef NOVELRT_UTILITIES_MEMORY_H
+#define NOVELRT_UTILITIES_MEMORY_H
+
+#include <string.h>
+#include <cassert>
+#include <cstring>
+#include <NovelRT/Exceptions/Exceptions.h>
+
+namespace NovelRT::Utilities
+{
+    class Memory
+    {
+    public:
+        static void Copy(void* destination, size_t destinationSize, const void* source, size_t count)
+        {
+            #ifdef WIN32
+            int result = memcpy_s(destination, destinationSize, source, count);
+
+            if (result != 0)
+            {
+                switch (result)
+                {
+                    case EINVAL:
+                        throw Exceptions::InvalidOperationException("Due to invalid values passed in.");
+                        break;
+                    case ERANGE:
+                        throw Exceptions::InvalidOperationException("Argument is out of the allowed range");
+                        break;
+                    default:
+                        throw Exceptions::InvalidOperationException("Unknown error");
+                }
+            }
+            #else
+            assert(destination != nullptr);
+            assert(source != nullptr);
+            assert(count != 0);
+            assert(destinationSize >= count);
+
+            memcpy(destination, source, count);
+            #endif
+        }
+    };
+}
+
+#endif //! NOVELRT_UTILITIES_MEMORY_H
--- a/include/NovelRT/Utilities/Memory.h
+++ b/include/NovelRT/Utilities/Memory.h
@@ -17,7 +17,7 @@ namespace NovelRT::Utilities
         static void Copy(void* destination, size_t destinationSize, const void* source, size_t count)
         {
             #ifdef WIN32
-            int result = memcpy_s(destination, destinationSize, source, count);
+            errno_t result = memcpy_s(destination, destinationSize, source, count);
 
             if (result != 0)
             {
@@ -34,10 +34,10 @@ namespace NovelRT::Utilities
                 }
             }
             #else
-            assert(destination != nullptr);
-            assert(source != nullptr);
-            assert(count != 0);
-            assert(destinationSize >= count);
+            if (destination == nullptr || source == nullptr || count <= 0 || destinationSize < count)
+            {
+                throw Exceptions::InvalidOperationException("Due to invalid values passed in.");
+            }
 
             memcpy(destination, source, count);
             #endif
--- a/include/NovelRT/Utilities/Memory.h
+++ b/include/NovelRT/Utilities/Memory.h
@@ -4,10 +4,10 @@
 #ifndef NOVELRT_UTILITIES_MEMORY_H
 #define NOVELRT_UTILITIES_MEMORY_H
 
-#include <string.h>
+#include <NovelRT/Exceptions/Exceptions.h>
 #include <cassert>
 #include <cstring>
-#include <NovelRT/Exceptions/Exceptions.h>
+#include <string.h>
 
 namespace NovelRT::Utilities
 {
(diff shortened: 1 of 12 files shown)


tannergooding:
> Changes the memcpy_s instances to use memcpy directly but adds asserts for debug time issues and if applicable nullptr checks on the destination/source

Why? What's the reasoning for not using an official C/C++ API in favor of rolling our own equivalent?

src/NovelRT/Ecs/Graphics/DefaultRenderingSystem.cpp
```diff
@@ -553,19 +553,22 @@ namespace NovelRT::Ecs::Graphics
...
+            throw NovelRT::Exceptions::OutOfMemoryException();
+        }
         ptr->sizeOfVert = dataTypeSize;
         ptr->stagingPtrLength = dataLength;
 
-#ifdef WIN32
-        memcpy_s(ptr->stagingPtr, size, data, size);
-#else
         memcpy(ptr->stagingPtr, data, size);
-#endif
```
tannergooding, line 568: Not sure why we have this `ifdef` in the first place. `memcpy_s` has been part of the C standard since C11.  It's defined in `string.h`

RubyNova:
> Why? What's the reasoning for not using an official C/C++ API in favor of rolling our own equivalent?

Because we can't actually use it in all locations and the errno_t stuff is additional faff I can't be arsed with, to put it flatly. It's easier for us to just validate the memcpy call is correct and be done. Only windows supports memcpy_s and I'd rather not maintain two branches of the code in this case.

RyadaProductions:
On the case of reusing a selfmade method for the assertions: Technically I could add a method that we can reuse like `memcpy_validation` but since we don't use memcpy that often I figured asserting it at the start of a method where we expect certain things to be passed in is cleaner. Ofc if we want a single method that does the assertion as well as the memcpy I'm not opposed to that but in this case I feel this is safer.

tannergooding:
> Ofc if we want a single method that does the assertion as well as the memcpy I'm not opposed to that but in this case I feel this is safer.

Rolling your own validation logic, especially logic that has to be duplicated at each callsite, is almost never safer ;)

Centralizing the checks to one place such that things can be correctly accounted for and will always be accounted for is the safest thing you can do.

In this case, if for some reason `memcpy_s` isn't available (such as GCC/Clang decided to not implement `Annex K`) then it'd be better to simply define a `memcpy_s` shim with the right surface area and to place the validation logic in that.

If for some reason you don't want to handle `errno_t` at each call site, then likewise define some C++ `MemoryCopy` utility that wraps `memcpy_s`, checks errno_t, and throws a proper exception on failure.

tannergooding:
So why not define our own `MemoryCopy` that wraps the right APIs where available and does the checks otherwise.

This centralizes everything, allows easy modification in the future if we need (say we want to disable the checks in release builds), allows ensuring consistent behavior (such as throwing C++ exceptions for failures or asserting correct results), etc.

RubyNova:
That is a possible solution. The original GH issue was just the minimally required work and I kind of wrote it out of frustration mid-Vulkan pipeline fixing ages ago. That's probably why. :D

tannergooding:
Ah, I missed that there was an issue at all.

Would be helpful to have it show as a regular issue https://github.com/novelrt/NovelRT/issues/388 and not as [link it here](https://github.com/novelrt/NovelRT/issues/388)

I think it would be much better/safer/extensible to make our own wrapper method in such cases. Having manually placed validation at each `memcpy` callsite is overall error prone and easy to forget to do. Wrapper ensures it always happens, can do the relevant kind of checks, expose the right kind of errors, and can easily change behavior for all callsites based on perf, security, or other needs.

FiniteReality:
Frankly, we should update `memcpy` calls to `std::copy` over a `std::span`/`gsl::span`. In addition, the APIs that return raw pointers should be updated to also use a span type - the C API would translate this to something along the lines of `void* fn(size_t* size)` and write the span's size to the `size` parameter if it's non-null, as is typical for C APIs.

RubyNova (approved):
LGTM

include/NovelRT/Utilities/Memory.h
```diff
@@ -0,0 +1,48 @@
...
+                        break;
+                    default:
+                        throw Exceptions::InvalidOperationException("Unknown error");
+                }
+            }
+            #else
+            assert(destination != nullptr);
+            assert(source != nullptr);
+            assert(count != 0);
+            assert(destinationSize >= count);
```
capnkenny, line 40: This appears to be ignored on macOS/Ubuntu in Release mode as the asserts don't come through. Can we make these asserts into logical checks that work on any configuration?

include/NovelRT/Utilities/Memory.h
```diff
@@ -0,0 +1,48 @@
...
+#include <cassert>
+#include <cstring>
+#include <NovelRT/Exceptions/Exceptions.h>
+
+namespace NovelRT::Utilities
+{
+    class Memory
+    {
+    public:
+        static void Copy(void* destination, size_t destinationSize, const void* source, size_t count)
```
capnkenny, line 17: Based on the usage I'm seeing, does it make sense to either remove the `destinationSize` parameter (and assume `count` == `size`? Or an overload that would do as such? (In a way that mimics `memcpy` over `memcpy_s`)

include/NovelRT/Utilities/Memory.h
capnkenny: To be a little more clear - I would expect the behaviour on the platforms to be synonymous, where as (although it seems expected due to the platform differences) one will throw regardless of the configuration whereas the other will just _go and yolo copy_ (lol) on Release configuration if I understand correctly

capnkenny (changes requested):
Just a general question (and concern about the build failure)

include/NovelRT/Utilities/Memory.h
RyadaProductions: Asserts are always compiled out due to the debug define not being there on release builds.
This was the approach at the time we decided on as this was just an initial implementation with more to be discussed about later on if I am remembering correctly. 
I have changed the asserts into a normal if check to it will throw if one of those conditions fails.

include/NovelRT/Utilities/Memory.h
RyadaProductions: The intent is to mimic memcpy_s on all platforms where it's not available. This is why we added the destinationSize even though due to asserts its usage was compiled out in non Windows builds.
Due to the change outlined above that should not be an issue anymore and we actually are checking against it. Keep in mind this is most likely just the initial implementation and as my knowledge grows I will extend the Memory utilities that we want to use in our own engine.

RubyNova (approved):
LGTM, there's some nice-to-haves we could maybe do with this but for an initial submission this does what we need.

PR #575, opened 2023-06-12
Improved flexiblity for ResourceManagement initialisation

jeuxjeux20:
Currently, if the default asset directory (using the executable file path) doesn't contain a valid asset database file, `GetResourceLoader` throws an exception. This makes it impossible to configure the asset directory (`ResourcesRootDirectory`), as this requires an instance of `ResourceLoader`.

This PR adds an `initAssets` parameter to `GetResourceLoader` (defaulted to true). If false, the `InitAssetDatabase` function will not be called. This lets the user do the following:
- Get the `ResourceLoader` without initialising assets (`provider.GetResourceLoader(false)`)
- Set the asset directory (`loader.ResourcesRootDirectory() = myPath`)
- Initialise the database (`loader.InitAssetDatabase()`)

The C API has been updated to reflect those changes, with added support for calling `InitAssetDatabase` and `GetIsAssetDBInitialised`.

Concerns:
- What should happen if the user forgets to initialise the database? Should we warn them using checks inside the ResourceLoader methods?


include/NovelRT/PluginManagement/IResourceManagementPluginProvider.h
```diff
@@ -13,12 +13,12 @@ namespace NovelRT::PluginManagement
...
     private:
-        [[nodiscard]] virtual ResourceManagement::ResourceLoader* GetResourceLoaderInternal() = 0;
+        [[nodiscard]] virtual ResourceManagement::ResourceLoader* GetResourceLoaderInternal(bool initAssets) = 0;
 
     public:
-        [[nodiscard]] inline std::shared_ptr<ResourceManagement::ResourceLoader> GetResourceLoader()
+        [[nodiscard]] inline std::shared_ptr<ResourceManagement::ResourceLoader> GetResourceLoader(bool initAssets = true)
         {
-            return GetResourceLoaderInternal()->shared_from_this();
+            return GetResourceLoaderInternal(initAssets)->shared_from_this();
```
RubyNova, line 21: redundant call to `shared_from_this()` might screw with copy ellipsis.

src/NovelRT.Interop/PluginManagement/NrtIResourceManagementPluginProvider.cpp
```diff
@@ -41,7 +42,8 @@ NrtResult Nrt_IResourceManagementPluginProvider_GetResourceLoader(NrtIResourceMa
         return NRT_FAILURE_NULL_ARGUMENT_PROVIDED;
     }
 
-    auto&& resourceLoader = reinterpret_cast<IResourceManagementPluginProvider*>(plugin)->GetResourceLoader();
+    const bool initBool = static_cast<bool>(initAssets);
```
RubyNova, line 45: is this how we handle `NrtBool` in other locations?

jeuxjeux20: Looks like there's no cast at all in other places in the code. Initially thought that a cast would be necessary to avoid some "narrowing cast is implementation-dependent" warnings but apparently that cast is defined in the standard. I'll remove that.

src/NovelRT/ResourceManagement/Desktop/DesktopResourceManagementPluginProvider.cpp
```diff
@@ -10,9 +10,9 @@ namespace NovelRT::ResourceManagement::Desktop
     {
     }
 
-    DesktopResourceLoader* DesktopResourceManagementPluginProvider::GetResourceLoaderInternal()
+    DesktopResourceLoader* DesktopResourceManagementPluginProvider::GetResourceLoaderInternal(bool initAssets)
     {
-        if (!_resourceLoader->GetIsAssetDBInitialised())
+        if (!_resourceLoader->GetIsAssetDBInitialised() && initAssets)
```
RubyNova, line 15:
```suggestion
        if (initAssets && !_resourceLoader->GetIsAssetDBInitialised())
```

RubyNova (changes requested):
Invert the check I have suggested please.

RubyNova (approved):
LGTM

PR #578, opened 2023-06-19
Isolate platform dependencies via CMake - Refactor 1

capnkenny:
- [X] Documentation has been added / updated (for bug fixes / features)

Begins to create dependency isolation based on the target platform. Also introduces target detection when none is provided.

This PR should not create _visibly new_ behaviour, however it will now specifically pipe in compile definitions, compile options, and libraries to be linked against when building NovelRT and co

Should not break existing setups at this time.

**Other information**:
Part of multiple changes to introduce platform dependency isolation - will begin to perform part 2 when we start to have platform-specific dependencies (maybe with audio ㄟ( ▔, ▔ )ㄏ)


CMakeLists.txt
```diff
@@ -63,123 +74,41 @@ set(NOVELRT_VORBIS_VERSION "1.3.7" CACHE STRING "Vorbis version")
...
 #
-# Alias "NovelRT::Utilities::Misc:Span" to "std::span" instead of "gsl::span".
-# If enabled, you need to make sure you current configuration supports "std::span".
+# Set default build type when not defined and not multi-config
 #
-if(NOVELRT_USE_STD_SPAN)
-  add_compile_definitions(NOVELRT_USE_STD_SPAN=true)
+if(NOT GENERATOR_IS_MULTI_CONFIG AND CMAKE_BUILD_TYPE STREQUAL "")
+  message(STATUS "Defaulting to Debug build as no build type was specified")
+  set(CMAKE_BUILD_TYPE "Debug")
```
RubyNova, line 86: Do we want the default to be debug?

cmake/NovelRTPlatformConfig.cmake
```diff
@@ -0,0 +1,16 @@
...
+if(NOT NOVELRT_TARGET IN_LIST NOVELRT_TARGET_OPTIONS)
+  message(FATAL_ERROR "Could not identify a proper target to build NovelRT for!")
+else()
+  list(FIND NOVELRT_TARGET_OPTIONS ${NOVELRT_TARGET} index)
+  if(index EQUAL 0) # "Win32"
+    include(${PLATFORMS_DIR}/windows.cmake)
+  elseif(index EQUAL 1) # "Linux"
+    include(${PLATFORMS_DIR}/linux.cmake)
+  elseif(index EQUAL 2) # "macOS"
+    include(${PLATFORMS_DIR}/macos.cmake)
```
RubyNova, line 12: Is there a reason we are depending on numbers here?

cmake/platforms/linux.cmake
```diff
@@ -0,0 +1,88 @@
...
+  Vorbis::vorbisenc
+  FLAC::FLAC
+  Opus::opus
+  Ogg::ogg
+)
+
+#
+# Compile Options
+#
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
RubyNova, line 57: we should potentially put a warning here that GCC isn't guaranteed to work at the moment maybe?

cmake/platforms/macos.cmake
```diff
@@ -0,0 +1,92 @@
...
+find_package(glfw3 ${NOVELRT_GLFW_VERSION} CONFIG REQUIRED)
+find_package(glm REQUIRED)
+find_package(Ogg ${NOVELRT_OGG_VERSION} REQUIRED)
+find_package(OpenAL CONFIG REQUIRED)
+find_package(Opus REQUIRED)
+find_package(Vorbis ${NOVELRT_VORBIS_VERSION} REQUIRED)
+find_package(flac ${NOVELRT_FLAC_VERSION} REQUIRED)
+find_package(SndFile REQUIRED)
+find_package(nlohmann_json REQUIRED)
+find_package(Vulkan ${NOVELRT_VULKAN_VERSION} REQUIRED COMPONENTS MoltenVK)
```
RubyNova, line 35: The duplication of this might be error-prone later down the line, though I can't really think of anything better myself right now, admittedly.

cmake/platforms/windows.cmake
```diff
@@ -0,0 +1,100 @@
...
+  /WX
+  /wd4611
+  /MP
+)
+
+#
+# Any verbose behaviour changes
+#
+if(NOVELRT_VERBOSE_BUILD)
+  if(CMAKE_GENERATOR STREQUAL "Visual Studio 2019 16" OR CMAKE_GENERATOR STREQUAL "Visual Studio 2022 17")
```
RubyNova, line 85: We don't work with VS2017 right? I'm not sure we have this check entirely given that.

internal/CMakeLists.txt
```diff
@@ -1,22 +1,17 @@
...
-cmake_policy(SET CMP0077 NEW)
-set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum macOS / OS X deployment version")
+#CMP0048 - a few packages don't set versions
+#CMP0077 - prevent clearing of certain variables
+#CMP0091 - Use CMAKE_MSVC_RUNTIME_LIBRARY Abstraction
+#CMP0135 - DOWNLOAD_TIMESTAMP follows new behaviour for CMake 3.24+
+foreach(policy CMP0048 CMP0077 CMP0091 CMP0135)
+  if(POLICY ${policy})
```
RubyNova, line 9: What's this actually checking?

cmake/platforms/linux.cmake
```diff
@@ -0,0 +1,88 @@
...
+    -Wall
+    -Wextra
+    -Werror
+    -Wno-c++98-compat
+    -Wno-c++98-compat-pedantic
+    -Wno-float-equal
+    -Wno-padded
+    -Wno-reserved-id-macro
+  )
+endif()
```
RubyNova, line 80: Can we not pattern match some of these to avoid the duplication with the if-else approach? similar to the original code

RubyNova (changes requested):
One change I definitely think we should be making, one change I am unsure of, and a couple questions.

cmake/platforms/macos.cmake
capnkenny: The duplication is purely because we dont have platform-specific dependencies - I'm expecting this to sort itself as we introduce them

cmake/platforms/windows.cmake
capnkenny: Unsure but probably not. This is a check that currently exists today, just relocated.

internal/CMakeLists.txt
capnkenny: This checks if the policy exists for the version of CMake that is configuring the project. If it does, it will set the policy's behaviour accordingly.

cmake/platforms/linux.cmake
capnkenny: I can give it a shot - unsure if the genex will pass through properly, that's the main concern. I'll follow up accordingly with a message or commit

cmake/NovelRTPlatformConfig.cmake
capnkenny: I guess STREQUAL would prevent magic numbers here for when we decide to add extra platforms. Ill change accordingly

CMakeLists.txt
capnkenny: Good question, do we?
It currently is today

RubyNova: Might be something for internal design discussion. :)

cmake/platforms/windows.cmake
capnkenny: Added an error message as we don't build currently with VS2017, nor is it supported by CI.

cmake/platforms/linux.cmake
capnkenny: Added a warning that GCC isn't fully supported, and YMMV if proceeding.

cmake/NovelRTPlatformConfig.cmake
capnkenny: Got rid of de magic numbers

cmake/platforms/linux.cmake
capnkenny: Looks like its working and not throwing fits - niceeee

RubyNova (approved):
LGTM

PR #609, opened 2025-05-18
Add support for push constants.

RubyNova:

This adds support for CmdPushConstants for @capnkenny since its needed by imgui.

No.

You can't use push constants. 😎 

you can use push constants. 😎 

No.

**Other information**:
I haven't written any unit tests for this since its Vulkan API. I've also  not actually tested this works as intended myself, because that would require me writing shaders I can't be bothered with to be honest, but the wiring is simple enough I don't think it matters.


Graphics/Core/include/NovelRT/Graphics/GraphicsPushConstantRange.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+namespace NovelRT::Graphics
+{
+    struct GraphicsPushConstantRange
+    {
+        ShaderProgramVisibility visibilityFlags = ShaderProgramVisibility::NoneOrUnknown;
+        size_t offset = 0;
+        size_t size = 0;
+    };
+}
```
capnkenny, line 16: Nit: EOL

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
```diff
@@ -241,4 +243,10 @@ namespace NovelRT::Graphics::Vulkan
     {
         vkCmdBindPipeline(_commandBuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline->GetVulkanPipeline());
     }
+
+
+    void VulkanGraphicsCmdList::CmdPushConstants(std::shared_ptr<VulkanGraphicsPipelineSignature> pipelineSignature, ShaderProgramVisibility visibility, size_t offset, NovelRT::Utilities::Misc::Span<uint8_t> values)
+    {
+        vkCmdPushConstants(_commandBuffer, pipelineSignature->GetVulkanPipelineLayout(), Utilities::GetVulkanShaderStageFlags(visibility), static_cast<uint32_t>(offset), static_cast<uint32_t>(values.size()), values.data());
```
capnkenny, line 250: does this mean that all values should be uint32_t?

Asking as ImGui currently uses floats in their example for scale and translation values

Graphics/Core/include/NovelRT/Graphics/GraphicsPushConstantRange.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <NovelRT/Graphics/ShaderProgramVisibility.hpp>
+
+namespace NovelRT::Graphics
+{
+    struct GraphicsPushConstantRange
+    {
+        ShaderProgramVisibility visibilityFlags = ShaderProgramVisibility::NoneOrUnknown;
+        size_t offset = 0;
+        size_t size = 0;
```
capnkenny, line 14: So we have to provide this data upfront for the pipeline, and then also provide it when calling the `PushConstants` command? More or less asking so I understand the reasoning behind these being stored here.

capnkenny (commented):
Overall LGTM, just questions and a nit

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
RubyNova: "all values"? elaborate?

Graphics/Core/include/NovelRT/Graphics/GraphicsPushConstantRange.hpp
RubyNova: Yes, its a metadata requirement when setting up the pipeline signature/layours. The CMD API I designed should handle the relevant data size assuming the buffer does not exceed the designated size that was initially specified here.

RubyNova: https://registry.khronos.org/vulkan/specs/latest/man/html/vkCmdPushConstants.html for more information

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
capnkenny: I think I understand more with the previous review comment - however, something of concern is that the `values` span would become a span of `uint8_t`s.
Specifically - if I have an array of float values to pass through the push constant, would we have a way of preserving those values if they get casted? Or is it purely just the array type, and the underlying values could literally be _whatever_?

RubyNova: its purely that we use `uint8_t` as the type to represent a byte, since a `uint8_t` is 8 bits. The data can be whatever. :)

capnkenny (approved):
LGTM - all questions answered :)

PR #611, opened 2025-05-19
Fix memory mapping assert.

RubyNova:

Bug fix

No.

You crash when you close the vulkan render.

You no longer crash when you close the vulkan render.

No.

**Other information**:


FiniteReality (approved):
LGTM. Though, would it be worth making the mapped memory regions variable atomic? I don't know if it's possible for multiple threads to want to map a GPU buffer at the same time, or if that even makes sense, but perhaps naively I can see a path where multiple CPU threads doing something like a mesh baking operation would want to map/write/unmap segments of a larger buffer.

PR #612, opened 2025-05-23
Placeholder merge for imgui support

capnkenny:
Doesn't fulfill guidelines, but we need these changes in so that @RubyNova knows what needs fixing/exposing


RubyNova (approved):
LGTM ship the C4 directly into the main branch wooooo

PR #616, opened 2025-05-28
Fix vkQueueSubmit duplicate calls.

RubyNova:

This fixes an issue where we are calling vkQueueSubmit multiple times for no real reason, which is making it harder for me to reason about fences as well as polluting RenderDoc framegrabs.

No.

We have pollution in framegrabs, and I think the multiple calls to vkQueueSubmit may have been breaking our fences.

We should be able to now both present and record commands for different context objects. One step closer to async rendering.

Yes. API for `GraphicsDevice<T>::Signal()` has been changed to accomodate this change.

**Other information**:


FiniteReality (approved):
Tested on linux, LGTM.

We're ignoring CI since this is a refactoring branch where we know things will fail.

PR #625, opened 2025-06-29
Use a build-time command to copy the resources instead

FiniteReality:

**What kind of change does this PR introduce?**
Instead of copying the files at configure time, requiring CMake to be re-ran if they change, we perform it as part of the build instead.

**Is there an open issue that this resolves?**
No. This was reported in Discord.

**What is the current behavior?**
CMake re-runs if the resources are modified.

**What is the new behavior?**
CMake does not re-run if the resources are modified.

**Does this PR introduce a breaking change?** 
It's CMake, who knows :upside_down_face: 

**Other information**:


capnkenny (approved):
LGTM

PR #629, opened 2026-05-05
Fabulist rewrite shenanigans

FiniteReality:
We're rewriting Fabulist because... Reasons. To make things a little simpler, we're just gonna do it in-tree for now and then extract it when we think it's stable enough.


Samples/Scripting/REPL/main.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#include <NovelRT/Exceptions/FileNotFoundException.hpp>
+#include <NovelRT/Exceptions/InvalidOperationException.hpp>
+
+// #include <NovelRT/ResourceManagement/Desktop/DesktopResourceLoader.hpp>
```
RubyNova: Dead comment. Can this be removed?

Samples/Scripting/REPL/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    if (!file.is_open())
+    {
+        throw NovelRT::Exceptions::FileNotFoundException(filePath.string());
+    }
+
+    auto filesize = file.tellg();
+    std::vector<uint8_t> data(filesize);
+    file.seekg(0);
+    file.read(reinterpret_cast<char*>(data.data()), filesize);
```
RubyNova, line 36: is char the right type here? I also wonder why are not using the resource loader now that we have confirmation it works ok.

Samples/Scripting/REPL/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+    file.close();
+
+    return data;
+}
+
+int main()
+{
+    // std::shared_ptr<DesktopResourceLoader> resourceLoader = std::make_shared<DesktopResourceLoader>();
```
RubyNova: Oh, okay, this explains a lot. Can you tell me what happened here?

Scripting/Core/Bindings/Fabulist/Choice.cpp
```diff
@@ -0,0 +1,87 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#include <NovelRT/Exceptions/InvalidOperationException.hpp>
+
+#include <NovelRT/Scripting/Statuses/Branch.hpp>
+
+#include "../Bindings.hpp"
```
RubyNova: Uhhh intended?

Scripting/Core/Bindings/Fabulist/Speaker.cpp
```diff
@@ -0,0 +1,69 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#include <NovelRT/Exceptions/InvalidOperationException.hpp>
+
+#include <NovelRT/Scripting/Statuses/SpokenLine.hpp>
+
+#include "../Bindings.hpp"
```
RubyNova: intended? We don't do header includes like this anywhere else.

Scripting/Core/Bindings/Fabulist/Speaker.cpp
```diff
@@ -0,0 +1,69 @@
...
+luaL_Reg SpeakerFuncs[]{
+    {"__call", ProduceSpeakerLine},
+    {"__gc", CleanupSpeaker},
+    {nullptr, nullptr},
+};
+
+void RegisterSpeaker(lua_State* L)
+{
+    if (luaL_newmetatable(L, "Speaker") == 0)
+        throw new NovelRT::Exceptions::InvalidOperationException("How did you get here?");
```
RubyNova: We should probably add braces around this. I am not a big fan of this style in C++ code, as much as this probably won't break anything.

Scripting/Core/Bindings/Fabulist.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#include <memory>
+
+#include "../Bindings.hpp"
```
RubyNova: Ah, this ol' chesnut again. Sorry to be annoying.

Scripting/Core/Bindings/Init.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#include <cstdlib>
+#include <exception>
+#include <random>
+
+#include "../Bindings.hpp"
```
RubyNova: This is the last time I will flag this include. I'll omit further uses of it.

Scripting/Core/Bindings/Init.cpp
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    return std::realloc(pointer, size);
+}
+
+int Panic(lua_State*)
+{
+    // TODO: the error object is on the top of the stack, we should log it.
+    std::terminate();
+}
```
RubyNova: What's the plan here, out of curiosity?

Scripting/Core/Bindings/Init.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+lua_State* NovelRT::Scripting::Bindings::CreateState()
+{
+    std::random_device r;
+    lua_State* L = lua_newstate(AllocMemory, nullptr, r());
+    // TODO: probably should handle this
+    if (L == nullptr)
+        return nullptr;
```
RubyNova: two things:

1 - the TODO. what is this exactly?
2 - can we get bracing on the if statement block please?

RubyNova (dismissed):
Mostly looks fine, these are mostly just questions I think.

Samples/Scripting/REPL/main.cpp
FiniteReality: It can, but ideally we'd use the resource loader for the Lua files too. There wasn't an API for loading plaintext as far as I could tell, so I left it commented out as a placeholder.

Samples/Scripting/REPL/main.cpp
FiniteReality: Yes. `std::ifstream` is a `std::basic_ifstream<char>` - meaning its `.read()` method takes a `char*`.

Scripting/Core/Bindings/Fabulist/Choice.cpp
FiniteReality: For now, yes. I need to tweak our CMake code to put this into a location we can read from but isn't included in the built outputs.

Scripting/Core/Bindings/Fabulist/Speaker.cpp
FiniteReality: For now, yes. I need to tweak our CMake code to put this into a location we can read from but isn't included in the built outputs.

Scripting/Core/Bindings/Fabulist/Speaker.cpp
FiniteReality: You're right. We shouldn't be using `new` here either; this is probably a C#/Javaism since I've been writing a bunch of that recently.

Scripting/Core/Bindings/Fabulist/Speaker.cpp
RubyNova: We have headers that are violating this in that case in the graphics namespace I think possibly.

Scripting/Core/Bindings/Init.cpp
FiniteReality: After calling the panic function, Lua calls abort() so there's not much we can do. The reason why I put a TODO for logging it is because we don't have a logger trivially available from this context, and I wanted to focus on getting the implementation ready before wiring up logging/etc.

Scripting/Core/Bindings/Init.cpp
FiniteReality: `lua_newstate` can return nullptr if it failed to create the state (e.g. because allocation failed) - we'd probably want to do something like throw bad_alloc or whatever, to be in-line with the rest of our C++ code.

capnkenny (approved):
Looks good to me! Interested to see how it will be integrated

PR #632, opened 2026-05-07
Add system dependency tree support to the ECS.

RubyNova:

This change introduces a new feature - a dependency graph/tree for ECS systems. This allows for systems to have dependencies without explicitly taking each other in the constructor, and makes the ECS simulation logic easier to reason about as a side effect.

No.

Vulkan takes out its metaphorical gun and gets mad at us because this fixes a parallel cmdlist problem we are having.

Vulkan no longer gets angry at us and we also get a really important feature. Hooray!

Yes. Please note the changes to the pre-existing SystemScheduler APIs - specifically dropping noexcept and the new exception handling. Any code that relied on this being noexcept is no longer good code.

**Other information**:
N/A

### DIFF
--- a/Ecs/Core/SystemScheduler.cpp
+++ b/Ecs/Core/SystemScheduler.cpp
@@ -7,12 +7,113 @@
 #include <NovelRT/Ecs/IEcsSystem.hpp>
 #include <NovelRT/Ecs/SystemScheduler.hpp>
 #include <NovelRT/Maths/Utilities.hpp>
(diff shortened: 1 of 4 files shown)


FiniteReality (approved):
This seems fine to me. I do however need to look into how the task arena/group/etc stuff from TBB works since I'm unfamiliar with it, as well as the MiscTemplateImpls. This is better suited to another PR though, since most of the code here doesn't actually touch that.

Looking at your toposort, it seems fine to me.

PR #634, opened 2026-05-17
Re-introduce sprite rendering to the ECS.

RubyNova:

This will bring us back up to feature parity with where we were before the graphics LLAPI refactor. Sprites should render and they will also respect the entity graph.

No.

the ECS cannot render anything at all. :)

The ECS can render things. :)

Not anything that is more broken than is currently in main :^)

**Other information**:
This is a lot of code. Sorry lol.


Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
```diff
@@ -25,13 +25,14 @@ namespace NovelRT::Ecs
             [&]()
             {
                 _asyncTasks->run(
-                    [work = std::forward<TWork>(work), completion = std::forward<TCompletion>(completion),
-                     this]() mutable
+                    [work = std::make_shared<std::decay_t<TWork>>(std::forward<TWork>(work)),
```
FiniteReality, line 28: Why are we using a shared ptr here?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Camera.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+namespace NovelRT::Ecs::Graphics::Components
+{
+    struct Camera
+    {
+        float left = -1.0f;
```
FiniteReality, line 10: Are these representing the camera frustum? If so, this should be named `Frustum` instead of `Camera`

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/TrackedSemaphore.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <memory>
+
+namespace NovelRT::Ecs::Graphics::Components
+{
+    template<typename TGraphicsBackend>
+    struct TrackedSemaphore
+    {
+        std::shared_ptr<NovelRT::Graphics::GraphicsSemaphore<TGraphicsBackend>>* semaphore = nullptr;
+        uint64_t signalValue = 0;
+        bool isWaitSemaphore = true;
```
FiniteReality, line 17: What does `isWaitSemaphore` mean?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Sprite.hpp
```diff
@@ -1,27 +1,31 @@
...
 
+#include <NovelRT/Graphics/RGBAColour.hpp>
+#include <uuid.h>
+
 namespace NovelRT::Ecs::Graphics::Components
 {
     struct Sprite
     {
+        uuids::uuid assetId{};
+        NovelRT::Graphics::RGBAColour tint = NovelRT::Graphics::RGBAColour(255, 255, 255, 255);
```
FiniteReality, line 13: Defaults like this should be part of the builder - we should make the default constructor preferably do as much zero-initialization as possible

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Viewport.hpp
```diff
@@ -0,0 +1,30 @@
...
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+namespace NovelRT::Ecs::Graphics::Components
+{
+    struct Viewport
+    {
+        float x = 0.0f;
+        float y = 0.0f;
+        float width = 1.0f;
```
FiniteReality, line 11: We should put these defaults into the builder, and have the default constructor zero-initialize as much as possible.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp
```diff
@@ -180,16 +181,23 @@ namespace NovelRT::Ecs::Graphics
...
-                    frameResources.descriptorSets.emplace_back(*descriptorSetPtr);
+                    auto renderPassComponent = renderPasses.GetComponent(entity);
+                    auto* descriptorSetsPtr = renderPassComponent.descriptorSets;
+
+                    for (size_t i = 0; i < renderPassComponent.descriptorSetCount; i++)
+                    {
+                        frameResources.descriptorSets.emplace_back(descriptorSetsPtr[i]);
+                    }
+
+                    delete subCmdListPtr;
```
FiniteReality, line 199: Why are we deleting them here? That will cause UB when we free them again when removing them, no?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
-            NovelRT::Graphics::GraphicsResourceMemoryRegion<NovelRT::Graphics::GraphicsBuffer, TGraphicsBackend>>
-            _indexRegion;
-
-        struct TexturedVertex
+        struct SpritePushConstant
         {
-            NovelRT::Maths::GeoVector3F Position;
-            NovelRT::Maths::GeoVector2F UV;
+            Maths::GeoMatrix4x4F model;
+            float tintR;
```
FiniteReality: Do we not have a Vector4 that would work for this instead?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                if (_textureCache.find(sprite.assetId) != _textureCache.end() ||
+                    _pendingLoads.find(sprite.assetId) != _pendingLoads.end())
+                {
+                    continue;
+                }
+
+                _pendingLoads.insert(sprite.assetId);
+
+                catalogue.ScheduleWithCompletion(
+                    [this, assetId = sprite.assetId]() { return _resourceLoader->LoadTexture(assetId); },
```
FiniteReality, line 166: We probably don't need to capture `this`, capturing `_resourceLoader` is safer.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                    [this, assetId = sprite.assetId](Timing::Timestamp /*delta*/, Catalogue completionCatalogue,
```
FiniteReality, line 167: Again, it's safer to capture the individual values in case the system gets moved or deallocated before this gets called (e.g. during shutdown)

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                                                     ResourceManagement::TextureMetadata textureMetadata)
+                    {
+                        NovelRT::Graphics::GraphicsBufferCreateInfo bufferCreateInfo{};
+                        bufferCreateInfo.cpuAccessKind = NovelRT::Graphics::GraphicsResourceAccess::Write;
+                        bufferCreateInfo.gpuAccessKind = NovelRT::Graphics::GraphicsResourceAccess::Read;
+                        bufferCreateInfo.size =
+                            (static_cast<size_t>(textureMetadata.width) * textureMetadata.height * 4) * 2;
```
FiniteReality: What are these constants referring to? If the 4 is a size of e.g. an int, it would be better to use `sizeof(int)`

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                }
+
+                _pendingLoads.insert(sprite.assetId);
+
+                catalogue.ScheduleWithCompletion(
+                    [this, assetId = sprite.assetId]() { return _resourceLoader->LoadTexture(assetId); },
+                    [this, assetId = sprite.assetId](Timing::Timestamp /*delta*/, Catalogue completionCatalogue,
+                                                     ResourceManagement::TextureMetadata textureMetadata)
+                    {
+                        NovelRT::Graphics::GraphicsBufferCreateInfo bufferCreateInfo{};
```
FiniteReality: Since we're now compiling as C++20, we can use [designated initializers](https://en.cppreference.com/cpp/language/aggregate_initialization#Designated_initializers) to assign these values in one block. There may be other spots I've missed this.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                        bufferCreateInfo.cpuAccessKind = NovelRT::Graphics::GraphicsResourceAccess::Write;
+                        bufferCreateInfo.gpuAccessKind = NovelRT::Graphics::GraphicsResourceAccess::Read;
+                        bufferCreateInfo.size =
+                            (static_cast<size_t>(textureMetadata.width) * textureMetadata.height * 4) * 2;
+
+                        auto textureStagingBuffer = _memoryAllocator->CreateBuffer(bufferCreateInfo);
```
FiniteReality, line 175: I'm assuming ideally we'd want to condense all of these uploads to a single command buffer and upload?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+
+                if (_cameraConstantBuffers.find(cameraEntityId) == _cameraConstantBuffers.end())
+                {
+                    NovelRT::Graphics::GraphicsBufferCreateInfo bufferCreateInfo{};
+                    bufferCreateInfo.bufferKind = NovelRT::Graphics::GraphicsBufferKind::Constant;
+                    bufferCreateInfo.cpuAccessKind = NovelRT::Graphics::GraphicsResourceAccess::Write;
+                    bufferCreateInfo.gpuAccessKind = NovelRT::Graphics::GraphicsResourceAccess::Read;
+                    bufferCreateInfo.size = sizeof(CameraConstantBuffer);
+
+                    auto constantBuffer = _memoryAllocator->CreateBuffer(bufferCreateInfo);
```
FiniteReality, line 270: I think ideally we'd want to create a single buffer which we allocate regions of, no?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+
+            for (auto [entityId, sprite] : spriteView)
             {
-                auto cmdList = context->CreateCmdList({});
+                if (_textureCache.find(sprite.assetId) == _textureCache.end())
+                {
+                    continue;
+                }
+
+                auto& texture = _textureCache.at(sprite.assetId);
```
FiniteReality: Instead of doing two lookups we can reuse the result of `.find()` - maybe a small try helper would be a good idea here:
```cpp
TextureCacheItem* texture;
if (!TryGetCachedTexture(sprite.assetId, &texture)) continue;

bool TryGetCachedTexture(uuids::uuid id, TextureCacheItem** result)
{
    auto& it = _textureCache.find(id);
    if ( == _textureCache.end())
    {
        *result = nullptr;
        return false;
    }

    *result = &it->second;
    return true;
}
```

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+
+                NovelRT::Ecs::Components::TransformComponent transform = transformView.GetComponent(entityId);
 
-                cmdList->Begin();
-                // rendering la la la
-                cmdList->CmdBindPipeline(nullptr); // TODO: sort this
+                auto currentCmdList =
+                    context->CreateCmdList(std::optional<NovelRT::Graphics::SecondaryCmdListInfo<TGraphicsBackend>>(
```
FiniteReality, line 312: We should change CreateCmdList to have two overloads instead of a std::optional - just noting this for later

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                    currentCmdList->CmdSetScissor(
+                        Maths::GeoVector2F(cameraData.viewport.x, cameraData.viewport.y),
+                        Maths::GeoVector2F(cameraData.viewport.width, cameraData.viewport.height));
+
+                    auto view = Maths::GeoMatrix4x4F::CreateFromLookAt(
+                        Maths::GeoVector3F(cameraData.transform.position.x, cameraData.transform.position.y, -1.0f),
+                        Maths::GeoVector3F(cameraData.transform.position.x, cameraData.transform.position.y, 0.0f),
+                        Maths::GeoVector3F(0.0f, -1.0f, 0.0f));
+
+                    auto projection = Maths::GeoMatrix4x4F::CreateOrthographic(
```
FiniteReality, line 371: We should likely add helpers to the camera component to create these matrices.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                        cameraData.camera.left, cameraData.camera.right, cameraData.camera.bottom,
+                        cameraData.camera.top, cameraData.camera.nearPlane, cameraData.camera.farPlane);
+
+                    auto& constantBufferRegion = _cameraConstantBuffers.at(cameraData.entityId);
+                    auto pCameraData = constantBufferRegion->GetOwningResource()->template Map<CameraConstantBuffer>(
```
FiniteReality, line 376: Future API consideration: being able to `region->Map(...)` directly

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                        constantBufferRegion);
+                    pCameraData[0].view = view;
+                    pCameraData[0].projection = projection;
+                    constantBufferRegion->GetOwningResource()->UnmapAndWrite(constantBufferRegion);
```
FiniteReality, line 380: Being able to delay the writing of these to be all-at-once (with a shared buffer, of course) would probably be ideal long-term.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+
+                    std::vector<std::shared_ptr<NovelRT::Graphics::GraphicsResourceMemoryRegion<
+                        NovelRT::Graphics::GraphicsResource, TGraphicsBackend>>>
+                        inputRegions{texture.textureRegion, constantBufferRegion};
+
+                    auto descriptorSet = _pipeline->CreateDescriptorSet();
+                    descriptorSet->AddMemoryRegionsToInputs(inputRegions);
+                    descriptorSet->UpdateDescriptorSetData();
```
FiniteReality, line 388: Future API consideration: a better way to encapsulate this create > add > update pattern which seems to be common with descriptor sets

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+
+                    descriptorSetArray[descriptorSetIndex] = descriptorSet;
+
+                    std::array<std::reference_wrapper<
+                                   const std::shared_ptr<NovelRT::Graphics::GraphicsDescriptorSet<TGraphicsBackend>>>,
+                               1>
+                        descriptorData{std::cref(descriptorSetArray[descriptorSetIndex])};
+
+                    currentCmdList->CmdBindDescriptorSets(descriptorData);
```
FiniteReality, line 397: Future API consideration: A single `BindDescriptorSet(...)` call.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+                    auto& constantBufferRegion = _cameraConstantBuffers.at(cameraData.entityId);
+                    auto pCameraData = constantBufferRegion->GetOwningResource()->template Map<CameraConstantBuffer>(
+                        constantBufferRegion);
+                    pCameraData[0].view = view;
+                    pCameraData[0].projection = projection;
+                    constantBufferRegion->GetOwningResource()->UnmapAndWrite(constantBufferRegion);
+
+                    std::vector<std::shared_ptr<NovelRT::Graphics::GraphicsResourceMemoryRegion<
+                        NovelRT::Graphics::GraphicsResource, TGraphicsBackend>>>
+                        inputRegions{texture.textureRegion, constantBufferRegion};
```
FiniteReality, line 384: Do we need to bind/push these for every sprite? It seems to me that most sprites will share the same values, so we can get away with a single one.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
```diff
@@ -19,146 +23,505 @@
...
+            std::shared_ptr<NovelRT::Graphics::GraphicsSurfaceContext<TGraphicsBackend>> surfaceContext)
+            : _resourceLoader(std::move(resourceLoader)),
+              _graphicsDevice(std::move(device)),
+              _memoryAllocator(std::move(memoryAllocator)),
+              _renderPass(std::move(renderPass)),
+              _surfaceContext(std::move(surfaceContext)),
+              _uploadSemaphore(_graphicsDevice->CreateSemaphore(0)),
+              _orchestratorSubmissionSemaphore(_graphicsDevice->CreateSemaphore(0))
+        {
+            NovelRT::Graphics::GraphicsBufferCreateInfo bufferCreateInfo{};
```
FiniteReality: Since we're now compiling as C++20, we can use [designated initializers](https://en.cppreference.com/cpp/language/aggregate_initialization#Designated_initializers) to assign these values in one block.

Graphics/Core/include/NovelRT/Graphics/GraphicsPipeline.hpp
```diff
@@ -22,9 +22,13 @@ namespace NovelRT::Graphics
...
 
         [[nodiscard]] std::shared_ptr<ShaderProgram<TBackend>> GetVertexShader() const noexcept;
+
         [[nodiscard]] std::shared_ptr<ShaderProgram<TBackend>> GetPixelShader() const noexcept;
+
         [[nodiscard]] std::shared_ptr<GraphicsPipelineSignature<TBackend>> GetSignature() const noexcept;
 
         [[nodiscard]] std::shared_ptr<GraphicsDescriptorSet<TBackend>> CreateDescriptorSet();
+
+        [[nodiscard]] std::shared_ptr<GraphicsRenderPass<TBackend>> GetRenderPass() const noexcept;
```
FiniteReality, line 32: Is this used anywhere? I'm not sure we should be exposing this here.

Graphics/Core/include/NovelRT/Graphics/RGBAColour.hpp
```diff
@@ -39,5 +39,15 @@ namespace NovelRT::Graphics
         {
             return a / 255.0f;
         }
+
+        [[nodiscard]] inline bool operator==(const RGBAColour& other) const noexcept
+        {
+            return r == other.r && g == other.g && b == other.b && a == other.a;
+        }
+
+        [[nodiscard]] inline bool operator!=(const RGBAColour& other) const noexcept
```
FiniteReality: Since we're C++20 we don't actually need this anymore - we can just declare `==` and it will synthesize a `!=`

Graphics/Vulkan/VulkanGraphicsSwapchainImage.cpp
```diff
@@ -198,6 +198,8 @@ namespace NovelRT::Graphics
         signalSemaphores.push_back(vulkanSemaphore);
         signalSemaphoreValues.push_back(0);
 
+            std::vector<VkPipelineStageFlags> waitDstStageMasks(waitSemaphores.size(), VK_PIPELINE_STAGE_ALL_COMMANDS_BIT);
```
FiniteReality: Is the indentation mismatching here?

Graphics/Vulkan/VulkanGraphicsSwapchainImage.cpp
```diff
@@ -175,7 +175,7 @@ namespace NovelRT::Graphics
             semaphoresToSignal)
     {
         VkSemaphore vulkanSemaphore = _swapchain->GetActiveSemaphore(shared_from_this())->GetVulkanSemaphore();
-        VkPipelineStageFlags allCommands = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
+        //VkPipelineStageFlags allCommands = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
```
FiniteReality: This is dead code

ResourceManagement/Desktop/DesktopResourceLoader.cpp
```diff
@@ -126,7 +126,7 @@ namespace NovelRT::ResourceManagement::Desktop
     {
         if (filePath.is_relative())
         {
-            filePath = _resourcesRootDirectory / "Images" / filePath;
+            filePath = _resourcesRootDirectory / "Images" / filePath.filename();
```
FiniteReality, line 129: This is mildly suspicious: if we have a filepath of `a/b.png`, won't this drop the `a/`?

Windowing/Glfw/Vulkan/GlfwWindowProviderVulkanGraphicsProvider.cpp
```diff
@@ -17,8 +17,7 @@ namespace NovelRT::Windowing
     {
         if (glfwVulkanSupported() == GLFW_FALSE)
         {
-            // throw Exceptions::InvalidOperationException("Vulkan is not supported on this platform.");
-            return nullptr;
+            throw Exceptions::InvalidOperationException("Vulkan is not supported on this platform.");
```
FiniteReality, line 20: Random question: why was this changed? Having the CreateGraphicsProvider return nullptr when a backend is unsupported means we can easily try another backend without having try/catch logic.

Windowing/Glfw/GlfwWindowProvider.cpp
```diff
@@ -14,6 +14,7 @@ namespace NovelRT::Windowing
 
     GlfwWindowProvider::WindowProvider(NovelRT::Windowing::WindowMode windowMode, Maths::GeoVector2F desiredWindowSize)
     {
+        glfwInitHint(GLFW_PLATFORM, GLFW_PLATFORM_X11);
```
FiniteReality: We should generally prefer Wayland if possible, so I'd like to see this removed if we can figure out the texture format issues.

Graphics/Vulkan/VulkanGraphicsSwapchain.cpp
FiniteReality: Thoughts for better logic here:
```cpp
auto& it = std::find_if(availableFormats.begin(), availableFormats.end(), [](auto& it){
    return it.colorSpace == VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
        && (it.format == VK_FORMAT_R8G8B8A8_UNORM || it.format == VK_FORMAT_B8G8R8A8_UNORM);
});

if (it != availableFormats.end())
    return *it;

_logger.logWarningLine("uh oh bad format :(");
return availableFormats[0];
```

Resources/Shaders/SpriteVertex.hlsl
```diff
@@ -0,0 +1,42 @@
...
+[[vk::push_constant]]
+SpritePushConstant pushConstant;
+
+[[vk::binding(1, 0)]]
+ConstantBuffer<CameraConstantBuffer> camera : register(b0);
+
+VSOutput main(VSInput input)
+{
+    VSOutput output;
+    float4 worldPos = mul(float4(input.Position, 1.0f), pushConstant.model);
```
FiniteReality, line 37: Is it not possible to do `model * float4(position, 1)`? GLSL lets you do this 😔

FiniteReality (changes requested):

Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
RubyNova: Workaround for an issue with OneTBB. It is internally copying the lambda which means the semantics of the captured objects end up getting a bit screwy just to ensure that we compile, basically. Shared pointer seemed like the safest bet so we didn't leak memory.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Camera.hpp
RubyNova: I don't know if I agree with this name change simply because while it might "functionally be a frustrum", I'm using it to drive the values of a camera, similar to a unity camera component.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/TrackedSemaphore.hpp
RubyNova: This is used to differentiate between semaphores to wait on, and semaphores to signal, when calling queue submit

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Sprite.hpp
RubyNova: Disagree in this case, delete instruction states differ from default settings in a component.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Viewport.hpp
RubyNova: Disagree in this case, default settings differ from delete instruction states.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: We do have GeoVector4F. I'll update things.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: This was mostly done because I was testing the waters with OneTBB. That said it is perfectly safe to do as a system getting deleted isn't a thing at runtime. I can update this either way though.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: The scheduler is designed to avoid these cases mainly because a system must always be behind a shared pointer and also the task arena won't let the destructor of the scheduler finish until task execution has fully completed.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: I just threw that in because it was complaining the buffer was too small for the texture, lol. I need to adjust this logic to fix this up.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Sure I'll try to figure this one out.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Long term yes, for now I really don't care since the buffers get disposed anyway once we're done with them and this works.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Not possible without complex resizing logic, the amount of cameras in a scene is n, which means we would have to effectively reimplement vector. Not really feeling like doing that right now.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: We should use the newer optional semantics if we're going to use a try method. I'll have a look into it.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: As long as it doesn't mess up the copyable nature of it then sure.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: We have an API header for this I believe it's just I was having linker errors with it and I was trying to ship.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Long term yes for now really don't care, see previous comment about reimplementing vector lol

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: The constant buffer region will change per camera and the texture region will also likely differ. We have to call bind each time even if the camera region changes currently.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Will give it a look.

Graphics/Core/include/NovelRT/Graphics/GraphicsPipeline.hpp
RubyNova: It seemed like a reasonable API inclusion at the time simply because a pipeline cannot exist without a render pass.

Graphics/Core/include/NovelRT/Graphics/RGBAColour.hpp
RubyNova: Oh really? That's blursed lol. I'll update.

Graphics/Vulkan/VulkanGraphicsSwapchainImage.cpp
RubyNova: No clue, ask clang format. I gave up wrestling with it last night.

Graphics/Vulkan/VulkanGraphicsSwapchainImage.cpp
RubyNova: Will remove.

ResourceManagement/Desktop/DesktopResourceLoader.cpp
RubyNova: Yes I put this in as a workaround for a separate bug I found involving the asset database. The issue is it would constantly try to load at images/images/a.png as opposed to images/a.png, this "fixes" it. I'm not happy with it for sure but I also didn't have the time to figure out a proper solution.

Windowing/Glfw/Vulkan/GlfwWindowProviderVulkanGraphicsProvider.cpp
RubyNova: Because I was struggling to debug what was going on at the time with Wayland so I just made it throw an exception again. If you recall that entire wrestling match I had with renderdoc on Linux this was a part of it. I can revert it if you want.

Windowing/Glfw/GlfwWindowProvider.cpp
RubyNova: This should be behind a renderdoc command line argument or something I kept adding it in when I was trying to frame grab and forgot to remove it this time. Woops. Will delete.

Graphics/Vulkan/VulkanGraphicsSwapchain.cpp
RubyNova: This logic will choose the Adobe RGB colour space again no?

Resources/Shaders/SpriteVertex.hlsl
RubyNova: It might be? I haven't tried it.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp
RubyNova: This is actually your code - we do this because it is a `shared_ptr*` not a `shared_ptr` AKA it is on the heap. We have to delete this to refcount it properly. it's already copied out on line 190.

Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
FiniteReality: Would a unique pointer not work too?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp
FiniteReality: Is it? https://github.com/novelrt/NovelRT/blob/main/Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp#L194 seems to suggest otherwise - the code I wrote to my knowledge relies on the update instruction doing the cleanup.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: Fair enough, I suppose. I figured we'd have the knowledge from the `cameraView` already since it pulls its data from the sparse set which *does* know how many items are allocated.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: Sure, that works too. We could probably even do something like what unordered_map::try_emplace does:
```
std::pair<_textureCache::iterator, bool> TryGetCachedValue(uuids::uuid id)
{
    auto& it = _textureCache.find(id);
    if (it == _textureCache.end())
    {
        return std::make_pair(it, false);
    }

    return std::make_pair(it, true);
}
```

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: The implementation might've been missed by accident.

Windowing/Glfw/Vulkan/GlfwWindowProviderVulkanGraphicsProvider.cpp
FiniteReality: In that case, we should likely come back to this when we implement more.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: It won't - they can even be static if you wanted to be sure 😛

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: I see. I think we'll want to revisit this eventually, but for now it's fine.

Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
RubyNova: That would make the lambda non-copyable. I don't know if OneTBB supports moving.

Graphics/Vulkan/VulkanGraphicsSwapchain.cpp
FiniteReality: No. It specifically checks for SRGB color space.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: This means we would need to have a constant buffer per frame instead then. Given our current use case I figured this would be good enough and I could come back to this without introducing scope creep in this PR.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Completely possible, we should revisit this later as a bug I think though otherwise this PR is gonna blow up further.

ResourceManagement/Desktop/DesktopResourceLoader.cpp
FiniteReality: Ah. Fair enough. We should keep track of this so we don't forget it.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp
RubyNova: Yeah my bad I goofed I'll look into this one more.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Camera.hpp
FiniteReality: The problem is, a camera stores a lot more than just the frustum - but a frustum may be used in multiple places for multiple reasons. Naming this a Frustum means we make it easier to grow what data we actually store as part of a "camera" (e.g. post processing settings), without actually needing a singular Camera struct.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/TrackedSemaphore.hpp
FiniteReality: I would separate these into two distinct components, then - it's possible an entity would produce both a "wait" semaphore *and* a "signal" semaphore. The name should also be changed to reflect that - something like `QueueSubmissionWaitSemaphore` and `QueueSubmissionSignalSemaphore`. This is because we may forget in the future and use a TrackedSemaphore by accident for something else :stuck_out_tongue:

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: Sure. The "constant buffer per frame" was the general idea, but this is valid enough for now.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Camera.hpp
RubyNova: How do we handle reference resolution then? That isn't really a frustrum thing.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/TrackedSemaphore.hpp
RubyNova: That would create a caching nightmare that I can't be bothered with right now unless you want to blow up the scope of this PR haha, the reason this was easy to do is because I already supported passing semaphores and the flag lets me know what type I passed

Graphics/Vulkan/VulkanGraphicsSwapchain.cpp
RubyNova: Oh I see. Okay sure I can simplify.

Graphics/Core/include/NovelRT/Graphics/GraphicsPipeline.hpp
FiniteReality: I checked the docs and you are correct, a pipeline can only exist for a given render pass (or one compatible) - I had misunderstood and assumed that it could be null to indicate "every render pass" :sweat_smile:

Resources/Shaders/SpriteVertex.hlsl
FiniteReality: I honestly don't know - most of my shader knowledge is in GLSL, where I do this regularly :sweat_smile:

RubyNova: I just find HLSL easier to read and understand honestly. Plus we can reuse these shaders with D3D12 later...that's the idea at least. Not sure how D3D12 handles push constants though lol. I'll figure that one out later.

Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
FiniteReality: It should do - if it doesn't that would be a bug in TBB as it explicitly states that it accepts *any* type fulfilling the function object requirements, which includes move-only types. In C++23 there's even a built-in wrapper for them, std::move_only_function

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Camera.hpp
FiniteReality: I missed those fields :sweat_smile: - they should be on a separate component. Some systems will want to scale their objects with regard to a reference resolution; others will want another factor, such as DPI for text; and some won't want to scale at all, like Terraria where it simply shows more of the world around you.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/TrackedSemaphore.hpp
FiniteReality: Fair enough. We should log this as something to do in the future.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/Components/Camera.hpp
RubyNova: Okay well now this just sounds like scope creep when you put it like that. This also sounds like it'll need to involve @capnkenny in the API design. Given this I don't know if we should really bother adjusting this right now.

FiniteReality: Sure. Just another one for the backlog :sweat_smile:

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
FiniteReality: Is the plan to adjust this in the future (i.e. outside of this PR)?

Resources/Shaders/SpriteVertex.hlsl
FiniteReality: I'm not sure that'll work, but fair enough on the readability argument.

Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
RubyNova: This no longer builds if I attempt to use a unique_ptr. I might need an example on how it should be done if you want me to do this.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Yeah reading the code I realised im stupid and could have just compared the result of find to end and then accessed it if its valid. I'm just gonna simplify to that instead since that'll be easier right now.

Ecs/Graphics/include/NovelRT/Ecs/Graphics/SpriteRendererSystem.hpp
RubyNova: Going to resolve this in a separate PR.

Ecs/Core/include/NovelRT/Ecs/MiscTemplateImpls.hpp
FiniteReality: Marking this resolved because I forgot about it and we can easily revisit this in the future.

FiniteReality (approved):

PR #649, opened 2026-05-29
Fix missing memory region APIs.

### LINKED ISSUE #641: Ensure GraphicsResourceMemoryRegion::Map is implemented correctly
Completely possible, we should revisit this later as a bug I think though otherwise this PR is gonna blow up further.

_Originally posted by @RubyNova in https://github.com/novelrt/NovelRT/pull/634#discussion_r3288378471_

RubyNova:

This fixes the missing memory region APIs discovered in PR #634 as well as an additional one I noticed during my fix for this. 

resolves #641 

Linker error. :)

No linker error. :)

Nope!

**Other information**:
N/A


Graphics/Core/GraphicsResourceMemoryRegion.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
+// for more information.
+
+#include <NovelRT/Graphics/GraphicsResource.hpp>
+#include <NovelRT/Graphics/GraphicsResourceMemoryRegion.hpp>
+
+namespace NovelRT::Graphics
+{
+
+    template<typename TBackend>
```
FiniteReality: These need to be in the .hpp as they're only partial specializations.

FiniteReality (changes requested):
We should probably also include a test/similar (e.g. by changing the sprite/graphics sample to use region->MapBytes) too.

Graphics/Core/GraphicsResourceMemoryRegion.cpp
RubyNova: We cannot do this due to circular includes I believe.

RubyNova: Ok this seems to build. Don't understand why but it does. Gonna do the other thing you asked me to do now.

Graphics/Vulkan/include/NovelRT/Graphics/Vulkan/VulkanGraphicsResourceMemoryRegion.hpp
```diff
@@ -109,5 +109,17 @@ namespace NovelRT::Graphics
...
+        template<typename T>
+        [[nodiscard]] Utilities::Span<T> Map()
+        {
+            return GetOwningResource()->template Map<T>(shared_from_this());
+        }
+
+        template<typename T>
+        [[nodiscard]] Utilities::Span<const T> MapForRead()
+        {
+            return GetOwningResource()->template MapForRead<T>(this);
```
FiniteReality: I think we want `shared_from_this()` here, this is still passing `this` as a pointer.

FiniteReality (changes requested):
One small typo that needs to be fixed, and I think this is good to merge.

FiniteReality (approved):

PR #654, opened 2026-06-07
Add posing/sprite support to speakers

FiniteReality:
Part three of #629.

This includes the character posing/sprite extensions necessary to decide how a character should appear on screen. Right now, the only field we expose is the sprite to use. This could easily be expanded to include more as necessary (e.g. position, scaling, tinting, etc.)

### DIFF
--- a/Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
+++ b/Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
@@ -27,7 +27,8 @@ NovelRT::Ecs::Scripting::SpokenLineStepSystem::SpokenLineStepSystem(DecisionTree
     stateManager.RegisterStateHandler(
         [](auto& status, auto& catalogue, auto entityId)
         {
-            auto [spokenLineComponents, poseComponents] = catalogue.template GetComponentViews<Components::SpokenLine, Components::Pose>();
+            auto [spokenLineComponents, poseComponents] =
+                catalogue.template GetComponentViews<Components::SpokenLine, Components::Pose>();
 
             if (auto* spokenLine = dynamic_cast<NovelRT::Scripting::Statuses::SpokenLine*>(status.get()))
             {
@@ -37,7 +38,8 @@ NovelRT::Ecs::Scripting::SpokenLineStepSystem::SpokenLineStepSystem(DecisionTree
 
                 poseComponents.PushComponentUpdateInstruction(
                     entityId, Components::Pose{new std::string(spokenLine->GetPose()->Name),
-                                               new std::string(spokenLine->GetPose()->Sprite)});
+                                               new std::string(spokenLine->GetPose()->Sprite),
+                                               spokenLine->GetPose()->Position, spokenLine->GetPose()->Scale});
 
                 return true;
             }
--- a/Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
+++ b/Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
@@ -8,6 +8,7 @@
 
 #include <NovelRT/Ecs/Scripting/Components/ActiveDecisionTree.hpp>
 #include <NovelRT/Ecs/Scripting/Components/ContinueDecisionTree.hpp>
+#include <NovelRT/Ecs/Scripting/Components/Pose.hpp>
 #include <NovelRT/Ecs/Scripting/Components/SpokenLine.hpp>
 #include <NovelRT/Ecs/Scripting/DecisionTreeStateManager.hpp>
 #include <NovelRT/Ecs/Scripting/StepSystems/SpokenLineStepSystem.hpp>
@@ -26,13 +27,17 @@ NovelRT::Ecs::Scripting::SpokenLineStepSystem::SpokenLineStepSystem(DecisionTree
     stateManager.RegisterStateHandler(
         [](auto& status, auto& catalogue, auto entityId)
         {
-            auto spokenLineComponents = catalogue.template GetComponentView<Components::SpokenLine>();
+            auto [spokenLineComponents, poseComponents] = catalogue.template GetComponentViews<Components::SpokenLine, Components::Pose>();
 
             if (auto* spokenLine = dynamic_cast<NovelRT::Scripting::Statuses::SpokenLine*>(status.get()))
             {
                 spokenLineComponents.PushComponentUpdateInstruction(
                     entityId, Components::SpokenLine{new std::string(spokenLine->GetSpeaker()),
                                                      new std::string(spokenLine->GetText())});
+
+                poseComponents.PushComponentUpdateInstruction(
+                    entityId, Components::Pose{new std::string(spokenLine->GetPose()->Name),
+                                               new std::string(spokenLine->GetPose()->Sprite)});
 
                 return true;
             }
@@ -62,6 +67,7 @@ void NovelRT::Ecs::Scripting::SpokenLineStepSystem::Update(Timing::Timestamp /* 
             Continue(catalogue, entity, spokenLine->Continue());
         }
 
+        // Note: we intentionally leave the Pose component here
         spokenLineComponents.RemoveComponent(entity);
         continueComponents.RemoveComponent(entity);
     }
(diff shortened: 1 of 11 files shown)


Ecs/Scripting/include/NovelRT/Ecs/Scripting/Components/Pose.hpp
```diff
@@ -0,0 +1,34 @@
...
+            delete name;
+            delete sprite;
+
+            *this = other;
+            return *this;
+        }
+
+        [[nodiscard]] inline bool operator==(const Pose& other) const noexcept
+        {
+            // TODO: is comparing the position/scale by value correct here?
```
RubyNova: Yes this is fine. NovelRT::Maths just uses GLM internally.

Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
```diff
@@ -26,14 +27,20 @@ NovelRT::Ecs::Scripting::SpokenLineStepSystem::SpokenLineStepSystem(DecisionTree
...
             if (auto* spokenLine = dynamic_cast<NovelRT::Scripting::Statuses::SpokenLine*>(status.get()))
             {
                 spokenLineComponents.PushComponentUpdateInstruction(
                     entityId, Components::SpokenLine{new std::string(spokenLine->GetSpeaker()),
                                                      new std::string(spokenLine->GetText())});
 
+                poseComponents.PushComponentUpdateInstruction(
+                    entityId, Components::Pose{new std::string(spokenLine->GetPose()->Name),
+                                               new std::string(spokenLine->GetPose()->Sprite),
+                                               spokenLine->GetPose()->Position, spokenLine->GetPose()->Scale});
```
RubyNova, line 42: Not something for now but I think we need to review how clang-format is deciding to format like this. Not sure if this is correct but functionally its fine and we turned this off in CI anyway so whatever.

Samples/Scripting/REPL/Resources/sample.lua
```diff
@@ -1,4 +1,14 @@
 speaker "narrator"
+speaker "character" {
+  pose = "happy",
+  poses = {
+    happy = { sprite = "happy.png" },
+    angry = { sprite = "angry.png" },
+    angry_left = { sprite = "angry.png", position = { -100, 0 } },
+    sad = { sprite = "sad.png", scale = { 1, 1 } },
+    sad_flip = { sprite = "sad.png", scale = { -1, 1 } }
```
RubyNova, line 9: I am guessing this isn't actually driving any real rendering just yet right?

Scripting/Core/Bindings/Fabulist/Speaker.cpp
```diff
@@ -7,22 +7,185 @@
...
 
+#include <optional>
+#include <string>
+#include <unordered_map>
+
+struct PoseInfo
+{
+    std::string sprite;
+    double xPos;
+    double yPos;
```
RubyNova: default should be 0, 0 for this.

Scripting/Core/include/NovelRT/Scripting/Statuses/SpokenLine.hpp
```diff
@@ -3,33 +3,48 @@
...
 {
     class SpokenLine final : public DecisionTreeStatus
     {
+    public:
+        struct Pose
+        {
+            std::string Name;
+            std::string Sprite;
+            NovelRT::Maths::GeoVector2F Position;
+            NovelRT::Maths::GeoVector2F Scale;
```
RubyNova, line 24: Do these not to be defaulted?

Scripting/Core/Statuses/SpokenLine.cpp
```diff
@@ -25,6 +29,11 @@ const std::string NovelRT::Scripting::Statuses::SpokenLine::GetSpeaker() const
     return _speaker;
 }
 
+auto NovelRT::Scripting::Statuses::SpokenLine::GetPose() const -> const std::optional<const Pose>
```
RubyNova, line 32: Can I ask why we're using a trailing return type here?

RubyNova (changes requested):

Samples/Scripting/REPL/Resources/sample.lua
FiniteReality: No, it's not. The data is exposed to the ECS however.

Scripting/Core/Bindings/Fabulist/Speaker.cpp
FiniteReality: That should be the default anyway, but I can make it explicit.

Scripting/Core/include/NovelRT/Scripting/Statuses/SpokenLine.hpp
FiniteReality: I don't think so, since we're always initializing them with a specific value.

Scripting/Core/Statuses/SpokenLine.cpp
FiniteReality: We're doing it because Pose is a member of SpokenLine, and if we didn't we'd need...
```cpp
const std::optional<const NovelRT::Scripting::Statuses::SpokenLine::Pose> NovelRT::Scripting::Statuses::SpokenLine::GetPose() const
```

Ecs/Scripting/include/NovelRT/Ecs/Scripting/Components/Pose.hpp
FiniteReality: I should have clarified - do we want to require them to have the same bit pattern (i.e. 1.010101010101 or whatever) or is "within a small epsilon because of floating point shenanigans" better?

Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
FiniteReality: On my end, clang-format preferred this format. I'm indifferent either way, but I already got frustrated with clang-format so while I'm running it locally I am checked-out of figuring out why it's doing things weirdly until we get something shipping.

Ecs/Scripting/include/NovelRT/Ecs/Scripting/Components/Pose.hpp
RubyNova: You can use epsilon equals if you want, this is a method that exists in the API. I believe we provide a default epsilon value as well? I could be wrong on that though been a minute.

Ecs/Scripting/StepSystems/SpokenLineStepSystem.cpp
RubyNova: Yep, agree. Mostly flagging it because clang format being clang format is really formatting my clang right now.

AKA its annoying me that it's doing this LMAO but like I said its fine. Not for now.

Ecs/Scripting/include/NovelRT/Ecs/Scripting/Components/Pose.hpp
FiniteReality: For an ECS component, which is better in your mind?

RubyNova: I think for a component this is probably good enough for now at least, I can't imagine this path is even going to get invoked often really anyway?

Scripting/Core/include/NovelRT/Scripting/Statuses/SpokenLine.hpp
FiniteReality: To clarify - I intend for a consumer specify all fields when they're initializing a struct of this type, so not specifying a default IMO sounds like a good idea?

RubyNova: I will warn you all of our other ECS components and such specify defaults for convenience because the defaults are usually the common values. Just something to consider.

FiniteReality: This isn't an ECS component however - this is in the Scripting namespace. The ECS namespace has its own component, based on the data from this struct. I've added the defaults to the ECS component, however.

RubyNova (approved):

PR #667, opened 2026-06-22
Implement DecisionTreeLoadingSystem

FiniteReality:


Ecs/Scripting/DecisionTreeLoadingSystem.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+struct NovelRT::Ecs::Scripting::DecisionTreeLoadingSystem::LoadResult
+{
+    NovelRT::Ecs::EntityId entity;
+    ResourceManagement::ScriptMetadata metadata;
+};
+
+auto NovelRT::Ecs::Scripting::DecisionTreeLoadingSystem::GetLoadRequests(ComponentView<Components::DecisionTreeLoadRequest>& requests) -> std::vector<LoadRequest>
+{
+    std::vector<LoadRequest> result;
```
RubyNova: nit - initialisation strategy.
```suggestion
    std::vector<LoadRequest> result{};
```

Ecs/Scripting/DecisionTreeLoadingSystem.cpp
```diff
@@ -0,0 +1,75 @@
...
+    const std::shared_ptr<NovelRT::ResourceManagement::ResourceLoader>& resourceLoader)
+    : _scriptManager{manager},
+      _resourceLoader{resourceLoader}
+{
+}
+
+void NovelRT::Ecs::Scripting::DecisionTreeLoadingSystem::Update(Timing::Timestamp /* delta */, Catalogue catalogue)
+{
+    auto loadRequests = catalogue.GetComponentView<Components::DecisionTreeLoadRequest>();
+    catalogue.ScheduleWithCompletion(
```
RubyNova, line 60: We are scheduling work in the case where there is no work to do, we should avoid this.

RubyNova (changes requested):

Ecs/Scripting/DecisionTreeLoadingSystem.cpp
```diff
@@ -0,0 +1,80 @@
...
+    const std::shared_ptr<NovelRT::ResourceManagement::ResourceLoader>& resourceLoader)
+    : _scriptManager{manager},
+      _resourceLoader{resourceLoader}
+{
+}
+
+void NovelRT::Ecs::Scripting::DecisionTreeLoadingSystem::Update(Timing::Timestamp /* delta */, Catalogue catalogue)
+{
+    auto loadRequests = catalogue.GetComponentView<Components::DecisionTreeLoadRequest>();
+    if (loadRequests.begin() == loadRequests.end()) return;
```
RubyNova: nit - formatting.
```suggestion
    if (loadRequests.begin() == loadRequests.end())
    {
        return;
    }
```

ResourceManagement/Desktop/DesktopResourceLoader.cpp
```diff
@@ -314,6 +316,45 @@ namespace NovelRT::ResourceManagement::Desktop
         return LoadShaderSource(GetGuidsToFilePathsMap().at(assetId));
     }
 
+    ScriptMetadata DesktopResourceLoader::LoadScript(std::filesystem::path filePath)
```
RubyNova, line 319: question on this one.

The resource loader already supports loading binary packages, this also appears to be reading it in as binary. Is there some reason the current binary package method doesn't work? It's fine if not I am just curious.

RubyNova (changes requested):

ResourceManagement/Desktop/DesktopResourceLoader.cpp
FiniteReality: The binary package seemed to be oriented around compressed archives, if you mean BinaryPackage/BinaryMemberMetadata. Considering that we have separate entry points for textures/shaders/etc, it seemed natural that we would have one for scripts too.

RubyNova: Makes sense.

RubyNova (approved):

PR #669, opened 2026-06-22
Translate Fabulist ECS components to Graphics ECS components

FiniteReality:
Because the Fabulist ECS state only pertains to the most recent status from the Lua code, it's not trivial to write a 1:1 mapping - as such we have to have some sort of a state tracking system which I've implemented here.

This is mostly untested since we don't have any code to do this yet. We'll write that later.


Ecs/Scripting/Graphics/include/NovelRT/Ecs/Scripting/Graphics/PoseToSpriteTranslationSystem.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <memory>
+#include <unordered_map>
+
+namespace NovelRT::Ecs::Scripting::Graphics
+{
+    class PoseToSpriteTranslationSystem : public NovelRT::Ecs::IEcsSystem
+    {
+    private:
+        struct TreeInfo;
```
RubyNova, line 19: What is this for?

Ecs/Scripting/Graphics/PoseToSpriteTranslationSystem.cpp
```diff
@@ -0,0 +1,148 @@
...
+            continue;
+        }
+
+        for (const auto& [_, character] : it->second.knownCharacters)
+        {
+            catalogue.DeleteEntity(character.sprite);
+        }
+
+        ++it;
+    }
```
RubyNova, line 51: Shouldn't this just be a range-for? What am I missing?

Also if you're wanting to delete all components of a given type we have a method for that too.

Ecs/Scripting/Graphics/PoseToSpriteTranslationSystem.cpp
```diff
@@ -0,0 +1,148 @@
...
+            {
+                continue;
+            }
+
+            poseIt = characterInfo.knownPoses.emplace_hint(poseIt, *pose.name, sprite.value());
+        }
+
+        characterInfo.activePose = poseIt->second;
+        characterInfo.position = pose.position;
+        characterInfo.scale = pose.scale;
```
RubyNova, line 112: Is this data applied to the relevant sprite's transform elsewhere?

RubyNova (changes requested):

Ecs/Scripting/Graphics/include/NovelRT/Ecs/Scripting/Graphics/PoseToSpriteTranslationSystem.hpp
FiniteReality: We use this type name in some of the private methods, but I didn't think it needed a full definition since it was only used in the private methods. That's why it's only a forward declaration.

Ecs/Scripting/Graphics/PoseToSpriteTranslationSystem.cpp
FiniteReality: It's supposed to be removing the values from the map, and I forgot to actually do that.

Ecs/Scripting/Graphics/PoseToSpriteTranslationSystem.cpp
FiniteReality: Yes, it's used in the ApplySpriteUpdates method.

RubyNova (approved):

Issue #664, opened 2026-06-21
Command list pipeline barrier commands


FiniteReality:
We don't have an issue for it. This change was mine I just kind of turbo'd out this duct tape solution while I work on a real one after we ship.

_Originally posted by @RubyNova in https://github.com/novelrt/NovelRT/pull/662#discussion_r3408400064_

Pipeline barriers are important, so we need to have some  way of inserting barriers on resources between pipeline stages as well as queues.

PR #671, opened 2026-06-23
Fix issue where a blank window would never appear when there is nothing to render.

RubyNova:
…ng rendered.

Fix that should resolve the original issue and some additional odd side effects we were seeing with imgui.

no.

The window won't open if there's nothing to render. It also will re-use the previous frame repeatedly if one exists after having nothing to render.

The window will now be empty based on the clear colour provided at configuration time, meaning you always get a new frame even if it's blank.

No, but it does introduce some new cmdlist APIs.

**Other information**:


Graphics/Vulkan/VulkanGraphicsCmdList.cpp
```diff
@@ -408,4 +448,70 @@ namespace NovelRT::Graphics
...
+        barrier.newLayout = vkNewLayout;
+        barrier.srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+        barrier.dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+        barrier.image = swapchainImage->GetVulkanImage();
+        barrier.subresourceRange = subresourceRange;
+
+        vkCmdPipelineBarrier(_commandBuffer, srcStageMask, dstStageMask, 0, 0, nullptr, 0, nullptr, 1, &barrier);
+    }
+
+    void VulkanGraphicsCmdList::CmdClearDepthAndColourSwapchainImage(
```
FiniteReality: These should be separate commands (that is, ClearColor and ClearDepthStencil)

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
```diff
@@ -408,4 +448,70 @@ namespace NovelRT::Graphics
...
+        const std::shared_ptr<GraphicsSwapchainImage<Vulkan::VulkanGraphicsBackend>>& swapchainImage,
+        ClearValue clearData)
+    {
+        // TODO: This assumes layout is in transfer optimal. Is this correct? - Matt J.
```
FiniteReality, line 485: We should probably use VK_IMAGE_LAYOUT_GENERAL until we figure out what layout to use.
> [VK_IMAGE_LAYOUT_GENERAL](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageLayout.html#) supports all types of device access, unless specified otherwise.

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
```diff
@@ -408,4 +448,70 @@ namespace NovelRT::Graphics
         auto vkCmdList = cmdList->GetVkCommandBuffer();
         vkCmdExecuteCommands(_commandBuffer, 1, &vkCmdList);
     }
+
+    void VulkanGraphicsCmdList::CmdTransitionSwapchainImageTo(
```
FiniteReality, line 451: When we transition the public API to using concepts we should be able to simplify this so we won't have to have multiple overloads for swapchain image vs regular image and so on.

I'm assuming that this is otherwise just a placeholder until we figure out a solution for #664?

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
```diff
@@ -20,9 +21,11 @@
 #include <NovelRT/Utilities/Span.hpp>
 
 #include <vulkan/vulkan.h>
+#include <vulkan/vulkan_core.h>
 
 namespace NovelRT::Graphics
 {
+
```
FiniteReality: Empty line?

Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp
```diff
@@ -139,11 +140,59 @@ namespace NovelRT::Ecs::Graphics
...
+
+            NovelRT::Graphics::ClearValue colourDataStruct{};
+            colourDataStruct.colour = _backgroundColour;
+            colourDataStruct.depth = 0;
+            colourDataStruct.stencil = 0;
+
             if (passes.empty())
             {
+                cmdList->CmdTransitionSwapchainImageTo(image, NovelRT::Graphics::ImageLayout::Undefined,
+                                                       NovelRT::Graphics::ImageLayout::DstTransfer);
```
FiniteReality, line 161: Since we're currently rendering directly to the swapchain we probably shouldn't use a transfer destination layout.

FiniteReality (changes requested):

Ecs/Graphics/include/NovelRT/Ecs/Graphics/RenderOrchestratorSystem.hpp
RubyNova: This is required unless we set it to general instead. I just chose the layout that made sense given what we're doing.

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
RubyNova: It's the beginnings of a simplified barrier API that I'm sure we'll improve over time, yeah. This is fine for now though.

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
RubyNova: I can change the name.

Graphics/Vulkan/VulkanGraphicsCmdList.cpp
RubyNova: We will have to change our transition calls if we change this here. This works and I think is what is likely expected given what we are doing (writing black everywhere)

FiniteReality (approved):

PR #704, opened 2026-07-01
Attempt to support macos app bundles

FiniteReality:
Some minor cleanup of the CMakeLists to support using the win32 subsystem and also building mac app bundles. These will get installed to a new `apps` folder in the installation directory.


cmake/NovelRTBuildSystem.cmake
```diff
@@ -180,24 +188,27 @@ function(NovelRTBuildSystem_DeclareModule moduleKind moduleName)
...
+      BUNDLE DESTINATION apps
+      FILE_SET interface_headers DESTINATION include
+      FILE_SET public_headers DESTINATION include
+
+      FILE_SET interface_resources DESTINATION bin/Resources
+      FILE_SET public_resources DESTINATION bin/Resources)
+
+    if(WIN32)
+      install(FILES $<TARGET_PDB_FILE:${cmakeSafeName}> DESTINATION lib OPTIONAL)
+    endif()
```
RubyNova, line 335: How does this handle the RPATHS out of curiosity?

FiniteReality: The call to `fixup_bundle` should handle it, if I'm reading it correctly. As for linux, my guess is that we'd package using flatpak/appimage/steamrt which should let us handle this.

.github/workflows/build-pr.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - uses: novelrt/setup-vulkan@v2
         with:
-          version: 1.4.313.0
+          version: 1.4.350.0
```
RubyNova, line 53: We need to make sure the readme is up to date.

RubyNova (changes requested):

.github/workflows/build-pr.yml
FiniteReality: This isn't necessary to build - any version of the Vulkan SDK should work. We're just using a newer version because of the Mac/MoltenVK shenanigans.

RubyNova: Right - which means macOS specifically will need this version, so we might as well just enforce it everywhere no?

RubyNova (approved):
LGTM!

PR #712, opened 2026-07-06
Add support for rudimentary system-based jobs.

RubyNova:

This PR introduces a new API for creating ECS based jobs, by allowing us to have temporary systems. These systems do not have dependencies, and run once per tick. They return false when not finished, and true when they are. I have also added an explicit API to cancel a job.

No.

This feature doesn't exist so things like animations etc. are very hard to do.

It is now possible to begin figuring out animation support.

Nope!

**Other information**:
This was only tested in a very barebones way, so we might run into problems as we use it.

### DIFF
--- a/Ecs/Core/SystemScheduler.cpp
+++ b/Ecs/Core/SystemScheduler.cpp
@@ -308,8 +308,7 @@ namespace NovelRT::Ecs
                             [&systemJob, this]()
                             {
                                 auto poolId = static_cast<size_t>(tbb::this_task_arena::current_thread_index());
-                                systemJob.isDone.store(systemJob.jobFnPtr(_currentDelta, Catalogue(poolId, *this)),
-                                                       std::memory_order_relaxed);
+                                systemJob.isDone = systemJob.jobFnPtr(_currentDelta, Catalogue(poolId, *this));
                             });
                     }
 
@@ -320,7 +319,7 @@ namespace NovelRT::Ecs
                     // a bit ugly, but couldn't think of a better way to do it right now - Matt J.
                     for (auto it = _systemJobs.begin(); it != _systemJobs.end();)
                     {
-                        if (it->second.isDone.load(std::memory_order_relaxed))
+                        if (it->second.isDone)
                         {
                             hadTombstones = true;
                             it = _systemJobs.unsafe_erase(it);
--- a/Ecs/Core/SystemScheduler.cpp
+++ b/Ecs/Core/SystemScheduler.cpp
@@ -1,6 +1,7 @@
 // Copyright © Matt Jones and Contributors. Licensed under the MIT Licence (MIT). See LICENCE.md in the repository root
 // for more information.
 
+#include "NovelRT/Timing/Timestamp.hpp"
 #include <NovelRT/Ecs/Catalogue.hpp>
 #include <NovelRT/Ecs/ComponentCache.hpp>
 #include <NovelRT/Ecs/EntityCache.hpp>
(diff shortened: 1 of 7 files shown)


Ecs/Core/SystemScheduler.cpp
```diff
@@ -9,12 +10,30 @@
...
 #include <cassert>
 #include <functional>
 #include <unordered_map>
+#include <utility>
 
 namespace NovelRT::Ecs
 {
+    SystemId SystemScheduler::ScheduleSystemJob(std::function<bool(Timing::Timestamp, Catalogue)> jobFnPtr)
+    {
+        static AtomFactory& systemIdFactory = AtomFactoryDatabase::GetFactory("SystemId");
```
FiniteReality, line 23: I think we should probably use a different name/key for this, but it's not a hard requirement.

Ecs/Core/include/NovelRT/Ecs/SystemScheduler.hpp
```diff
@@ -41,6 +40,17 @@ namespace NovelRT::Ecs
         friend class Catalogue;
 
     private:
+        struct SystemJobInfo
+        {
+            std::function<bool(Timing::Timestamp, Catalogue)> jobFnPtr;
+            std::atomic_bool isDone{false};
```
FiniteReality: Out of curiosity, why are we using an atomic here?

Ecs/Core/SystemScheduler.cpp
```diff
@@ -275,12 +300,58 @@ namespace NovelRT::Ecs
...
+                                systemJob.isDone.store(systemJob.jobFnPtr(_currentDelta, Catalogue(poolId, *this)),
+                                                       std::memory_order_relaxed);
+                            });
+                    }
+
+                    _ecsTasks->wait();
+
+                    bool hadTombstones = false;
+
+                    // a bit ugly, but couldn't think of a better way to do it right now - Matt J.
```
FiniteReality, line 319: Are we not able to remove `std::remove_if` to move the completed ones to the end?

Ecs/Core/SystemScheduler.cpp
```diff
@@ -275,12 +300,58 @@ namespace NovelRT::Ecs
...
+                    for (auto& [_, systemJob] : _systemJobs)
+                    {
+                        unused(_);
+
+                        _ecsTasks->run(
+                            [&systemJob, this]()
+                            {
+                                auto poolId = static_cast<size_t>(tbb::this_task_arena::current_thread_index());
+                                systemJob.isDone.store(systemJob.jobFnPtr(_currentDelta, Catalogue(poolId, *this)),
+                                                       std::memory_order_relaxed);
```
FiniteReality: If the atomics are necessary here, is memory_order_relaxed the correct value here? My gut feeling is we probably want seq_cst by default?

Ecs/Core/SystemScheduler.cpp
RubyNova: I explicitly chose this to avoid confusion or clashes from a UX perspective

Ecs/Core/include/NovelRT/Ecs/SystemScheduler.hpp
RubyNova: because it is technically concurrent access and so this seemed safer to me, i could be wrong though

Ecs/Core/SystemScheduler.cpp
RubyNova: From what I gathered, none of the algorithm stuff would work here. I could be wrong on that? but yeah.

Ecs/Core/SystemScheduler.cpp
RubyNova: Maybe? I wasn't too sure at the time. I just did what worked and what made sense in the moment in this particular case.

FiniteReality: From what I understand about atomics (Which isn't much, mind you) you generally want seq_cst (sequentially consistent) unless you know a weaker one is more appropriate.

Ecs/Core/SystemScheduler.cpp
FiniteReality: That's fine then, we can look at this in the future.

Ecs/Core/include/NovelRT/Ecs/SystemScheduler.hpp
FiniteReality: I disagree - we only ever have a single thread writing to it (the lambda passed to ecsTasks->run) and we explicitly wait for them all to finish before reading from it (again, on a single thread). The synchronization is already handled for us.

Ecs/Core/SystemScheduler.cpp
FiniteReality: :woman_shrugging: I'm not sure what value that nets us here.

EDIT: I also don't have a better name, so I guess this is fine for now.

FiniteReality (approved):

